PR #6, opened 2017-07-10
Webapp 10: Impressum for every location

klinzo:

### DIFF
--- a/src/routes/DisclaimerPage/index.js
+++ b/src/routes/DisclaimerPage/index.js
@@ -62,7 +62,7 @@ class DisclaimerPage extends React.Component {
     return (
       <Layout languageCallback={this.changeLanguage}
               currentLanguage={this.props.language}>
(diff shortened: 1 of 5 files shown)


src/routes/DisclaimerPage/index.js
```diff
@@ -0,0 +1,83 @@
...
+
+  getLocation () {
+    return this.props.match.params.location
+  }
+
+  render () {
+    return (
+      <Layout languageCallback={this.changeLanguage}
+              currentLanguage={this.props.language}>
+        { this.props.disclaimerPayload.data !== null ? <Page page={this.props.disclaimerPayload.data}/> : null }
```
maxammann: Can be simplified to `{ this.props.disclaimerPayload.data ? <Page page={this.props.disclaimerPayload.data}/> : null }`

maxammann: Not important just JS style and it's useful for testing the review function :D

PR #8, opened 2017-07-24
Fixes WEBAPP-17 and also adds a simple search page

maxammann:


maxammann:
@Schedulaar all these changes are basically to remove the url generation from the ContentList/Categories component. These components take now a map url -> page, instead of a parent url and a parent page

This makes these componets reusable. 
Basically just check for bad style or exception sources

src/routes/LocationPage/Hierarchy.js
```diff
@@ -1,16 +1,20 @@
...
+
+import normalizeUrl from 'normalize-url'
+
 import { last } from 'lodash/array'
 import { isEmpty } from 'lodash/lang'
 
 export default class Hierarchy {
-  constructor (path) {
+  constructor (path = '') {
     this._path = path ? path.split('/').filter((path) => path !== '') : []
```
maxammann: Not currently, but an empty Hierarchy makes sense

Schedulaar: You're right, ToBoolean(emptyString) returns false, I didnt know that ~ JSLove

PR #10, opened 2017-07-29
WEBAPP-25: Add __VERSION__ define

maxammann:


maxammann:
Indeed, this will help us to identify the current version on https://integreat-1173.firebaseapp.com/

Schedulaar:
What about merging? 😄

maxammann:
@Schedulaar yee, meh. Footer is now ready right? So we could use this now :P

PR #11, opened 2017-08-01
WEBAPP-35

Schedulaar:
Add search button to header: Comments / other improvements:
Please check the current header design and functionality live.
Location passed from Route-Pages to Layout, where a navigation object is intialized (which holds urls for basic navigation).
Fixed SearchPage to support switching languages.
Other Header-UI-Improvements: Added a new smaller logo (on thin devices) for 'home' which now points to the route url (categories) of the Location. (we still have a menuitem for location-selection)
better responsiveness needed, bc another icon was added)


src/components/Layout/HeaderDropDown.js
```diff
@@ -30,14 +30,19 @@ class HeaderDropDown extends React.Component {
...
+      <a className={cx(this.props.className, style.dropDownItem)}>
         <FontAwesome name={this.props.fontAwesome}
                      className={cx(this.state.dropDownActive ? style.itemActive : '', style.item)}
                      onClick={this.toggleDropDown}/>
         <div className={cx(style.dropDown, this.state.dropDownActive ? style.dropDownActive : '')}>
-          {this.props.children}
+          {
+            React.Children.map(this.props.children,
+              child => React.cloneElement(child, {
+                closeDropDownCallback: this.handleClickOutside
```
Schedulaar: the way it was done before didn't exist. It just closed because the language callback was executed which made the endpoint refetch data about available languages and delete the DropDown from the header for a short period of time.
Since this is quite common practice I think (about the first google page) no doc needed.

src/components/Layout/index.js
```diff
@@ -3,13 +3,15 @@ import PropTypes from 'prop-types'
...
+import Navigation from 'Navigation'
 import Header from './Header'
 import style from './style.css'
 
 class Layout extends React.Component {
   static propTypes = {
     className: PropTypes.string,
-    currentLanguage: PropTypes.string.isRequired,
+    currentLanguage: PropTypes.string,
+    currentLocation: PropTypes.string,
```
Schedulaar: problem when requiring: landing page / location selection uses layout.

src/components/Layout/HeaderDropDown.js
```diff
@@ -30,14 +30,19 @@ class HeaderDropDown extends React.Component {
 
   render () {
     return (
-      <div className={cx(this.props.className, style.dropDownItem)}>
+      <a className={cx(this.props.className, style.dropDownItem)}>
```
Schedulaar: not quite happy with that either, but NavElement renders as Anchor and browsers have default values for many css-props - so the lang dropdown was slightly bigger than the other nav-elements because of these default-values. To avoid long messy css - props I set this as an anchor

Schedulaar: replaced it with span - should work better since span is - like a - an "inline tag" and thus closer to the defaults of a

PR #13, opened 2017-08-05
WEBAPP-36

Schedulaar:
Add link to disclaimer to footer


locales/ar/common.json
```diff
@@ -1 +1,3 @@
-{}
+{
+  "ImprintAndContact": "هيئة التحرير وجهة الاتصال"
```
maxammann: We should stick to camelCase here

locales/de/common.json
```diff
@@ -1 +1,3 @@
-{}
+{
+  "ImprintAndContact": "Impressum und Kontakt"
```
maxammann: see above and you should create a new language file Footer.json, see Search.json

Schedulaar: I think maintaining all these language files seperately is more work than just one file per language. To be considered: There are about 10 languages to support ~ if we have ten different sections like footer, header,... we got 100 files which is very messy. Or do we have some other tools to manage these files?

maxammann: true, i18next-resource-store-loader is used to generate the translations.
The impementation of the translations is in i18n. Different ns (namespaces) mean different files. We could merge all C
component translations in one json file

PR #15, opened 2017-08-08
WEBAPP-23

maxammann:
I'll open this now so you can look into the changes. Probably will need some mor work.

Major changes:

- Fetcher component handles all fetching, re-fetching and passes the result as soon as it's available to the child components
- Errors and the Spinner are displayed automatically and can be hidden
- The RichLayout shows now the Header (and Footer)
- The Layout is dump now
- Latest develop is merged


src/components/Fetcher/index.js
```diff
@@ -0,0 +1,133 @@
...
+  let payloadName = stateName + 'Payload'
+
+  function mapStateToFetcherProps (state) {
+    let props = mapStateToProps(state)
+
+    props[payloadName] = state[stateName]
+    return props
+  }
+
+  let Fetcher = class extends React.PureComponent {
```
Schedulaar: Have you checked the [docs](https://facebook.github.io/react/docs/react-api.html#react.purecomponent) for that? We should use PureComponent only if it "renders the same result given the same props and state." I'm not quite sure if the fetcher meets this criteria since we have dynamic loading of content.

src/components/Fetcher/index.js
```diff
@@ -0,0 +1,133 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Spinner from 'react-spinkit'
+import { connect } from 'react-redux'
+
+import LANGUAGE_FETCHER from 'endpoints/language'
+import PAGE_ENDPOINT from 'endpoints/page'
+import DISCLAIMER_FETCHER from 'endpoints/disclaimer'
+import LOCATION_FETCHER from 'endpoints/location'
```
Schedulaar: Why call these (except page_endpoint) fetcher?

src/components/Fetcher/index.js
```diff
@@ -0,0 +1,133 @@
...
+  mapLanguageToProps,
+  updateOnLanguageChange
+)
+
+export const LanguageFetcher = createFetcher(LANGUAGE_FETCHER,
+  (props) => {
+    return {
+      location: props.location,
+      language: props.language
+    }
```
Schedulaar: We should consider always using 'de' as urloptions for languageFetcher to ensure that we ALWAYS can return some available languages (currently: using French on Brandenburg -> switching to Augsburg -> error message (site can't be found) and the language dropdown doesn't work (since there's no cms backend for french, augsburg))

src/components/RichLayout/index.js
```diff
@@ -0,0 +1,65 @@
...
+        <LanguageFetcher location={this.props.location} hideError={true} hideSpinner={true}>
+          <HeaderAdapter languageCallback={this.gotoParent} language={this.props.language} location={this.props.location}/>
+        </LanguageFetcher>
+
+        <Layout className={this.props.className}>
+          {this.props.children}
+        </Layout>
+      </div>
+    )
+  }
```
Schedulaar: This looks really cool. The only thing I don't like about this 'way' of fetching data is: For example, every time the language fetcher fetches data (because the user switched language) the header completely disappears for a second (since the fetcher loads the data in) and pops back in. I think, it would be even cooler to have the language-fetcher as parent element of the languageflyout, so only the components that really depend on the data are not displayed (another side effect would be that the "languages-data" wouldn't be passed through all the other elements like header-adapter, header, etc.

src/routes/SearchPage/index.js
```diff
@@ -101,29 +53,34 @@ class SearchPage extends React.Component {
...
+          />
+          <ContentListAdapter location={this.getLocation()} filterText={this.state.filterText}/>
+        </PageFetcher>
+      </RichLayout>
+    )
+  }
 }
 
-export default connect(mapStateToProps)(SearchPage)
+export default SearchPage
```
Schedulaar: This looks really cool now. 👍

Schedulaar:
Remove .loading class from components/Location/style.css (since this is part of fetcher/style.css now)

src/components/Fetcher/index.js
```diff
@@ -0,0 +1,133 @@
...
+
+    componentWillMount () {
+      this.fetch(this.props)
+    }
+
+    componentWillUpdate (nextProps) {
+      if (shouldUpdate(this.props, nextProps)) {  // todo: this will need some more work to test -> an other issue as
+        // this is getting too big
+        this.fetch(nextProps)
+      }
```
Schedulaar: What would be an example use case of passing a shouldUpdate function? Can't think of one right now 🤔

src/components/Fetcher/index.js
```diff
@@ -0,0 +1,133 @@
...
+        newProps[payloadName] = payload
+
+        return (
+          <div>
+            {
+              React.Children.map(this.props.children,
+                (child) => React.cloneElement(child, Object.assign({}, this.props, newProps)))
+            }
+          </div>
+        )
```
Schedulaar: Well... As i'm not a fan of creating "useless" wrapper divs (not a good DOM Convention - bad readability, performance) we maybe should go back to having just one child for such a "wrapper" component like fetcher (I think this also applies to the HeaderDropDown Class I wrote 🤔 ). What do you think?

src/components/Fetcher/index.js
```diff
@@ -0,0 +1,133 @@
...
+
+    props[payloadName] = state[stateName]
+    return props
+  }
+
+  let Fetcher = class extends React.PureComponent {
+    static propTypes = {
+      hideError: PropTypes.bool,
+      hideSpinner: PropTypes.bool
+    }
```
Schedulaar: maybe we also want an optional style property, so that the components can style their loading and error messages independently. If no style property is provided, take default.

src/components/Fetcher/index.js
maxammann: The props of a Fetcher are generally more simple, they are just params for the http query. But I'll change it back to Component. There's no reason to optimize if we have no probs :D

src/components/Fetcher/index.js
maxammann: They are used on the bottom of the file. This will change in an other CR. We need to merge the endpoins/ and Fetcher/ package. The Fetchers belong to the endpoint API and are not a standalone component.

Please take a look at this: https://www.npmjs.com/package/react-redux-fetch
I think we should not remove our code and just use this package because it is not used much and we're fucked when they stop supporting it. But we maybe can steal some ideas.

src/components/Fetcher/index.js
maxammann: I will do this and add a comment with todo:. I'm not sure how languages differ in different languages.

src/components/RichLayout/index.js
maxammann: Yes that makes sense but is an issue for another task. As soon as intelligent caching and refetching is implemented this won't be much of an issue.

I already noticed it when only the language button took some time to appear and this already was aweful. Therefore I think if it only happens once per location it actually looks better if the whole header takes some time to appear.

But I think this will be solved by the intelligent refreshing. But feel free to create another task after this merge.

src/routes/SearchPage/index.js
maxammann: thank yuuuu!!

src/components/Fetcher/index.js
maxammann: some components will need more properties than language or location so it can't be static.

shouldUpdate is needed because of infinite recursion:
language prop updates -> refetching -> state[stateName] updates -> props update -> refetching -> ....

src/components/Fetcher/index.js
maxammann: I think we will have a problem if we nest multiple Fetcher, because a div does not propagate the properties. so if we have:

```
<Fetcher>
<Fetcher>
<div>
<Component/>
<Component/>
</div>
</Fetcher>
</Fetcher>
```
The components won't get the properties. But I'm not sure.

maxammann: But we could create a component which takes the properties and passes them to multiple components. (replace the div)

Schedulaar: Yeah, I think such a wrapper-component would make more sense than the current "wrapper-div" implementation.

src/components/Fetcher/index.js
Schedulaar: Or optionally even let the Adapter Components render the spinner / error messages (like in react-redux-fetch)

src/components/Fetcher/index.js
maxammann: I have there something in my mind and will create a higher-order-components. Currently trying it but not sure whether it will succeeed.

maxammann: But yes the div will be gone :D

src/components/Fetcher/index.js
maxammann: yess will add an optional className. Error/Spinner should be handled in one place so they stay consistent

src/endpoints/disclaimer.js
```diff
@@ -8,6 +8,9 @@ export default new Endpoint(
   'disclaimer',
   'https://cms.integreat-app.de/{location}/{language}/wp-json/extensions/v0/modified_content/disclaimer?since={since}',
   (json) => {
+    if (!json) {
+      return {}
```
Schedulaar: Are you sure, you don't want to return [] as default value?

src/endpoints/location.js
```diff
@@ -6,10 +6,12 @@ export default new Endpoint(
   'locations',
   'https://cms.integreat-app.de/wp-json/extensions/v1/multisites',
   json => {
+    if (!json) {
+      return {}
```
Schedulaar: Same: don't you want to return `[]`?

Schedulaar:
(copy because of no response) Remove .loading class from components/Location/style.css (since this is part of endpoints/Fetcher.css now)

src/endpoints/location.js
maxammann: groupBy returns an object

src/endpoints/disclaimer.js
maxammann: I think reduce returns an object aswell, doesn't it?

src/endpoints/location.js
Schedulaar: whoops yeah that's right, don't like it though - this sorts the array where it should only load an array in. Sorting should happen somewhere else, but yeah... :(

Schedulaar: Well... let's stick with it 😆

src/endpoints/disclaimer.js
Schedulaar: yap forget about it :D

src/components/Fetcher/index.js
Schedulaar: Is it gone yet?
Also according to the Higher-order-component conventions (HOC) [(click)](https://facebook.github.io/react/docs/higher-order-components.html) our endpt.withFetcher() is not an HOC per def, because an HOC has a component as an argument that the return-component renders. Did you look into that doc?

PR #23, opened 2017-10-17
WEBAPP-54: adjusted margins of list categories and elements on the landing page 

klinzo:
fixed search icon above search input field in firefox


src/components/Location/Search.css
```diff
@@ -8,7 +8,6 @@
 }
 
 .search-icon {
-  width: var(--search-logo-width);
```
maxammann: I'm not sure but  maybe search-logo-width is just too small for some browsers for 1.2em font size and the search icon.

PR #32, opened 2017-10-21
WEBAPP-7: Implement Caching Strategy

Schedulaar:
The way I implemented it:
Cache all the data, we once load, for two days. This way we still have a solid "update"-frequency.
Caching with workBox only works with ServiceWorker: The service-worker is registered in the beginning and works as a proxy between our app and the backend. Sadly, service workers are currently only supported in Chrome and Firefox. So it's not working for IE, Edge and Safari 😞, but in dev for Edge and Safari: https://jakearchibald.github.io/isserviceworkerready/.
Do you think this is enough as a first step, or should I research for some other caching strategy - then I'll probably need to touch endpoint-fetching-strategy.


maxammann:
actually thought about caching it by using local storage but local storage is probably even better.

The only problem I see is support. The caching mostly targets mobile devices. The iOS browser does not support it though: https://caniuse.com/#feat=serviceworkers

So we cover only 66% of all browsers. I think we can accept this, but we definitely need a fallback option :/

.gitignore
```diff
@@ -90,3 +90,7 @@ fabric.properties
 public/dist
 public/index.html
 public/sitemap.xml
+public/sw.js
+.vscode/
```
maxammann: Add this instead:
```
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
```

.gitignore
```diff
@@ -90,3 +90,7 @@ fabric.properties
...
+public/workbox-sw.prod.v2.0.1.js
```
maxammann: Use wildcard instead of specific version:

`public/workbox-sw.prod.v*.js`

src/main.js
```diff
@@ -83,6 +83,11 @@ ReactDOM.render(App, container)
 // Sets the splash to hidden when the page is rendered
 document.getElementById('splash').className += ' splash-hidden'
 
+if (navigator.serviceWorker) {
```
maxammann: Can not test it in IE11 currently, but is navigator always available?

PR #33, opened 2017-10-22
WEBAPP-53: Find better solution for breadcrumbs

klinzo:
removed name of top page from breadcrumbs and added a title instead
changed size of breadcrumbs


maxammann:
For the future: You can remove the newline for commit messages. If they are short they should be only one line.

maxammann:
I think this should also have a back button for mobile devices, right?

src/components/Content/Page.css
```diff
@@ -1,3 +1,8 @@
+.remote-heading {
+  margin: 30px 0 50px;
+  font-size: 40px;
```
Schedulaar: Same as above (except the centering).
I would also not call it .remote-heading, but instead just .heading because we don't include remote HTML content but only a string containing the title.

src/components/Content/TitledCategoriesTable.css
```diff
@@ -0,0 +1,5 @@
+
+.title {
+  margin: 30px 0 50px;
+  font-size: 40px;
+}
```
Schedulaar: I would prefer a `text-align: center;` here, because all other content is centered on the categories page.
Also I would decrease font-size to sth like `35px`, 40px can be a little bit loo large for longer titles and smaller screens.
From personal preference also `margin: 10px 0 30px;` But this can be discussed 😄 
But the styling of the headings should all match, so we should also adjust the "Where am I?"-Heading of the landing page imo.

src/components/Content/Page.js
```diff
@@ -12,6 +12,8 @@ class Page extends React.Component {
     /* We can insert our html here directly since we trust our backend cms */
     return (
       <div>
+        <div className={style.remoteHeading}
+             dangerouslySetInnerHTML={{__html: (this.props.page.title)}}/>
```
Schedulaar: Can't we adjust this to `<div className={style.remoteHeading}>{ this.props.page.title}</div>`?

src/components/Content/TitledCategoriesTable.js
```diff
@@ -0,0 +1,34 @@
...
+    return (
+      <div>
+        <div className={style.title}>{this.getTitle()}</div>
+        <CategoriesTable pages={this.props.pages}/>
+      </div>
+    )
+  }
+}
+
+export default withFetcher(LOCATIONS_ENDPOINT, true, true)(TitledCategoriesTable)
```
Schedulaar: It seems, we currently cannot have two instances of the same fetcher in two different spots at the same time (here: LocationsFetcher in Breadcrumbs and in TitledCategoriesTable).
A possible workaround for now would be to use the LocationFetcher in the LocationPage-Route in the FetchingContentWrapper and pass the title down to the Breadcrumbs and the Content.
But I would actually prefer, that the Fetchers care about that logic, so we can easily put Fetchers where we want to.
I think we need to discuss this with @maxammann (preferably on Wed)?

Schedulaar:
Regarding: https://github.com/Integreat/integreat-webapp/pull/33#issuecomment-338482701
I kind of like the implementation Steff made, so we still have all layers of the menu visible, but have the title of the current site stand out more, since it now has a large heading and is outside of the breadcrumbs.
Also we usually only have 3 layers, like `Augsburg > Welcome > Arrival Information`. Then the Breadcrumb reads `Augsburg > Welcome` and below that, the title says `Arrival Information`.
I like this more, than just having a back-arrow saying sth like `< Welcome`, because - again - if you're visiting from google and get directly directed ( 😮 ) to `Arrival Information`, then the visitor can't even immediately see that he's been directed to the Augsburg instance, because the back-arrow only reads `Welcome`.
Things like that prohibit me from thinking we should just copy the native app, because experience can be different (same with branding integreat logo).

src/components/Content/Breadcrumb.css
```diff
@@ -6,14 +6,14 @@
...
-  font-size: 22px;
+  font-size: 16px;
   content: ' > ';
   user-select: none;
 }
 
 .level {
   color: #9c9c9c;
-  font-size: 20px;
+  font-size: 15px;
```
Schedulaar: I think we shold still increase the font-size of the breadcrumbs to at least 16-17px, so the user can easily touch them.

src/components/Content/TitledCategoriesTable.js
Schedulaar: It seems, the actual problem occurs, e.g. when transitioning from Augsburg (Home), to Augsburg-> Welcome: At Welcome the breadcrumbs should appear, but dont. The problem is that the TitledCategoriesTable's fetcher invalidates the locations endpoint data and then the breadcrumbs are hidden by the its locationsFetcher (since the payload-state has changed by invalidation).

src/components/Content/Breadcrumb.css
maxammann: We can increase the touch-target. I think if it looks ok we should definitely make it smaller to save space.

src/components/Content/Page.js
maxammann: Probably titles are html, but I could be wrong :P

maxammann: Or... I thtink they shouldn't. hmmm

src/components/Content/TitledCategoriesTable.js
maxammann: going to test this

maxammann:
@Schedulaar 
The problem is that we have very, very long titles for pages. It maybe looks strange if we have a breadcrumb over multiple lines. So for this case a back button would probably look better

src/components/Content/TitledCategoriesTable.js
maxammann: Created issue: https://integreat.atlassian.net/browse/WEBAPP-74

maxammann:
This is the worst case: https://www.browserstack.com/download?url=https://www.browserstack.com/issue-tracker/c27e9247a22dbbf516c32d0d7a20ba2bc65bf13d/realdroid_Samsung_Galaxy_S8.jpg

src/components/Content/Page.css
klinzo: done

src/components/Content/Page.js
klinzo: done

src/components/Content/TitledCategoriesTable.css
klinzo: centered the text and adjusted the style to the "Where are you?" heading

src/components/Content/TitledCategoriesTable.js
klinzo: works well with the solution of maxamman, WEBAPP-74

src/components/Content/Breadcrumb.css
klinzo: do you really think we need a bigger font-size/touch-target? at least on desktop this shouldn't be a problem, on mobile maybe

Schedulaar: No it's ok for now :D. Leave it at 15.

klinzo: added 10px on top and bottom of the links, so that the touch-target is better
this could be increased/decreased easily if you want to

klinzo: removed it again, otherwise scrolling would be possible in vertical direction

src/components/Content/Breadcrumb.css
```diff
@@ -1,19 +1,21 @@
 @import url('../Helper/Helper.css');
 
 .breadcrumbs {
+  overflow: scroll;
   margin-bottom: 20px;
```
Schedulaar: You can remove the margin-bottom now, since this has no effect (somehow :D)

src/components/Content/Breadcrumb.css
```diff
@@ -1,19 +1,21 @@
 @import url('../Helper/Helper.css');
 
 .breadcrumbs {
+  overflow: scroll;
```
Schedulaar: You could replace `overflow: scroll;` with `overflow-x: auto`, so that the vertical scrollbar is never shown, and the horizontal scrollbar is only shown when it's needed.

src/components/Content/Breadcrumb.css
```diff
@@ -1,19 +1,21 @@
...
   content: ' > ';
   user-select: none;
 }
 
 .level {
   color: #9c9c9c;
-  font-size: 20px;
+  font-size: 15px;
   user-select: none;
 }
```
Schedulaar: You can also remove `breadcrumb:last-child .level {
   color: black;
 }` since this makes no sense anymore.

src/components/Content/Page.js
```diff
@@ -12,6 +12,9 @@ class Page extends React.Component {
     /* We can insert our html here directly since we trust our backend cms */
     return (
       <div>
+        <div>
+          <h1 className={style.heading}>{this.props.page.title}</h1>
+        </div>
```
Schedulaar: You don't need to wrap the `<h1>` in a `<div>`. You can just replace
```html
<div>
  <h1 className={style.heading}>{this.props.page.title}</h1>
</div>
```
with
```html
  <h1 className={style.heading}>{this.props.page.title}</h1>
```
It would also be cool, if you create a simple component `Heading` which gets a string `title` as a property and displays just this h1 snippet with the correct styling. Then we can use that component on the Content Page, the LandingPage and so on, and have the styling in one place.

src/components/Content/Breadcrumb.css
klinzo: done

src/components/Content/Breadcrumb.css
klinzo: done

src/components/Content/Breadcrumb.css
klinzo: done

src/components/Content/Page.js
klinzo: i called it Caption since there is already a Heading component on the LocationPage

src/components/Location/Heading.css
```diff
@@ -1,12 +1,11 @@
...
 .logo {
   display: block;
   width: auto;
   height: 70px;
   margin: auto;
 }
+
+.caption {
+  margin: 0 0 30px;
+  vertical-align: middle;
```
Schedulaar: You can remove the `vertical-align: middle;` (vertical-align only has effect on inline elements (h1 isn't inline) or tables or similar) and give some top margin, so the caption doesnt stick to the image too much, like `margin: 20px 0 30px;`
(Although having a special margin here means, we still don't have uniform styling for the captions everywhere, but oh well :D )

src/components/Content/Caption.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import style from './Caption.css'
+
+export default class Caption extends React.Component {
+  propTypes = {
```
Schedulaar: change this to `static propTypes = {`

maxammann: @klinzo did you change this?

PR #35, opened 2017-10-26
WEBAPP-74: Fix fetchers if there are more than one in the view

maxammann:
This is work in progress so this is just open for discussion


src/endpoints/Endpoint.js
```diff
@@ -86,20 +83,28 @@ class Endpoint {
...
       /*
        todo:  check if there are any paramters left in the url: formattedURL.match(/{(.*)?}/)
        currently this does not work as unused paramaters are just removed from the url
        */
 
+      const lastUrl = getState()[this.name].requestUrl
+
+      if (lastUrl && lastUrl === formattedURL) {
+        // fixme: Use "cached"
+        return
```
maxammann: If there is a cached version in the state use it instead of refetching. This needs a timeout.

Schedulaar: Timeout for what exactly ?

src/endpoints/disclaimer.js
```diff
@@ -29,10 +30,5 @@ export default new Endpoint({
         order: page.order
       })
     }, null)
-  },
-  mapStateToOptions: (state) => ({
-    language: state.router.params.language,
-    location: state.router.params.location
-  }),
```
Schedulaar: IMO this should be defined in every endpoint itself, so the endpoint has full control over what's in its options.

src/endpoints/disclaimer.js
```diff
@@ -29,10 +30,5 @@ export default new Endpoint({
...
-  shouldRefetch: (options, nextOptions) => options.language !== nextOptions.language
```
Schedulaar: IMO same as above. Also: Currently we do not have any chance to have multiple separate endpoint-payloads from the same endpoint available. (Like if we had an /events/:id endpoint and want to load both /events/2 and /events/3, we cannot with this architecture. Should we take this into account at all?

src/endpoints/events.js
```diff
@@ -1,36 +1,29 @@
...
+        startDate: parseDate(event.event.start_date, event.event.start_time),
+        endDate: parseDate(event.event.end_date, event.event.end_time),
+        allDay: event.event.all_day === '0'
+      })
+    }))
+    .filter(event => event.date.startDate)
+    .filter(event => event.date.startDate > Date.now() - MILLISECONDS_IN_A_DAY)
+    .sort((event1, event2) => event1.date.startDate - event2.date.startDate)
+  )
+  .build()
```
Schedulaar: TBH, I don't see benefits with this builder-structure.

src/endpoints/EndpointBuilder.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  build () {
+    return new Endpoint(this._name, this._url, this._mapper, this._stateMapper, this._refetchLogic)
+  }
+}
+
+export default EndpointBuilder
+
+export const endpoint = (name) => new EndpointBuilder(name)
```
Schedulaar: I don't like that you call the EndpointBuilder endpoint, but not endpointBuilder, like
`export const endpointBuilder = (name) => new EndpointBuilder(name)`

src/endpoints/Endpoint.js
```diff
@@ -54,10 +50,10 @@ class Endpoint {
    * @param {string} name The name of this endpoint. This is used as key in the state and as Payload name. The Payload name is name + 'Paylaod'
    * @param {string} url The url with params (params are used like this: https://cms.integreat-app.de/{location}/{language})
    * @param {function} jsonToAny Transforms the json input to a result
-   * @param {mapStateToOptionsCallback} mapStateToUrlParams Maps the state to the url params which are needed in the Fetcher component
+   * @param {mapStateToOptionsCallback} mapStateToOptions Maps the state to the url params which are needed in the Fetcher component
    * @param shouldRefetch Takes the current and the next props and should return whether we should refetch
    */
-  constructor ({name, url, jsonToAny, mapStateToOptions = DUMMY, shouldRefetch = () => false}) {
```
Schedulaar: you could rename jsonToAny to transformJsonToResult or similar? 🤓

src/endpoints/EndpointBuilder.js
```diff
@@ -0,0 +1,47 @@
...
+  _refetchLogic
+
+  constructor (name) {
+    this._name = name
+    this._stateMapper = (state) => ({
+      language: state.router.params.language,
+      location: state.router.params.location
+    })
+
+    this._refetchLogic = (options, nextOptions) => options.language !== nextOptions.language
```
Schedulaar: I don't like that the stateMapper has a default value.
Although I'd like the default refetchLogic to do just a deepComparison between options and nextOptions 🤔 Is that even possible? :D

src/endpoints/Endpoint.js
maxammann: Sure, this is still in progress, just wanted to share my ideas

src/endpoints/Endpoint.js
maxammann: So with this fix the data from the endpoint is only fetched once per page visit. Usually this shouldn't be a problem because the state is empty on each page visit but it can be cached by some extension of redux.

So maybe a lastFetchedDate state variable would be good.

src/endpoints/EndpointBuilder.js
maxammann: Hm maybe that's now possible, right. Previsouly options contained stuff like dispatch or t from the components properties.

src/endpoints/disclaimer.js
maxammann: yeah I will add a withDefaultStateMapper() or similar and remove the defaults

src/endpoints/EndpointBuilder.js
maxammann: +1

src/endpoints/events.js
maxammann: The positive effect is that we can move validation of all properties/default values from the contructor of Endpoint to a separate class. The Endpoint constructor is already quite full. So just code style

src/endpoints/Endpoint.js
maxammann: So those three lines need discussion. The first component which is displayed actually fetches the data and then the data from the state is used.

src/endpoints/disclaimer.js
maxammann: Thats quite difficult with our design. Also where should the 2 or 3 come from? Currently they can only come from the redux state. So I would suggest to do lazy-loading in this case. Return an array of all events. If you need a specific one it will be loaded jit (and maybe update the redux state).

But yeah it really depends which queries we have.

src/endpoints/Endpoint.js
Schedulaar: Ah, so you mean a timespan for cache-expiration. (Timeout sounds like sth's going wrong :D ) Yes, a lastFetchedDate in the payload would be cool! I would suggest a span of 1-2 hours. Maybe we could even have a logic, that, when there's no internet connection and the cache has expired, we still use the cached version. (Even though we could probably also do this on service-worker lvl)
Furthermore (at least once the webapp runs in standalone) we can show a label saying when we last refreshed the content, when it's been loaded from cache.
Additionally it'd be cool, if the user can then force a refresh (meaning not invalidating, but only expiring the cache)

src/endpoints/EndpointBuilder.js
Schedulaar: No, it didn't. There were already the options in a separate property of the fetcher.

src/endpoints/Endpoint.js
maxammann: I'm not really sure if it's worth it. I think it will be quite rare that that timeout gets triggered. I mean who has the app opened for more than 2 hours and expects the app to refresh after you have changed something in the cms?
You would just reload the page and then see the changes. So maybe the current implementation is enough

maxammann: I think there is just no use case.

src/endpoints/Endpoint.js
```diff
@@ -53,22 +49,23 @@ class Endpoint {
   /**
    * @param {string} name The name of this endpoint. This is used as key in the state and as Payload name. The Payload name is name + 'Paylaod'
    * @param {string} url The url with params (params are used like this: https://cms.integreat-app.de/{location}/{language})
-   * @param {function} jsonToAny Transforms the json input to a result
-   * @param {mapStateToOptionsCallback} mapStateToUrlParams Maps the state to the url params which are needed in the Fetcher component
+   * @param {function} mapData Transforms the json input to a result
```
Schedulaar: mapData is too generic imo. Why not mapJsonResponse since this would also fit the javaDoc comment?

src/endpoints/Endpoint.js
```diff
@@ -86,37 +83,45 @@ class Endpoint {
...
+      const formattedURL = format(this.url, urlParams)
       /*
        todo:  check if there are any paramters left in the url: formattedURL.match(/{(.*)?}/)
        currently this does not work as unused paramaters are just removed from the url
        */
 
+      const lastUrl = getState()[this.name].requestUrl
+
+      if (lastUrl && lastUrl === formattedURL) {
+        // Use "cached"
```
Schedulaar: I think you your comment wants to say, "Use the current payload which we already have in place in the redux store.". "Use cached" sounds for me like "We have stored sth somewhere and load that in now instead of fetching from server again."?

src/endpoints/Endpoint.js
Schedulaar: Well, I was talking about the case where we would have that Redux-Cache-Plugin-thingy in place. Then you can have content really cached over a long period and when you're offline the user could be unsure how up-to-date the content is.

Schedulaar: But then that discussion should be outside of this PR.

src/endpoints/language.js
```diff
@@ -1,17 +1,15 @@
...
+  .withStateMapper().fromFunction((state) => ({
     location: state.router.params.location,
     language: 'de'  // todo:  This forces that the languages are always fetched in german language. German always
                     //        exists in the backend -> a langauge switch always works
-  })
-})
+  }))
+  .withStateMapper().fromArray(['location', 'language'], (state, paramName) => state.router.params[paramName])
+  .withMapper(json => json.map(language => new LanguageModel(language.code, language.native_name)))
+  .withRefetchLogic(() => false)
```
Schedulaar: We only need location from the router params. Also Refetch-Logic should be the default imo.

src/endpoints/EndpointBuilder.js
```diff
@@ -0,0 +1,68 @@
...
+  _url
+  _mapper
+  _stateMapperBuilder
+  _refetchLogic
+
+  constructor (name) {
+    this._name = name
+    this._stateMapperBuilder = new StateMapperBuilder(this)
+
+    this._refetchLogic = (options, nextOptions) => isEqual(options, nextOptions)
```
Schedulaar: Replace with
```javascript
this._refetchLogic = (options, nextOptions) => !isEqual(options, nextOptions)
```

src/endpoints/language.js
maxammann: done

src/endpoints/EndpointBuilder.js
maxammann: done

src/endpoints/Endpoint.js
maxammann: done

PR #37, opened 2017-10-29
WEBAPP-72, WEBAPP-58, WEBAPP-65, WEBAPP-34, WEBAPP-81

maxammann:
Fixes:

- Scroll behaviour
- Remove sitemap
- Translations
- Location name
- Add version display


tools/webpack.config.js
```diff
@@ -228,6 +228,8 @@ const config = {
 // Optimize the bundle in release (production) mode
 if (!isDebug) {
   config.plugins.push(new webpack.optimize.UglifyJsPlugin({
+    parallel: true,
+    cache: true,
```
Schedulaar: What do these two?

src/components/RichLayout/Footer.js
```diff
@@ -26,10 +26,21 @@ class Footer extends React.Component {
     }
   }
 
+  getVersion () {
+    // eslint-disable-next-line no-undef
+    if (__DEV__) {
+      // eslint-disable-next-line no-undef
+      return <div className={style.item}>{__VERSION__}</div>
```
Schedulaar: __VERSION__ only includes a commit hash when the version is not exactly on that hash, right?

Schedulaar: There needs to be a little bit more visual separation of the disclaimer link and the version. Maybe you can use '|' for that?

maxammann: This is just for debugging so I think visuals are not important here

tools/webpack.config.js
maxammann: Execute in paralell and cache some stuff for a speedup :P

src/components/RichLayout/Footer.js
Schedulaar: ah yeah, right. :D

PR #38, opened 2017-10-31
WEBAPP-85: Convert LocationBig to svg

Schedulaar:
If you're concerned: I tested it in every browser (IE11, Chrome, FF, Edge, iOS >= 6)


maxammann:
merge!!!

PR #39, opened 2017-11-02
WEBAPP-73: Stay on the same site when changing language if the site is available in that language

klinzo:
for now this is just a pull request to enable comparing and discussing of changes made and possible solutions, the changes made in PageModel, page endpoint and LocationPage/index should be okay, the changes in RichLayout/index are not really working because the pages fetched by withFetcher(PAGE_ENDPOINT) are still the pages with the old language, I'm still looking for a solution


Schedulaar:
I've set the base branch to WEBAPP-74, so I can have a better overview of what's new. We should change it back to develop once WEBAPP-74 has been merged (hopefully sooner than later @maxammann :))

Schedulaar:
A solution I'd suggest would be to do the work in two steps: When you're at `/augsburg/de/willkommen` and have 1222 as an Id for the english welcome page, then just link the user to `/augsburg/en?pageId=1222` (so create a new routing routine somewhere) and then resolve the pageId to the right path and forward the user to that path.

PR #41, opened 2017-11-10
WEBAPP-73: Use promise for result of fetch

maxammann:


Schedulaar:
Closed, because similar solution in https://github.com/Integreat/integreat-webapp/pull/42

PR #42, opened 2017-11-10
WEBAPP-73: Fix Fetcher API

Schedulaar:
// todo

### DIFF
--- a/src/endpoints/withFetcher.js
+++ b/src/endpoints/withFetcher.js
@@ -38,6 +38,9 @@ function withFetcher (endpoint, hideError = false, hideSpinner = false) {
       }
 
       componentWillMount () {
+        // We need to have this discussion in mind, when building the fetcher architecture
+        // Because a store dispatch in the componentWillMount has no immediate effect on the props for the first render.
+        // https://github.com/reactjs/react-redux/issues/210#issuecomment-166055644
         this.fetch(this.props.options)
       }
 
--- a/src/endpoints/withFetcher.js
+++ b/src/endpoints/withFetcher.js
@@ -5,6 +5,7 @@ import Spinner from 'react-spinkit'
 
 import Error from 'components/Error'
 import style from './Fetcher.css'
+import {isEqual} from 'lodash/lang'
 
 function createStateToPropsMapper (endpoint) {
   return (state) => ({
@@ -20,7 +21,7 @@ function withFetcher (endpoint, hideError = false, hideSpinner = false) {
 
       constructor () {
         super()
-        this.state = { requesting: false }
+        this.state = { isDataAvailable: false }
       }
 
       fetch (options) {
@@ -28,14 +29,12 @@ function withFetcher (endpoint, hideError = false, hideSpinner = false) {
           throw new Error('options are not valid! This could mean your mapStateToOptions() returns ' +
             'a undefined value!')
         }
-
-        this.setRequesting(true)
-        this.props.dispatch(endpoint.requestAction(options, options))
-          .then(() => { this.setRequesting(false) })
+        const isDataAvailable = this.props.dispatch(endpoint.requestAction(options, options))
+        this.setIsDataAvailable(isDataAvailable)
       }
 
-      setRequesting (requesting) {
-        this.setState({ requesting })
+      setIsDataAvailable (isDataAvailable) {
+        this.setState({ isDataAvailable })
       }
 
       componentWillMount () {
@@ -43,7 +42,10 @@ function withFetcher (endpoint, hideError = false, hideSpinner = false) {
       }
 
       componentWillUpdate (nextProps) {
-        if (endpoint.shouldRefetch(this.props.options, nextProps.options)) {
+        // Dispatch new RequestAction to ask the endpoint whether the fetcher can display the data, if
+        // (a) the endpoint properties change or
+        // (b) the fetcher receives new payload information from the store (e.g. because a payload has been fetched)
+        if (endpoint.shouldRefetch(this.props.options, nextProps.options) || !isEqual(this.props[endpoint.payloadName], nextProps[endpoint.payloadName])) {
           this.fetch(nextProps.options)
         }
       }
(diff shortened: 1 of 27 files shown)


src/endpoints/withFetcher.js
```diff
@@ -20,30 +21,31 @@ function withFetcher (endpoint, hideError = false, hideSpinner = false) {
...
         if (!options) {
           throw new Error('options are not valid! This could mean your mapStateToOptions() returns ' +
             'a undefined value!')
         }
-
-        this.setRequesting(true)
-        this.props.dispatch(endpoint.requestAction(options, options))
-          .then(() => { this.setRequesting(false) })
+        const isDataAvailable = this.props.dispatch(endpoint.requestAction(options, options))
+        this.setIsDataAvailable(isDataAvailable)
```
maxammann: setIsDataAvailable is called only once here. So if it is false here it will always stay false and the child component won't get displayed. So fetch() has to be called multiple times that the component state changes to true?

Schedulaar: Yep, the fetcher triggers the dispatch via fetch() which returns if the data is available yet. If not, the fetch-logic (in the endpoint) is triggered and once it's done the fetcher's componentWillUpdate will be triggered (and thus fetch() again) because the payload has been changed.

maxammann: do we need to trigger the dispatch after fetch() is called the second time?

Schedulaar: Well, it could (potentially) be, that the payload is still not correct. To assure, that no data is shown, until valid data is there, the dispatch is called a second time (where it only returns true if everything's valid).

src/routes/LocationPage/index.js
```diff
@@ -30,20 +31,36 @@ class ContentWrapper extends React.Component {
...
       return
     }
 
-    this.props.dispatch(setCurrentAvailableLanguages(hierarchy.top()))
+    const currentPage = hierarchy.top()
+    const redirect = (id, language) => `/${this.props.location}/${language}/redirect?id=${id}`
+    const currentAvailableLanguages = reduce(currentPage.availableLanguages, (acc, id, language) => {
+      acc[language] = redirect(id, language)
+      return acc
+    }, {})
```
Schedulaar: How can I rewrite that?
I want to map `{ 'de': 12042, 'en': 2131 }` into `{ 'de': '/redirected/path', 'en': '/redirected/path/2' }`

src/main.js
```diff
@@ -38,32 +41,52 @@ let App = (
...
-              <Fragment forRoute="/redirect"><PageRedirector/></Fragment>
+              <Fragment forRoute="/events">
+                <RichLayout><EventsPage/></RichLayout>
+              </Fragment>
+              {/* Matches /augsburg/de/redirect -> Redirect */}
+              <Fragment forRoute="/redirect">
+                <Layout><PageRedirector/></Layout>
+              </Fragment>
+              {/* Matches /augsburg/de/fetchPdf/* -> Redirect */}
+              <Fragment forRoute="/fetchPdf/*">
```
Schedulaar: I think we should avoid having camelCase in the url. `fetch-pdf` would be more appropriate imo

src/routes/DisclaimerPage/index.js
```diff
@@ -2,61 +2,26 @@ import React from 'react'
...
-      <RichLayout>
-        <FetchingContentWrapper/>
-      </RichLayout>
-    )
-  }
-}
-
-export default DisclaimerPage
+  withAvailableLanguageUpdater((location, language) => `/${location}/${language}/disclaimer`)
+)(DisclaimerPage)
```
Schedulaar: Gorgeous 🥇

src/routes/EventsPage/index.js
```diff
@@ -1,58 +1,27 @@
...
-      <RichLayout>
-        <FetchingContentWrapper/>
-      </RichLayout>
-    )
-  }
-}
-
-export default EventsPage
+  withAvailableLanguageUpdater((location, language) => `/${location}/${language}/events`)
+)(EventsPage)
```
Schedulaar: Gorgeous 🥇

src/routes/LocationPage/index.js
```diff
@@ -32,35 +29,46 @@ class ContentWrapper extends React.Component {
...
+    this.props.dispatch(setLanguageChangeUrls(languageChangeUrls))
   }
 
   componentWillUnmount () {
-    this.props.dispatch(setCurrentAvailableLanguages({}))
+    this.props.dispatch(setLanguageChangeUrls({}))
+  }
+
+  getPdfFetchPath () {
+    let path = `/${this.props.location}/${this.props.language}/fetchPdf/`
```
Schedulaar: also here `fetch-pdf` imo

src/main.js
klinzo: why did you change from RichLayout to a normal Layout for PageRedirector and PdfFetcherPage? I think it's flickering with the normal layout which doesn't have our header

Schedulaar: Yup, pageredirector should have Richlayout, whereas PDF Fetcher shouldnt have a Richlayout (since its Always opened in a New Tab where we just wait for the PDF to have been generated and immediately redirect to that PDF. So there's no header necessary there.)

maxammann: But we should disable the buttons in richlayout. I'm not sure what happens if you change the language during a redirect :D

Schedulaar:
No, we should be good to go, if the fetcher (of the pageredirector) is
unmounted before the fetching finishes then we'll never mount the
pageredirector and therefore never have a "delayed redirect". The buttons
in the header still have the correct meaning (except for the language
flyout which sends you to the home of the corresponding language)
So maybe we can disable the buttons in the language flyout for that
particular moment (via RichLayout property ?)

Max Ammann <notifications@github.com> schrieb am Do., 16. Nov. 2017, 10:40:

maxammann:
@Schedulaar @klinzo and I come to the conclusion that we will just change it to RichLayout for now and find a solution for our Layouts to unifi them. We have too many layouts atm. And also an other task which would introduce a new one.

PR #47, opened 2017-11-18
WEBAPP-98: adjust the design of the footer

klinzo:
force footer to stay on bottom of page/bottom of viewport depending on which is bigger
adjusted design of the footer, maybe we have to change the color when we get our new color palette

### DIFF
--- a/src/components/RichLayout/Header.css
+++ b/src/components/RichLayout/Header.css
@@ -7,7 +7,7 @@
 }
 
 .spacer {
-  margin-bottom: var(--header-height);
+  padding-bottom: var(--header-height);
 }
 
 .header {
--- a/src/components/RichLayout/Header.css
+++ b/src/components/RichLayout/Header.css
@@ -7,7 +7,7 @@
 }
 
 .spacer {
-  padding-bottom: var(--header-height);
+  padding-bottom: var(--header-height-small);
 }
 
 .header {
--- a/src/components/RichLayout/index.js
+++ b/src/components/RichLayout/index.js
@@ -14,12 +14,10 @@ class RichLayout extends React.Component {
   render () {
     return (
       <div className={style.richLayout}>
-        <Header />
-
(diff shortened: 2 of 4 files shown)


src/components/RichLayout/style.css
```diff
@@ -0,0 +1,7 @@
+.richLayout {
```
maxammann: should use kebab-case formatting like: `.rich-layout {`

post-css takes care of making it available as richLayout to other modules.

src/components/RichLayout/index.js
```diff
@@ -4,13 +4,16 @@ import Layout from 'components/Layout'
...
     className: PropTypes.string
   }
 
   render () {
-    return (<div>
+    return (
+      <div className={style.richLayout}>
         <Header />
 
         <Layout className={this.props.className}>
```
Schedulaar: We have too much margin between the header and the breadcrumbs. (Because remaining space is equally distrubeted between `<Header />` `<Layout />` and `<Footer />`. To solve this, wrap `<Header />` and `<Layout />` in a seperate div, so we have
``` javascript
<div className={style.richLayout}>
 <div>
  <Header />
  <Layout className={this.props.className}>{this.props.children}</Layout>
 </div>
 <Footer />
</div>
```

src/components/RichLayout/Header.css
```diff
@@ -7,7 +7,7 @@
 }
 
 .spacer {
-  margin-bottom: var(--header-height);
+  padding-bottom: var(--header-height);
```
Schedulaar: On line 46 change `margin-bottom: var(--header-height-small);` to `padding-bottom: var(--header-height-small);`

klinzo: done

src/components/RichLayout/index.js
klinzo: done

PR #48, opened 2017-11-19
WEBAPP-100: Make Breadcrumbs selectable

klinzo:


src/components/Content/Breadcrumb.css
```diff
@@ -9,13 +9,11 @@
   color: #e8e8e8;
   font-size: 16px;
   content: ' > ';
-  user-select: none;
 }
 
 .level {
   color: #9c9c9c;
   font-size: 15px;
-  user-select: none;
```
maxammann: I think we can remove the `user-select: none;` from LanguageFlyout/style.css too

Schedulaar: no, we shouldn't. Buttons should never be selectable

PR #50, opened 2017-11-23
WEBAPP-130: Disable service-worker caching

maxammann:


src/main.js
```diff
@@ -100,11 +100,6 @@ ReactDOM.render(App, container)
 // Sets the splash to hidden when the page is rendered
 document.getElementById('splash').className += ' splash-hidden'
 
-if (navigator.serviceWorker) {
-  navigator.serviceWorker.register('/sw.js', {scope: '/'})
-    .catch((err) => console.error('Unable to register service worker.', err))
-}
```
Schedulaar: Maybe we need to unregister the current ServiceWorker... Otherwise the service worker may still be in place. 🤔

maxammann: I think sw.js is no longer available to it should fail to start

Schedulaar: I think sw.js is probably cached by the service worker itself :D

Schedulaar: https://stackoverflow.com/questions/33704791/how-do-i-uninstall-a-service-worker

maxammann: lool you could be right 😭 
k Ill add code to unregister all

maxammann: done

PR #51, opened 2017-11-25
WEBAPP-103: Remove margins from caption

klinzo:
we still want our captiontext  to be centered, right?


src/components/Content/Caption.css
```diff
@@ -1,4 +1,3 @@
 .caption {
-  margin: 50px 0;
   text-align: center;
```
Schedulaar: Do we have any margin now ? :D

Schedulaar: Maybe you can leave the `margin: 50px 0;` in there (it looks good on Desktop)
And add (at the bottom): ```css
@media screen and (max-width: 750px) {
  .caption {
    margin: 10px 0;
  }
```
Play a little bit with the 10px, maybe do some more, maybe also adjust the font-size, in that media query.

Schedulaar: .

PR #52, opened 2017-11-25
WEBAPP-1: Redesign Header

Schedulaar:
This is a preview. Please check if you're content with the header. I think I'll extract the swiping part (between extras, categories and events) of the header to another task, since it'll involve some more complexity, Because we need to have multiple 'views' loaded at the same time (categories page should stay at the exact same page when swiping to the events view and therefore changing the url to the events url, then swiping back to the categories view)


src/components/RichLayout/Header.js
```diff
@@ -1,87 +1,94 @@
...
           </div>
+          {
+            this.isMenuEnabled() &&
+            <div className={style.menuItems}>
+              { this.isExtrasEnabled() && <Link href={this.props.navigation.home}>{t('common:extras')}</Link> }
+              <Link href={this.props.navigation.home}>{t('common:categories')}</Link>
+              { this.isEventsEnabled() && <Link href={this.props.navigation.events}>{t('common:news')}</Link> }
+            </div>
+          }
         </div>
```
Schedulaar: I need to have some logic, to detect wether Categories, Extras or Events is currently shown, but I am unsure how to do it. Do I really need to get router.path and check the url? 🤔

maxammann: https://github.com/FormidableLabs/redux-little-router#link
You can do:
<Link
  href='/wat'
  className='normal-link' 
  activeProps={{ className: 'active-link' }}
>
  Wat
</Link>`

:)
`

locales/ar/common.json
```diff
@@ -1,6 +1,9 @@
 {
   "location": "الموقع",
   "date": "تاريخ",
+  "extras": "إضافات",
+  "categories": "الفئات",
+  "news": "الأخبار",
```
maxammann: Are those professional translations?

maxammann: Farsi translations are missing :)

Schedulaar: Yep, These are the constants sheet, Farsi is missing there as well

package.json
```diff
@@ -91,6 +91,7 @@
     "postcss-selector-matches": "^3.0.1",
     "postcss-selector-not": "^3.0.1",
     "prop-types": "^15.6.0",
+    "react-headroom": "^2.2.2",
```
maxammann: +1 for this :D let's see how this behaves :D

locales/ar/common.json
maxammann: Farsi has been added recently, french aswell :)

src/components/RichLayout/Header.js
Schedulaar: I think the active class will only take effect, if the href *exactly* matches the current url (so if the url is `/augsburg/de/welcome`, then the active-class will be removed, but I still want the `Categories` to be printed bold.) But I'll give it a try...

locales/ar/common.json
Schedulaar: Farsi  is done, french will be added in WEBAPP-133

PR #55, opened 2017-11-28
WEBAPP-99: Redesign language flyout 

TheNewCivilian:


src/components/Layout/index.js
```diff
@@ -12,7 +12,7 @@ class Layout extends React.Component {
   componentWillMount () {
     WebFont.load({
       google: {
-        families: ['El Messiri:300,400,700', 'Raleway:300,400,400i,700,700i']
+        families: ['El Messiri:300,400,700', 'Raleway:300,400,400i,600,700,700i,800']
```
Schedulaar: do we need the Raleway font in 800 font-weight? Fonts are expensive (in terms of size), we shouldn't have to have too many loaded.

TheNewCivilian: Yeah you're right. In the end i didn't use the 800 in my solution. I'll correct that. In case it's a big hitch to load 600, we could experiment with changing only the font-size.

src/components/LanguageFlyout/style.css
```diff
@@ -16,6 +16,7 @@
   min-width: 90px;
   max-width: 120px;
   flex: 1 1 90px;
+  color: #888;
```
Schedulaar: Please change `color: #888;` to `color: var(--text-secondary-color);`

src/components/LanguageFlyout/style.css
```diff
@@ -25,11 +26,13 @@
 }
 
 .element-active {
-  font-weight: 900;
+  color: #000;
```
Schedulaar: Please change color: #888; to color: var(--text-color);

src/components/LanguageFlyout/style.css
```diff
@@ -25,11 +26,13 @@
...
+  font-weight: 700;
 }
 
 .element:hover {
-  background-color: rgba(251, 218, 24, 0.5);
+  color: #000;
```
Schedulaar: Please change color: #888; to color: var(--text-color);

src/components/LanguageFlyout/style.css
```diff
@@ -16,6 +16,7 @@
   min-width: 90px;
```
Schedulaar: Please merge dev into your branch and add `@import url('../Common/Colors.css');` at the top of this file.

src/components/Layout/index.js
```diff
@@ -12,7 +12,7 @@ class Layout extends React.Component {
   componentWillMount () {
     WebFont.load({
       google: {
-        families: ['El Messiri:300,400,700', 'Raleway:300,400,400i,700,700i']
+        families: ['El Messiri:300,400,700', 'Raleway:300,400,400i,600,700,700i']
```
maxammann: We can also think about hadding intelligent lodaing on font here. If language is arabic load El Messiri etc.

src/components/LanguageFlyout/style.css
TheNewCivilian: Done!

src/components/Layout/index.js
TheNewCivilian: Jap. I would try to do this with WEBAPP-96

PR #56, opened 2017-11-28
WEBAPP-106: Introduce colors.css

Schedulaar:


maxammann:
See discussion in #design-marketing

PR #57, opened 2017-12-02
WEBAPP-138: Test components

klinzo:
for now there are just to tests for Breadcrumb.js and ContentList.js, this pull request is mainly to discuss testing

### DIFF
--- a/src/components/Content/__tests__/Breadcrumb.spec.js
+++ b/src/components/Content/__tests__/Breadcrumb.spec.js
@@ -1,0 +1,80 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import { shallow } from 'enzyme'
+import { Provider } from 'react-redux'
+
+import { Breadcrumb } from '../Breadcrumb'
+import PageModel from '../../../endpoints/models/PageModel'
+import Hierarchy from 'routes/LocationPage/Hierarchy'
+import LocationModel from '../../../endpoints/models/LocationModel'
+import store from '../../../store'
+
+describe('Breadcrumb', () => {
+  const page4 = new PageModel({
+    id: 'page4',
+    numericId: 4,
+    title: 'Page4',
+    parent: 2
+  })
+
+  const page2 = new PageModel({
+    id: 'page2',
+    numericId: 2,
+    title: 'Page2',
+    parent: 0,
+    children: [page4]
+  })
+
+  const page3 = new PageModel({
+    id: 'page3',
+    numericId: 3,
+    title: 'Page3',
+    parent: 0
+  })
+
+  const rootPage = new PageModel({
+    id: 'rootId',
+    numericId: 0,
+    title: 'augsburg',
+    children: [page2, page3]
+  })
+
+  const locations = [
+    new LocationModel({
+      name: 'Augsburg',
+      code: 'augsburg'
+    }),
+    new LocationModel({
+      name: 'Other city',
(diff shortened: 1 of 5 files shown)


src/components/Content/__tests__/Breadcrumb.spec.js
```diff
@@ -0,0 +1,80 @@
...
+      name: 'Other city',
+      code: 'otherCity'
+    })]
+
+  let hierarchy = new Hierarchy('augsburg/page2/page4')
+  hierarchy.build(rootPage)
+
+  test('should render', () => {
+    const wrapper = renderer.create(
+      <Provider store={store}>
```
maxammann: You shouldn't need the Provider here as you don't use the connect() function.

src/components/Content/__tests__/Breadcrumb.spec.js
```diff
@@ -0,0 +1,80 @@
...
+          hierarchy={hierarchy}
+          location={'augsburg'}
+          language={'de'}
+          locations={locations}/>
+      </Provider>)
+
+    expect(wrapper.toJSON()).toMatchSnapshot()
+  })
+
+  test('getFormattedTitle', () => {
```
maxammann: This line should complete "Breadcrumb " to a full sentence. Like "Breadcrumb should render". So split the two cases you're testing below in two and describe the test.

maxammann: first one would be: "should return 'Augsburg' as formatted title for the root page"

src/components/Content/__tests__/ContentList.spec.js
```diff
@@ -0,0 +1,46 @@
...
+    {page: page4, url: '/augsburg/de/page3/page4'}
+  ]
+
+  test('should render', () => {
+    const wrapper = renderer.create(
+      <Provider store={store}>
+        <ContentList pages={pages} />
+      </Provider>)
+
+    expect(wrapper.toJSON()).toMatchSnapshot()
```
maxammann: I'm ok with a snapshot test here. But it would be nice if we could test for more specific stuff. You can also change the implmentation to make it more testable. You could test for example for the amount of pages which are displayed.

src/components/Content/__tests__/ContentList.spec.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
```
maxammann: 🔥 You probably will get here merge conflicts with @Schedulaar s task 🔥

src/components/Content/__tests__/Breadcrumb.spec.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react'
```
maxammann: 🔥 You probably will get here merge conflicts with @Schedulaar s task 🔥

src/components/Content/__tests__/__snapshots__/Breadcrumb.spec.js.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Breadcrumb should render 1`] = `
+<div
```
maxammann: This snapshot looks a little bit empty. Are you sure you're rendering it right?

maxammann:
What should we do with this PR?

PR #58, opened 2017-12-02
WEBAPP-112: Refactor Content Component

Schedulaar:


maxammann:
We also have major merge conflicts with #59. Maybe we should wait with this task.

maxammann:
What should we do with this PR?

PR #59, opened 2017-12-03
WEBAPP-50: Redesign events

klinzo:
I would appreciate suggestions for improvement, mainly regarding design, also some functionality is still missing and I don't now how to deal with the fact that some events are not available  in all languages, maybe we should always load the events in the language and german to fill missing events with the german ones


src/components/Content/Event.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import EventModel from '../../endpoints/models/EventModel'
+import RemoteContent from './RemoteContent'
+
+import style from './Event.css'
+import EventPlaceholder1 from '../../components/Content/assets/EventPlaceholder1.jpg'
```
maxammann: There imports can be simplified to:

`'./assets/EventPlaceholder1.jpg'`

src/components/Content/EventsList.js
```diff
@@ -0,0 +1,79 @@
...
+import EventModel from '../../endpoints/models/EventModel'
+import Caption from './Caption'
+
+import style from './EventsList.css'
+
+import EventPlaceholder1 from '../../components/Content/assets/EventPlaceholder1.jpg'
+import EventPlaceholder2 from '../../components/Content/assets/EventPlaceholder2.jpg'
+import EventPlaceholder3 from '../../components/Content/assets/EventPlaceholder3.jpg'
+
+class Event extends React.Component {
```
maxammann: This looks very similar to Event.js. Are there major differences?

src/components/Content/EventsList.js
```diff
@@ -0,0 +1,79 @@
...
+        <Caption title={'Events'}/>
+        <div className={style.horizontalLine}/>
+        { this.props.events
+          ? this.props.events.map((event) =>
+            <Event key={event.event.id}
+                   event={event.event}
+                   parentUrl={this.props.url}
+                   thumbnailPlaceholder={event.thumbnailPlaceholder}
+                   language={this.props.language}/>)
+          : <div>Currently there are no events</div>
```
maxammann: `Currently there are no events` should be localized.

src/components/Content/EventsList.js
```diff
@@ -0,0 +1,79 @@
...
+      thumbnailPlaceholder: PropTypes.number.isRequired
+    })).isRequired,
+    url: PropTypes.string.isRequired,
+    language: PropTypes.string.isRequired
+  }
+
+  render () {
+    return (
+      <div className={style.list}>
+        <Caption title={'Events'}/>
```
maxammann: `Events` should be localized.

maxammann:
The thumbnails should be reduced in resolution. I think about 400pixels in width should be enough. Format should be ok but png is maybe better.

maxammann:
We also have major merge conflicts with #58. Maybe we should wait with #58 until this is done.

maxammann:
Design 👍

src/components/Content/EventsList.js
klinzo: already done, just not pushed yet

src/components/Content/EventsList.js
klinzo: already done, just not pushed yet

src/components/Content/EventsList.js
klinzo: you mean the Placeholder logic? nope, exactly the same, should it be excluded?

src/components/Content/Event.js
klinzo: copy n paste :D

src/components/Content/EventsList.js
maxammann: I meant the whole Event component. Because there is another one in Event.js. But it's a bit different. Maybe rename this to EventLink?

maxammann:
Blocked by: Question about events in multiple languages

src/components/Content/Event.js
```diff
@@ -0,0 +1,54 @@
...
+          <span className={style.identifier}>{t('common:location')}: </span>
+          <span className={style.date}>{this.props.event.event.address}</span>
+        </div>
+        <RemoteContent dangerouslySetInnerHTML={{__html: this.props.event.event.content}}/>
+      </div>
+    )
+  }
+}
+
+export default translate('common')(Event)
```
maxammann: Add withAvailableLanguageUpdater as discussed

src/components/Content/EventsList.js
```diff
@@ -0,0 +1,82 @@
...
+    url: PropTypes.string.isRequired,
+    language: PropTypes.string.isRequired
+  }
+
+  render () {
+    const {t} = this.props
+    return (
+      <div className={style.list}>
+        <Caption title={t('common:news')}/>
+        <div className={style.horizontalLine}/>
```
maxammann: This should be at the start and at the end of the list.

src/routes/LocationPage/index.js
```diff
@@ -93,13 +89,18 @@ class LocationPage extends React.Component {
   }
 }
 
+const mapLanguageToUrl = (location, language, id) => (
+  id !== '' ? `/${location}/${language}/redirect?id=${id}` : `/${location}/${language}`
```
maxammann: Use a constant here instead of an empty string. What does empty string mean? Maybe undefined is better here.

PR #61, opened 2017-12-06
WEBAPP-96: New Font for arabic-languages + Content with latin letters use Open-Sans + Fonts load on first use

TheNewCivilian:

### DIFF
--- a/src/components/Common/Helper.css
+++ b/src/components/Common/Helper.css
@@ -1,3 +1,9 @@
+:root {
+  --standard-fonts: 'Raleway', 'Lateef', 'El Messiri', sans-serif;
+  --standard-fonts-size-adjust: 0.55;
+  --content-fonts: 'Open+Sans', 'Lateef', 'El Messiri', sans-serif;
+}
+
 @define-mixin remove-a {
   color: inherit;
   text-decoration: none;
--- a/src/components/LanguageFlyout/style.css
+++ b/src/components/LanguageFlyout/style.css
@@ -7,8 +7,8 @@
   flex-flow: row wrap;
   justify-content: center;
   color: var(--text-color);
-  font-family: 'Raleway', 'Lateef', 'El Messiri', sans-serif;
-  font-size-adjust: 0.55;
+  font-family: var(--standard-fonts);
+  font-size-adjust: var(--standard-fonts-size-adjust);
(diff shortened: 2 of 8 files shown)


src/components/Layout/index.js
```diff
@@ -3,18 +3,41 @@ import PropTypes from 'prop-types'
...
+        break
+      }
+      default: {
+        WebFont.load({
+          google: {
+            families: ['Raleway:300,400,400i,600,700,700i', 'Open+Sans:400']
+          }
+        })
+      }
+    }
```
maxammann: Can be simplified to:
```
    const arabicFonts = ['Lateef:400']
    const latinFonts =  ['Raleway:300,400,400i,600,700,700i', 'Open+Sans:400']
    const families = {
      de: latinFonts,
      ar: arabicFonts,
      fa: arabicFonts,
      ku: arabicFonts,
      ti: ['El Messiri:300,400,700']
    }

    WebFont.load({
      google: {
        families: families[i18next.language] || latinFonts
      }
    })
```

Not sure though :/

src/components/LanguageFlyout/style.css
```diff
@@ -7,7 +7,8 @@
   flex-flow: row wrap;
   justify-content: center;
   color: var(--text-color);
-  font-family: 'Raleway', 'El Messiri', sans-serif;
+  font-family: 'Raleway', 'Lateef', 'El Messiri', sans-serif;
```
maxammann: Shouldn't be needed because RichLayout already has this definition

src/components/RichLayout/Header.css
```diff
@@ -12,7 +12,8 @@
   box-sizing: border-box;
   align-items: center;
   box-shadow: 0 2px 5px -3px rgba(0, 0, 0, 0.2);
-  font-family: 'Raleway', 'El Messiri', sans-serif;
+  font-family: 'Raleway', 'Lateef', 'El Messiri', sans-serif;
```
maxammann: Should be moved to the RickLayout style.css

src/components/Common/Helper.css
```diff
@@ -1,3 +1,9 @@
+:root {
```
maxammann: Will be changed in WEBAPP-112

PR #65, opened 2017-12-13
WEBAPP-154: Simplify language change

klinzo:
we don't need to save the availableLanguages in state


src/modules/language/actions/setLanguageChangeUrls.js
```diff
@@ -1,5 +1,17 @@
 import { createAction } from 'redux-actions'
 
 export const setLanguageChangeUrls = createAction('SET_LANGUAGE_CHANGE_URLS',
-  (languageChangeUrls) => (languageChangeUrls)
-)
+  (mapLanguageToUrl, languages, availableLanguages = undefined) => (
```
maxammann: This is the same as:

```
 (mapLanguageToUrl, languages, availableLanguages = undefined) => {
    let a = (language) => mapLanguageToUrl(language.code, availableLanguages[language.code])
    let b = (language) => mapLanguageToUrl(language.code, undefined)

    return availableLanguages
      // languageChange of a specific page/event with ids in availableLanguages
      ? languages.reduce((accumulator, language) => {
        return ({
          ...accumulator,
          [language.code]: a(language)
        })
      }, {})
      : languages.reduce((accumulator, language) => {
        return ({
          ...accumulator,
          [language.code]: b(language)
        })
      }, {})
  }
```

Which is the same as:
```
(mapLanguageToUrl, languages, availableLanguages = undefined) => {
    let languageMapper
    if (availableLanguages) {
      languageMapper = (language) => mapLanguageToUrl(language.code, availableLanguages[language.code])
    } else {
      languageMapper = (language) => mapLanguageToUrl(language.code, undefined)
    }

    return languages.reduce((accumulator, language) => ({
      ...accumulator,
      [language.code]: languageMapper(language)
    }), {})
```

Language changing still works, so I'm pretty sure this is correct.

src/routes/events/containers/EventsPage.js
```diff
@@ -26,6 +29,43 @@ class EventsPage extends React.Component {
     return `/${this.props.location}/${this.props.language}/events`
   }
 
+  mapLanguageToUrl = (language, id) =>
+    id ? `/${this.props.location}/${language}/events/${id}`
+      : `/${this.props.location}/${language}/events`
+
+  componentDidMount () {
+    if (this.props.path && this.props.events) {
```
maxammann: Can be simplified to:
```
    let availableLanguages = {}

    if (this.props.path && this.props.events) {
      // specific event
      const event = this.props.events.find(
        (event) => event.id.toString() === this.props.path.replace('/', '') // Works because we can not have nested events
      )
      availableLanguages = event.availableLanguages
    }

    this.props.dispatch(setLanguageChangeUrls(this.mapLanguageToUrl, this.props.languages, availableLanguages))
```

src/modules/language/actions/setLanguageChangeUrls.js
maxammann: Which can be simplified to:
```
(mapLanguageToUrl, languages, availableLanguages = {}) => {
    const languageMapper = (language) => mapLanguageToUrl(language.code, availableLanguages[language.code])

    return languages.reduce((accumulator, language) => ({
      ...accumulator,
      [language.code]: languageMapper(language)
    }), {})
  }
```

Because of the simplification in EventsPage.js

maxammann: If availableLanguages = undefined -> availableLanguages[language.code] = undefined

src/routes/events/containers/EventsPage.js
```diff
@@ -26,6 +29,43 @@ class EventsPage extends React.Component {
...
+      )
+      this.props.dispatch(setLanguageChangeUrls(this.mapLanguageToUrl, this.props.languages, event.availableLanguages))
+    } else {
+      // all events
+      this.props.dispatch(setLanguageChangeUrls(this.mapLanguageToUrl, this.props.languages))
+    }
+  }
+
+  componentWillReceiveProps (nextProps) {
+    if (nextProps.events !== this.props.events || nextProps.path !== this.props.path) {
```
maxammann: Simplify this like componentDidMount

src/routes/events/containers/EventsPage.js
```diff
@@ -26,6 +29,43 @@ class EventsPage extends React.Component {
...
+        (event) => event.id.toString() === this.props.path.replace('/', '')
+      )
+      this.props.dispatch(setLanguageChangeUrls(this.mapLanguageToUrl, this.props.languages, event.availableLanguages))
+    } else {
+      // all events
+      this.props.dispatch(setLanguageChangeUrls(this.mapLanguageToUrl, this.props.languages))
+    }
+  }
+
+  componentWillReceiveProps (nextProps) {
```
maxammann: Add a comment why we have to use componentWillReceiveProps here :P

src/modules/language/actions/setLanguageChangeUrls.js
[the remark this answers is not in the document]
klinzo: done

klinzo: done, good suggestions

src/routes/events/containers/EventsPage.js
klinzo: done

src/routes/events/containers/EventsPage.js
klinzo: done

src/routes/events/containers/EventsPage.js
klinzo: simplified it a little bit, but if we would simplify this further, we would have to dispatch twice

PR #66, opened 2017-12-13
WEBAPP-134:  Make footer stick to the bottom on the LandingPage

klinzo:


src/routes/landing/components/FilterableLocationSelector.css
```diff
@@ -1,27 +1,10 @@
-@import url('../../../modules/common/constants/colors.css');
-
 :root {
   --hover-color: rgba(0, 0, 0, 0.1);
   --item-height: 30px;
   --list-margin: 30px;
 }
 
-.language-list-item {
```
maxammann: 👍

src/routes/landing/components/FilterableLocationSelector.css
```diff
@@ -1,27 +1,10 @@
...
-  color: inherit;
-  text-decoration: inherit;
-
-  &:hover {
-    color: inherit;
-    text-decoration: inherit;
-    transition: background-color 0.5s ease;
-    background-color: var(--hover-color);
-  }
+.top-spacing {
```
maxammann: I think this should still be part of LandingPage. This would make the component more reusable.

src/modules/app/components/Layout.css
```diff
@@ -1,10 +1,6 @@
 @import url('../../common/constants/colors.css');
 @import url('../../common/constants/fonts.css');
 
-.top-spacing {
-  padding-top: 22px;
-}
-
 .content {
```
Schedulaar: Since this removes the space between Header and breadcrumbs, the breadcrumbs now stick to the header. But now you could add sth. like `padding: 10px 0;` to the Breadcrumbs container, which is good anyway, because that way we increase the touch-target for scrolling the breadcrumbs when they're too long.

src/routes/landing/components/FilterableLocationSelector.css
```diff
@@ -1,27 +1,10 @@
-@import url('../../../modules/common/constants/colors.css');
-
 :root {
   --hover-color: rgba(0, 0, 0, 0.1);
   --item-height: 30px;
   --list-margin: 30px;
```
Schedulaar: you need to move the --item-height definition to LocationSelector.css and you can omit --list-margin, I think.

src/routes/landing/components/LocationSelector.css
```diff
@@ -0,0 +1,21 @@
...
+    color: inherit;
+    text-decoration: inherit;
+    transition: background-color 0.5s ease;
+    background-color: var(--hover-color);
+  }
+}
+
+.language-list-parent {
+  height: var(--item-height);
+  line-height: var(--item-height);
```
Schedulaar: --item-height is not yet defined here (see above).

maxammann: There are no warning when something is not defined in css?

maxammann: We should take a look at: https://github.com/styled-components/styled-components
Makes more sense than having real .css files which are then compiled to js modules. We could write all our css directly in js, which also provies a lot more possibilities.

src/modules/app/components/Layout.css
klinzo: i think we never have breadcrumbs as part of Layout, just as part of RichLayout

src/routes/landing/components/FilterableLocationSelector.css
klinzo: done

src/modules/app/components/Layout.css
klinzo: ok, you were right, done :D

src/routes/landing/components/FilterableLocationSelector.css
klinzo: I agreed with @maxammann that this is ok

src/routes/landing/components/LocationSelector.css
```diff
@@ -10,7 +14,7 @@
     color: inherit;
     text-decoration: inherit;
     transition: background-color 0.5s ease;
-    background-color: var(--hover-color);
+    background-color: var(--background-accent-color);
```
klinzo: this color is a bit lighter than the one we used before (rgba(0, 0, 0, 0.1)), but since we agreed to only use some defined colors this should be ok

maxammann:
I think this can be merged, right?

klinzo:
yeah, me too, but if someone hast to approve :D

maxammann:
oops :P

PR #71, opened 2017-12-24
WEBAPP-163: Change PageModel to a simple map instead of tree

klinzo:


src/modules/endpoint/models/CategoryModel.js
```diff
@@ -0,0 +1,65 @@
...
+    return this._content
+  }
+
+  get parent () {
+    return this._parent
+  }
+
+  get children () {
+    return this._children
+  }
```
Schedulaar: In my opinion we don't need children as a property of a CategoryModel anymore (that would be an optimization) - since we can always replace that by 
```javascript
 allCategores.filter(category => category.parent === myCategory.id)
```

src/routes/categories/containers/CategoriesPage.js
```diff
@@ -73,35 +102,55 @@ class CategoriesPage extends React.Component {
   }
 
   render () {
-    const url = this.getParentPath()
-    const hierarchy = new Hierarchy(this.props.path)
+    const category = CategoryModel.getCategoryByPath(this.props.categories, this.props.path)
+    if (!category) {
+      // todo show an error that the page has not been found, redirect to the notFound route
+    }
+    const children = this.props.categories.filter(_category => category.children.includes(_category.id))
```
Schedulaar: This is equivalent to 
```javascript
const children = this.props.categories.filter(_category => _category.parent === category.id)
```

Schedulaar:
@klinzo Is the search page working in branch WEBAPP-145

src/routes/categories/containers/CategoriesPage.js
```diff
@@ -73,35 +102,55 @@ class CategoriesPage extends React.Component {
...
+    setLanguageChangeUrls(urls, languages, availableLanguages)
+  ),
+  dispatchRedirectToUrl: (url) => dispatch(replace(url))
+})
+
 const mapStateToProps = (state) => ({
   language: state.router.params.language,
   location: state.router.params.location,
-  path: state.router.params['_'] // _ contains all the values from *
+  path: state.router.params['_'], // _ contains all the values from *
```
Schedulaar: You probably want to use `path: state.router.pathname` now which contains not only the values from *, but the whole path.

src/modules/endpoint/models/CategoryModel.js
```diff
@@ -0,0 +1,65 @@
...
+  get order () {
+    return this._order
+  }
+
+  get availableLanguages () {
+    return this._availableLanguages
+  }
+
+  static getCategoryByPath (categories, path = '') {
+    return categories.find(category => category.url === encodeURI(path).toLowerCase())
```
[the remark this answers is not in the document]
Schedulaar: We rather need some url-normalizing here: Calling `/augsburg/de/` and `/augsburg/de` should both work

src/modules/endpoint/models/CategoryModel.js
klinzo: you're right, i'll remove it

maxammann:
⚠️ Tests for PageModel and pages endpoint need to be done in this task.

src/modules/endpoint/models/CategoryModel.js
maxammann: We'll need the children to render a list of children. And we already have those there. So why not use them? We'd actually have to test more code.

src/modules/endpoint/models/CategoryModel.js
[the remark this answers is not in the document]
maxammann: Yeah we need to normalize here. Also encodeURI implementation differs from enviroment to environment. So it produces different outputs. As long as we don't have any user input we do not need to encode/decode our URIs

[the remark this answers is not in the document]
maxammann: I'm find with  getCategoryByPath. The method does not need to be named after the implementation.

src/modules/endpoint/models/CategoryModel.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  static getCategoryById (categories, id) {
+    return categories.find(category => category.id === Number(id))
+  }
```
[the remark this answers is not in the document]
maxammann: See comment about Categorie_s_Model

src/routes/categories/components/Breadcrumbs.js
```diff
@@ -0,0 +1,60 @@
...
+    categories: PropTypes.arrayOf(PropTypes.instanceOf(CategoryModel)).isRequired,
+    category: PropTypes.instanceOf(CategoryModel).isRequired,
+    locations: PropTypes.arrayOf(PropTypes.instanceOf(LocationModel)).isRequired
+  }
+
+  /**
+   *
+   * @param location
+   */
+  getLocationTitle = (location) => (this.props.locations.find(_location => location === _location.code)).name
```
maxammann: It's possible that find() returns undefined to better catch that case and return null.

src/routes/categories/components/Breadcrumbs.js
```diff
@@ -0,0 +1,60 @@
...
+
+  getBreadcrumbs (categories, currentCategory) {
+    if (currentCategory.id === this.props.category.id && currentCategory.id === 0) return []
+
+    if (currentCategory.id === 0) {
+      return [{title: this.getLocationTitle(currentCategory.title), url: currentCategory.url}]
```
maxammann: I think currentCategory.title is not guaranteed to match a location.

src/modules/endpoint/endpoints/categories.js
```diff
@@ -0,0 +1,39 @@
...
+import EndpointBuilder from '../EndpointBuilder'
+
+import CategoryModel from '../models/CategoryModel'
+
+export default new EndpointBuilder('categories')
+  .withUrl('https://cms.integreat-app.de/{location}/{language}/wp-json/extensions/v0/modified_content/pages?since=1970-01-01T00:00:00Z')
+  .withStateMapper().fromArray(['location', 'language'], (state, paramName) => state.router.params[paramName])
+  .withMapper((json, urlParams) => {
+    const baseUrl = `/${urlParams.location}/${urlParams.language}`
+    let categories = json.filter((category) => category.status === 'publish')
```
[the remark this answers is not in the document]
klinzo: done

src/modules/endpoint/models/CategoryModel.js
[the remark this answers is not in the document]
klinzo: we need encodeURI because our cms provides as url for arabic and russian charachters UTF-8 encoded strings while our router gives us decoded url, i.e. "%d8%a7%d9%84%d8%a7%d8%b3%d8%aa%d9%82%d8%a8%d8%a7%d9%84-%d8%a7%d9%84%d8%a3%d9%88%d9%84%d9%8a" vs. الاستقبال-الأولي

src/modules/endpoint/models/CategoryModel.js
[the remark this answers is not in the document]
klinzo: kind of, but otherwise we would have this functions very often, so i just put them into CategoryModel to reduce code duplication, and i didn't find a class where it fits better

src/routes/categories/components/Breadcrumbs.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    let breadcrumbs = this.getBreadcrumbs(categories, CategoryModel.getCategoryById(categories, currentCategory.parent))
```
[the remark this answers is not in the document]
klinzo: done

src/routes/categories/components/Breadcrumbs.js
```diff
@@ -0,0 +1,60 @@
...
+
+    if (currentCategory.id !== this.props.category.id) {
+      breadcrumbs.push({title: currentCategory.title, url: currentCategory.url})
+    }
+    return breadcrumbs
+  }
```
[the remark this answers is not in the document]
maxammann: You don't believe in the force of the npm repository?! You should feel ashamed!
We can use this library: https://www.npmjs.com/package/un-flatten-tree
we only need one half of it, but webpack will optimize out the other half.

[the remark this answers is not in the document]
maxammann: 👍

src/routes/categories/components/Breadcrumbs.js
```diff
@@ -0,0 +1,60 @@
...
+
+  render () {
+    const breadcrumbs = this.getBreadcrumbs(this.props.categories, this.props.category)
```
maxammann: getBreadcrumbs can be part of Categorie_s_Model then. Ofc renamed to something like: flatten(mapCategoryToSomething) wich will follow all parents. mapCategoryToSomething will take a category and map it to the title for example.

src/routes/categories/containers/CategoriesPage.js
```diff
@@ -1,66 +1,95 @@
...
+import CategoryModel from 'modules/endpoint/models/CategoryModel'
 import LanguageModel from 'modules/endpoint/models/LanguageModel'
+import LocationModel from 'modules/endpoint/models/LocationModel'
+import { setLanguageChangeUrls } from 'modules/language/actions/setLanguageChangeUrls'
+
+import Breadcrumb from 'routes/categories/components/Breadcrumbs'
+import PdfButton from 'routes/categories/components/PdfButton'
+import Page from '../components/Page'
+import CategoryTiles from '../components/CategoryTiles'
+import TitledContentList from '../components/CategoryList'
```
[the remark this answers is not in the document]
klinzo: done

src/routes/categories/components/CategoryList.js
```diff
@@ -0,0 +1,37 @@
...
+
+import style from './CategoryList.css'
+
+/**
+ * Displays a ContentList which is a list of categories, a caption and a thumbnail
+ */
+class CategoryList extends React.Component {
+  static propTypes = {
+    parentCategory: PropTypes.instanceOf(CategoryModel).isRequired,
+    categories: PropTypes.arrayOf(PropTypes.instanceOf(CategoryModel)).isRequired
```
maxammann: This will also be simple if we have a Categorie_s_Model

src/routes/categories/containers/CategoriesPage.js
```diff
@@ -1,66 +1,95 @@
...
-    const hierarchy = new Hierarchy(path)
-    const error = hierarchy.build(this.props.pages)
-    if (error) {
-      // todo handle this error
-      return
-    }
-    this.props.dispatch(setLanguageChangeUrls(
-      this.mapLanguageToUrl, this.props.languages, hierarchy.top().availableLanguages)
+    this.props.dispatchLanguageChangeUrls(
+      this.mapLanguageToUrl, this.props.languages, CategoryModel.getCategoryByPath(this.props.categories, path).availableLanguages
```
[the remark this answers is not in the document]
klinzo: yeah sure, but render() is called before componentDidMount, so if we don't find a category in render () and go to the notFountPage, componentDidMount won't be called so we don't have to check for this here

src/routes/categories/containers/CategoriesPage.js
```diff
@@ -1,66 +1,95 @@
...
+   * @param categories The categories to search in
+   * @param id The id of the category to search for
+   * @returns {string} The url of the category
+   */
+  getRedirectUrl = (categories, id) => {
+    const category = CategoryModel.getCategoryById(categories, id)
+    if (!category) {
+      // todo show an error that the page has not been found, redirect to the notFound route
+    }
+    return `/${this.props.location}/${this.props.language}/${category.url}`
```
maxammann: When calculating this redirect url do we really have the correct language? Did you test the language change?

src/routes/categories/containers/CategoriesPage.js
klinzo: yop

src/routes/categories/containers/CategoriesPage.js
```diff
@@ -73,35 +102,55 @@ class CategoriesPage extends React.Component {
...
+      // todo show an error that the page has not been found, redirect to the notFound route
+    }
+    const children = this.props.categories.filter(_category => category.children.includes(_category.id))
 
-    // Pass data to hierarchy
-    const error = hierarchy.build(this.props.pages)
-    if (error) {
-      return <Error error={error} />
+    let Content
+    if (category.children.length === 0) {
```
maxammann: Extract this if else if else to a separate method. (No more evil let :P)

src/routes/categories/containers/CategoriesPage.js
```diff
@@ -73,35 +102,55 @@ class CategoriesPage extends React.Component {
...
+  ),
+  dispatchRedirectToUrl: (url) => dispatch(replace(url))
+})
+
 const mapStateToProps = (state) => ({
   language: state.router.params.language,
   location: state.router.params.location,
-  path: state.router.params['_'] // _ contains all the values from *
+  path: state.router.params['_'], // _ contains all the values from *
+  categoryId: state.router.query.id
```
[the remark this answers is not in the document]
klinzo: imo not, it's not that much logic for the redirect, and I find this more logic because otherwise we dispatch the languageChangeUrls in CategoriesPage, go to the redirectPage on an actual language change, and redirect back to CategoriesPage, which isn't quite logic for me.
@maxammann also said that we want to get rid of it.

but if you both agree it would be better to still have this RedirectPage, i can refactor this of course

src/routes/categories/containers/CategoriesPage.js
maxammann: 👍 this is similar to EventsPage now

src/routes/categories/containers/CategoriesPage.js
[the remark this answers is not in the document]
maxammann: I think the same, but the redirect pages will be gone as soon as my task is merged in the cms. So we can leave it for now.

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
```diff
@@ -31,20 +31,14 @@ class PdfFetcherPage extends React.Component {
...
-
-    this.fetchUrl(hierarchy.top())
+    this.fetchUrl(CategoryModel.getCategoryByPath(this.props.categories, this.props.path))
   }
 
-  addPageIdsRecursively (pageIds, parentPage) {
-    parentPage.children.forEach((page) => {
-      pageIds.push(page.numericId)
-      this.addPageIdsRecursively(pageIds, page)
+  addCategoryIdsRecursively (categoryIds, currentCategory) {
```
maxammann: Also a usecase for https://www.npmjs.com/package/un-flatten-tree

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
```diff
@@ -31,20 +31,14 @@ class PdfFetcherPage extends React.Component {
...
   componentWillMount () {
-    const hierarchy = new Hierarchy(this.props.path)
-
-    const error = hierarchy.build(this.props.pages)
-    if (error) {
-      return <Error error={error}/>
-    }
-
-    this.fetchUrl(hierarchy.top())
+    this.fetchUrl(CategoryModel.getCategoryByPath(this.props.categories, this.props.path))
```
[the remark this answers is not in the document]
klinzo: done

src/routes/search/containers/SearchPage.js
```diff
@@ -36,57 +34,35 @@ class SearchPage extends React.Component {
...
-  getParentPath () {
-    return `/${this.props.location}/${this.props.language}`
-  }
-
-  acceptPage (page) {
-    let title = page.title.toLowerCase()
-    let content = page.content
+  acceptCategory (category) {
+    let title = category.title.toLowerCase()
+    let content = category.content
```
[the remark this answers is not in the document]
klinzo: done

src/routes/categories/containers/CategoriesPage.js
klinzo: i should  have read this before I removed the baseUrl from CategoryModel again....

src/routes/categories/containers/CategoriesPage.js
```diff
@@ -1,66 +1,95 @@
...
+    location: PropTypes.string.isRequired,
     language: PropTypes.string.isRequired,
     path: PropTypes.string,
-    pages: PropTypes.instanceOf(PageModel).isRequired
-  }
-
-  getParentPath () {
-    return `/${this.props.location}/${this.props.language}`
+    categoryId: PropTypes.string,
+    dispatchLanguageChangeUrls: PropTypes.func.isRequired,
```
maxammann: name these just like the functions they are calling: setLanguageChangeUrls

src/modules/endpoint/endpoints/categories.js
```diff
@@ -0,0 +1,39 @@
...
+    // Set children
+    forEach(categories, category => {
+      if (category.id === 0) return
+      const parent = find(categories, _category => _category.id === category.parent)
+      if (parent) {
+        parent.addChild(category.id)
+      }
+    })
+
+    return categories
```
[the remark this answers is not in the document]
klinzo: okay, so we would have two models? CategoryModel and CategoriesModel?

src/modules/endpoint/endpoints/categories.js
```diff
@@ -0,0 +1,39 @@
+import { find, forEach } from 'lodash/collection'
+
+import EndpointBuilder from '../EndpointBuilder'
+
+import CategoryModel from '../models/CategoryModel'
+
```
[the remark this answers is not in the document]
klinzo: okay

src/routes/categories/components/Breadcrumbs.js
klinzo: done

src/routes/categories/containers/CategoriesPage.js
klinzo: yeah language change works fine

src/routes/categories/containers/CategoriesPage.js
klinzo: done

src/routes/categories/components/Breadcrumbs.js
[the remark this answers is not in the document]
klinzo: I think it's not necessary to use an external library for this, take a look at my changes in CategoriesModel

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
klinzo: we also need recursion here if we have a flattened tree

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
```diff
@@ -74,16 +69,18 @@ class PdfFetcherPage extends React.Component {
...
 
-    if (!PdfFetcherPage.isRootCategory(category)) {
+    if (category.id !== 0) {
       categoryIds.push(category.id)
     }
-    this.addCategoryIdsRecursively(categoryIds, category)
-
+    this.addCategoryIdsRecursively(categoryIds, children)
+    console.log(categoryIds)
+    console.log(this.props.fetchUrl)
```
Schedulaar: remove these logs

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
```diff
@@ -154,7 +151,7 @@ class PdfFetcherPage extends React.Component {
 const mapStateToProps = (state) => ({
   location: state.router.params.location,
   language: state.router.params.language,
-  path: state.router.params['_'] // _ contains all the values from *
+  fetchUrl: state.router.query.url
```
Schedulaar: Why did you change this to query.url ? 🤔

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
```diff
@@ -31,14 +31,18 @@ class PdfFetcherPage extends React.Component {
...
-  addCategoryIdsRecursively (categoryIds, currentCategory) {
-    currentCategory.children.forEach((id) => {
-      const child = CategoryModel.getCategoryById(this.props.categories, id)
+  addCategoryIdsRecursively (categoryIds, children) {
+    children.forEach((child) => {
       categoryIds.push(child.id)
-      this.addCategoryIdsRecursively(categoryIds, child)
+      this.addCategoryIdsRecursively(categoryIds, this.props.categories.getChildren(child.url))
     })
   }
```
Schedulaar: With "un-flatten-tree" this could be refactored to
```javascript
getCategoryIds (category){
  const ids = flatten(
       [category],
       _category => this.props.categories.getChildren(_category))
     .map(_category => _category.id)
  if(category.id === 0) {
    ids.shift()
  }
  return ids
}
```

src/modules/endpoint/models/CategoryModel.js
```diff
@@ -1,30 +1,25 @@
-class PageModel {
-  constructor ({id, numericId, title = '', parent = 0, content = '', thumbnail = null, order = 0, children = [], availableLanguages = {}}) {
+class CategoryModel {
+  constructor ({ id, url, title = '', parent = -1, content = '', thumbnail = null, order = 0, availableLanguages = {} }) {
```
Schedulaar: Do we need -1 as default parameter of parent ?

src/routes/categories/containers/CategoriesPage.js
```diff
@@ -1,107 +1,150 @@
...
+   * b) table with categories
+   * c) list with categories
+   * @param category The current category
+   * @return {*} The content to be displayed
+   */
+  getContent (category) {
+    if (!category) {
+      // todo show an error that the page has not been found, redirect to the notFound route
+      return
+    }
```
Schedulaar: You can remove this check, if you apply my suggestion below.

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
```diff
@@ -31,20 +31,18 @@ class PdfFetcherPage extends React.Component {
...
-
-    const error = hierarchy.build(this.props.pages)
-    if (error) {
-      return <Error error={error}/>
+    const category = this.props.categories.getCategoryByUrl(this.props.fetchUrl)
+    if (category) {
+      this.fetchUrl(category)
+    } else {
+      // todo display an error
     }
```
Schedulaar: You can remove this else block, because if we don't call fetchUrl() our state will always be {pdf: '', loading: false}, and therefore the render() method will return an Error-component.

src/modules/endpoint/models/CategoriesModel.js
```diff
@@ -0,0 +1,79 @@
+import normalizeUrl from 'normalize-url'
+
+/**
+ * Contains a Map [string -> CategoryModel] and some helper functions
+ */
+class CategoriesModel {
```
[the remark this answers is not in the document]
klinzo: I think we agreed on a Map? @maxammann is implementing some changes in the cms, so that availableLanguages and parent is also returned as urls instead of the ids that are used right now. So if these changes are applied, we also have direct url mapping there. converting the map to an array will then only be necessary anymore for getting all children, so imo the Map is the right implementation. Regarding the naming I agree, CategoriesContainer would be more appropriate

src/modules/endpoint/models/CategoriesModel.js
```diff
@@ -0,0 +1,79 @@
...
+   * @param categories
+   */
+  constructor (categories = []) {
+    categories.forEach(category => {
+      if (category.id !== 0) {
+        // every category except from the root category should have a parent, so we don't have to check if it exists
+        const parentUrl = categories.find(_category => _category.id === category.parent).url
+        category.setParent(parentUrl)
+      }
+    })
```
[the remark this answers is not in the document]
klinzo: regarding ii) I agree,
for the rest I like it better to have the url as parent, expecially because this is only a workaround until the changes in the cms are applied, than we get the url anyway, but we can rename category.parent to category.parentUrl

src/modules/endpoint/models/CategoriesModel.js
```diff
@@ -0,0 +1,79 @@
...
+  getCategoryById (id) {
+    return this.categories.find(category => category.id === Number(id))
+  }
+
+  /**
+   * Returns all children of the category with the given url
+   * @param url The url
+   * @return {CategoryModel[] | undefined} The children
+   */
+  getChildren (url = '') {
```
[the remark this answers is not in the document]
klinzo: I think we should discuss this in #web-app on slack

src/modules/endpoint/models/CategoriesModel.js
```diff
@@ -0,0 +1,79 @@
...
+    const parents = []
+    let category = this.getCategoryByUrl(url)
+
+    if (category) {
+      while (category.id !== 0) {
+        category = this.getCategoryByUrl(category.parent)
+        parents.unshift(category)
+      }
+    }
+    return parents
```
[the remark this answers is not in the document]
klinzo: I think there is not that much difference between the solutions, so I think introducing a library for this is a bit exaggerated, what does @maxammann say?

src/modules/endpoint/models/CategoryModel.js
```diff
@@ -1,25 +1,19 @@
 class CategoryModel {
-  constructor ({ id, url = '', title = '', parent = 0, content = '', thumbnail = null, order = 0, availableLanguages = {} }) {
+  constructor ({ id, url, title = '', parent = -1, content = '', thumbnail = null, order = 0, availableLanguages = {} }) {
```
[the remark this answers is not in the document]
klinzo: most of the time, except for the root category which doesn't have a parent. we could also remove title then.

src/routes/categories/containers/CategoriesPage.js
klinzo: we just want to display an error, not redirect to a not found page? because disclaimer, language change and so on don't make sense if no category is found, so imo it would be better to redirect to a not found page, maybe we should do this in WEBAPP-90

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
klinzo: jop, we should discuss this on slack

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
klinzo: done

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
klinzo: done

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
klinzo: because we have a whole url now to identify a category (see comments above why url to category map is useful) and so imo it looks better with a query parameter, but maybe this is just a question of personal preference. also for me the url is just /location/language/fetch-pdf, and the query url is just a parameter

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
```diff
@@ -144,10 +135,10 @@ class PdfFetcherPage extends React.Component {
     if (this.state.loading) {
       return <div className={style.pdfFetcher}>
         <p>{t('creatingPdf')}</p>
-        <Spinner name='line-scale-party'/>
+        <Spinner name='line-scale-party' />
       </div>
     } else if (!this.state.pdf) {
-      return <Error error="errors:page.loadingFailed"/>
+      return <Error error='errors:page.loadingFailed' />
```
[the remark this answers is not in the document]
klinzo: done

src/modules/endpoint/models/CategoryModel.js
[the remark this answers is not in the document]
klinzo: removed the default for title and changed the default for parent to null

src/modules/endpoint/models/CategoriesModel.js
```diff
@@ -0,0 +1,79 @@
...
+      return this.categories.filter(_category => _category.parent === category.url)
+    }
+  }
+
+  /**
+   * Returns all (mediate) parents of the category with the given url
+   * @param url The url
+   * @return {CategoryModel[]} The parents, with the immediate parent last
+   */
+  getParents (url = '') {
```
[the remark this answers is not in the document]
klinzo: renamed to getAncestors, changed to category

src/modules/endpoint/models/CategoriesModel.js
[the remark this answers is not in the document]
klinzo: i would just keep this as it is till the changes in the cms are applied

src/modules/endpoint/models/CategoriesModel.js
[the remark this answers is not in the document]
klinzo: without library it's even shorter

src/modules/endpoint/models/CategoriesModel.js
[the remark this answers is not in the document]
Schedulaar: Afaik the url is only planned for the availableLanguages-array, but that doesn't affect our data structure, this only avoids having to redirect in order to load the payload in another language with the right ids.
But -> slack

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
Schedulaar: It's alright to me, but we shouldn't change urls too often. Since it's just the pdf-fetch-page it's not a big deal though.

src/routes/categories/containers/CategoriesPage.js
Schedulaar: Hmm.. I think the disclaimer isn't affected, if a page is not found in a specific location. Also Language change should default back to the home of the location, if theres no Id to go to. I think we shouldn't redirect to some other url, but just show the error at the page where the error occurs.

klinzo: okay, done

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
klinzo: jop, so this is okay?

src/modules/endpoint/models/CategoriesContainer.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  /**
+   * Returns all children of the given category
+   * @param category The category
+   * @return {CategoryModel[] | undefined} The children
+   */
+  getChildren (category) {
+    if (category) {
+      return this.toArray().filter(_category => _category.parentUrl === category.url)
```
Schedulaar: Don't we need to add `.sortBy(_category => _category.order)` here?

Schedulaar: Also why do you check if category is not null/undefined? We should do that before we call getChildren I think...

klinzo: I don't know, it's kind of strange but its sorted like it should, even without this sortBy (at least whenever I checked it, which are of course only samples). Should I add it anyway just to make sure?

Schedulaar: When there's no other place, where you sort by the order, then yes.

klinzo: done

src/modules/endpoint/models/CategoriesContainer.js
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * Returns all children of the given category
+   * @param category The category
+   * @return {CategoryModel[] | undefined} The children
+   */
+  getChildren (category) {
+    if (category) {
+      return this.toArray()
+        .filter(_category => _category.parentUrl === category.url)
+        .sortBy(_category => _category.order)
```
Schedulaar: Sorry, my bad :/ .sortBy is not part of ES6, you need to use
`.sort((category1, category2) => category1.order - category2.order)`

Schedulaar: Maybe you can include the sort functionality in a test of CategoriesContainer? 🤔

src/routes/search/containers/SearchPage.js
```diff
@@ -33,61 +32,43 @@ class SearchPage extends React.Component {
...
-      pages.push({url, page})
-    }
-    page.children.forEach(page => this.findPages(pages, url, page))
-  }
-
   render () {
-    const url = normalizeUrl(this.getParentPath(), {removeTrailingSlash: true})
-    const pages = []
-    this.props.pages.children.forEach(page => this.findPages(pages, url, page))
+    const categories = this.props.categories.categories.filter(category => this.acceptCategory(category))
```
Schedulaar: this gets
```javascript
const categories = this.props.categories.toArray().filter(category => this.acceptCategory(category))
```

src/modules/endpoint/endpoints/__tests__/categories.spec.js
```diff
@@ -0,0 +1,84 @@
...
+import CategoryModel from '../../models/CategoryModel'
+
+describe('categories', () => {
+  const categoriesJSON = [{
+    id: 3650,
+    permalink: {
+      url_page: 'anlaufstellen'
+    },
+    title: 'Anlaufstellen zu sonstigen Themen',
+    status: 'publish',
```
maxammann: Add a page with status 'trash' to test the filter.

src/modules/endpoint/endpoints/__tests__/disclaimer.spec.js
```diff
@@ -36,7 +36,7 @@ describe('disclaimer', () => {
 
   test('should map fetched data to models', () => {
     const disclaimerModel = disclaimer.mapResponse([pageJson])
-    expect(disclaimerModel).toEqual(new PageModel({
+    expect(disclaimerModel).toEqual(new CategoryModel({
```
maxammann: Is there already a task for DisclaimerModel?

src/modules/endpoint/models/CategoriesContainer.js
```diff
@@ -0,0 +1,72 @@
...
+ * Contains a Map [string -> CategoryModel] and some helper functions
+ */
+class CategoriesContainer {
+  /**
+   * Creates a Map [url -> category] from the categories provided,
+   * whose parent attributes are first changed from id to url
+   * @param categories CategoryModels as array
+   */
+  constructor (categories = []) {
+    this._categories = new Map(categories.map(category => ([category.url, category])))
```
maxammann: 👍

src/modules/endpoint/models/CategoriesContainer.js
```diff
@@ -0,0 +1,72 @@
...
+        .sortBy(_category => _category.order)
+    }
+  }
+
+  /**
+   * Returns all (mediate) parents of the given category
+   * @param category The category
+   * @return {CategoryModel[]} The parents, with the immediate parent last
+   */
+  getAncestors (category) {
```
maxammann: I like the refactor by @Schedulaar:
```
getAncestors(category) {
 return flatten(category, _category => this.getCategoryById(_category.parentId))
    .slice(1)
    .reverse()
}
```

src/modules/endpoint/models/__tests__/CategoriesContainer.spec.js
```diff
@@ -0,0 +1,42 @@
...
+  const categories = [
+    new CategoryModel({id: 0, url: '/augsburg/de', title: 'augsburg'}),
+    new CategoryModel({id: 20, url: '/augsburg/de/willkommen', parentId: 0, parentUrl: '/augsburg/de', title: 'willkommen'}),
+    new CategoryModel({id: 21, url: '/augsburg/de/erste-schritte', parentId: 0, parentUrl: '/augsburg/de', title: 'erste-schritte'}),
+    new CategoryModel({id: 22, url: '/augsburg/de/erste-schritte/asylantrag', parentId: 21, parentUrl: '/augsburg/de/erste-schritte', title: 'asylantrag'})
+  ]
+
+  const categoriesContainer = new CategoriesContainer(categories)
+
+  const category1 = categoriesContainer.getCategoryByUrl('/augsburg/de/')
```
maxammann: This should be in a separate test. You can not be sure here that category1 is the first CategoryModel here.
So test against the previsously defined test data.

src/routes/categories/components/CategoryList.js
maxammann: This is now a CategoriesContainer, right?

src/modules/endpoint/models/CategoriesContainer.js
```diff
@@ -0,0 +1,72 @@
+import normalizeUrl from 'normalize-url'
```
maxammann: Should we name this CategoriesContainerModel to be more consistent? I don't think it's too long.

src/modules/endpoint/models/CategoriesContainer.js
Schedulaar: @maxammann That's not possible. un-flatten-tree needs to get arrays. Possible refactor would be
```javascript
getAncestors(category) {    
  return flatten([category],
      _category => {
        const parent = this.getCategoryById(_category.parent)
        return parent ? [parent] : []
      })
      .slice(1)
      .reverse()
}
```

src/routes/categories/components/CategoryList.js
klinzo: no, we don't need mapping to a url here anymore, because this just displays a list of categoryModels (the children of the current category), so a CategoriesContainer does not make sense here

src/modules/endpoint/models/CategoriesContainer.js
klinzo: no, I should be more carefully and testing the stuff I change :/
yeah sure, I'll add a test for this

src/routes/search/containers/SearchPage.js
klinzo: done

src/modules/endpoint/endpoints/__tests__/categories.spec.js
klinzo: okay

src/modules/endpoint/models/CategoriesContainer.js
klinzo: is it a model?

src/modules/endpoint/models/CategoriesContainer.js
klinzo: :D

src/modules/endpoint/models/__tests__/CategoriesContainer.spec.js
klinzo: you're right, done

src/modules/endpoint/models/CategoriesContainer.js
klinzo: @Schedulaar what do you think?

src/modules/endpoint/endpoints/__tests__/disclaimer.spec.js
klinzo: now it is :D

src/modules/endpoint/models/CategoriesContainer.js
Schedulaar: Hmm.. not sure about that, I think it's a container for a model and not a model itself. But our naming is difficult here (since we're in the models directory) :D

src/modules/endpoint/models/CategoriesContainer.js
```diff
@@ -0,0 +1,72 @@
...
+    return this.toArray().find(category => category.id === Number(id))
+  }
+
+  /**
+   * Returns all children of the given category
+   * @param category The category
+   * @return {CategoryModel[] | undefined} The children
+   */
+  getChildren (category) {
+    if (category) {
```
maxammann: better throw an error here is category is undefined. Or just ignore it and let the language handle it.
You could also return an empty array but that it also somewhat not intuitive :)

klinzo: we only use getChildren in CategoriesPage getContent(), where category can't be undefined, so category should always be defined

klinzo: so I would just remove the if (category), and the same for getAncestors

klinzo: @maxammann okay?

klinzo: getChildren() is also called in PdfFetcherPage, but category === undefined is also not possible here

PR #73, opened 2017-12-30
WEBAPP-144,WEBAPP-166: Test and refactor route disclaimer, create DisclaimerModel

klinzo:
Regarding WEBAPP-167: Testing the connected components is working with a real store, but only the first HOC in compose() returns a value, so if you change the order of the HOCs in compose() the test is failing, because the DisclaimerPage has different props. Feel free to try and to solve it :D Apart from that, if withFetcher is the first HOC in compose(), the props of the DisclaimerPage aka WrappedComponent are afterwards exactly the same as those of withFetcher, i.e. disclaimer/languagePayload, requestAction and urlParams, although they should be just DisclaimerPayload.data


src/routes/disclaimer/containers/__tests__/DisclaimerPage.spec.js
```diff
@@ -0,0 +1,125 @@
...
+  new LanguageModel('ar', 'Arabic')
+]
+
+const disclaimer = new DisclaimerModel({
+  id: 1689, title: 'Feedback, Kontakt und mögliches Engagement', content: 'this is a test content'
+})
+
+describe('DisclaimerPage', () => {
+  beforeEach(() => {
+    mockSetLanguageChangeUrls.mockClear()
```
maxammann: I think it would be clearer to just create a new mock function for every test. So you don't have to look up where it's created and cleared.

klinzo: hmm okay, sounds okay for me

PR #74, opened 2017-12-30
WEBAPP-143: Test and refactor route categories

klinzo:


src/modules/common/components/Page.js
```diff
@@ -3,18 +3,18 @@ import PropTypes from 'prop-types'
 
 import Caption from 'modules/common/components/Caption'
 import RemoteContent from 'modules/common/components/RemoteContent'
-import CategoryModel from '../../../modules/endpoint/models/CategoryModel'
 
 class Page extends React.Component {
```
maxammann: Is the name Page here still correct? Or should it be Category?

src/routes/categories/containers/__tests__/CategoriesPage.spec.js
```diff
@@ -0,0 +1,334 @@
...
+]
+
+const language = 'en'
+
+const mockSetLanguageChangeUrls = jest.fn()
+const mockReplaceUrl = jest.fn()
+
+describe('CategoriesPage', () => {
+  beforeEach(() => {
+    mockSetLanguageChangeUrls.mockClear()
```
maxammann: Same as in other PR. Maybe it's better to create the mock fns in the tests,

src/modules/common/components/Page.js
klinzo: yeah it is still correct, because the disclaimerPage also uses this, so this is also the reason why I moved it to common

PR #75, opened 2018-01-02
WEBAPP-126: Refactor and test app module

Schedulaar:
Not yet finished, for revision only.
* Created and tested 'layout' module with generalized Layout. Header and Footer components which are now independent of their actual content
* Moved Store.js and I18n.js to app module and refactored them to createReduxStore.js and container/I18nProvider.js (the latter of which is now a React component, so we don't have to manually subscribe to the store)
* Test for app module incoming... (hopefully today)


src/modules/app/containers/RouterFragment.js
```diff
@@ -0,0 +1,77 @@
...
+          <Layout footer={<GeneralFooter />}><LandingPage /></Layout>
+        </Fragment>
+
+        {/* Matches zero arguments like / */}
+        <Fragment forRoute='/'>
+          <Layout footer={<GeneralFooter />}><LandingPage /></Layout>
+        </Fragment>
+
+        {/* There are no missing routes. Covered:
+              * Two or more arguments (Search/Disclaimer/Events/PdfFetcher/CategoriesPage)
```
maxammann: /augsburg/ without language is kind of missing

src/modules/app/createReduxStore.js
```diff
@@ -0,0 +1,52 @@
+import { applyMiddleware, combineReducers, compose, createStore } from 'redux'
```
maxammann: ⚠️ Probably will conflict with @klinzo s PR. Because he depends on Store.js

src/modules/common/components/Caption.css
```diff
@@ -1,9 +1,11 @@
+@import url('../constants/dimens.css');
```
maxammann: I would prefere dimensions.css. No need to abbreviate here.

src/modules/endpoint/hocs/__tests__/withFetcher.spec.js
```diff
@@ -37,12 +37,12 @@ describe('withFetcher', () => {
...
   }
 
-  test('should should show error if there is one and it\'s not hidded', () => {
+  test('should should show error if there is one and it\'s not hidden', () => {
     const hoc = createComponent({
       endpoint,
       hideError: false,
       requestAction: () => new StoreResponse(true),
-      otherProps: {[endpoint.payloadName]: new Payload(false, null, 'Yepp... Error time! Wuschhh!')}
+      otherProps: {[endpoint.payloadName]: new Payload(false, null, 'Yepp... Error time! Wushhh!')}
```
maxammann: You don't tell me how to write Wuschh! :P

src/modules/layout/__tests__/Footer.spec.js
```diff
@@ -0,0 +1,19 @@
...
+    const component = <Footer items={[
+        { text: 'item1', href: '/link1' },
+        { text: 'item2', href: '/link2' }
+      ]} />
+    expect(shallow(component)).toMatchSnapshot()
+  })
+
+  test('should show Version on dev builds', () => {
+    global.__DEV__ = true
+    global.__VERSION__ = 'vX.X'
```
maxammann: These need to be set to their previous values. Else information leaks the test.

src/modules/layout/__tests__/HeaderDropDown.spec.js
```diff
@@ -0,0 +1,83 @@
...
+  let wrapperComponent
+
+  beforeEach(() => {
+    wrapperComponent = shallow(<HeaderDropDown iconSrc='/someImg'>
+      <MockNode />
+    </HeaderDropDown>)
+  })
+
+  test('should be wrapped by onClickOutside', () => {
+    // todo: ??
```
maxammann: _Don't forget me :)_

src/modules/layout/__tests__/HeaderDropDown.spec.js
```diff
@@ -0,0 +1,83 @@
...
+  })
+
+  test('should match snapshot', () => {
+    expect(wrapperComponent.dive()).toMatchSnapshot()
+  })
+
+  test('should pass correct closeDropDown callback', () => {
+    const component = wrapperComponent.dive()
+    expect(component.find('MockNode')).toHaveLength(1)
```
maxammann: You should be able to write `component.find(MockNode)` here. I'm not sure though. Would be interresting how the find() actually works :/

src/modules/layout/__tests__/HeaderDropDown.spec.js
```diff
@@ -0,0 +1,83 @@
...
+    expect(component.instance().closeDropDown).toHaveBeenCalled()
+  })
+
+  test('should be closed from the beginning', () => {
+    expect(wrapperComponent.dive().instance().state.dropDownActive).toBe(false)
+  })
+
+  test('should add class if active', () => {
+    const component = wrapperComponent.dive()
+    expect(component.find('.dropDown').prop('className')).toEqual('dropDown')
```
maxammann: Not sure if snapshot tests would fit better here. I think verifying snapshots is easy.
Refactoring is probably easier with snapshots.

src/modules/layout/components/Header.js
```diff
@@ -0,0 +1,40 @@
...
+import style from './Header.css'
+import logoWide from '../assets/integreat-app-logo.png'
+import HeaderMenuBar, { MENU_ITEMS_PROP_TYPE } from './HeaderMenuBar'
+import HeaderActionBar, { ACTION_ITEMS_PROP_TYPE } from './HeaderActionBar'
+
+/**
+ * The standard header which can supplied to a Layout. Displays a logo left, a HeaderMenuBar in the middle and a
+ * HeaderActionBar at the right (RTL: vice versa). On small viewports the HeaderMenuBar is shown underneath the rest
+ * of the Header.
+ * Uses Headroom to save space when scrolling.
```
maxammann: Did we verify headroom already? I'm my tests its responsible for more than 10ms delay for every page switch.

src/modules/layout/components/HeaderActionBar.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Link } from 'redux-little-router'
+import cx from 'classnames'
+
+import HeaderDropDown from './HeaderDropDown'
+import style from './HeaderActionBar.css'
+
+export const ACTION_ITEMS_PROP_TYPE = PropTypes.arrayOf(PropTypes.shape({
```
maxammann: Better create a class for an ActionItem. Then the proptype is also very simple.

src/routes/disclaimer/containers/__tests__/DisclaimerPage.spec.js
```diff
@@ -63,15 +62,14 @@ describe('DisclaimerPage', () => {
 
   describe('connect', () => {
     test('should map state to props', () => {
-      const store = new Store()
-      store.init({
+      const store = mockStore({
```
maxammann: 🔥 You need a real implementation of a Store in this test. Merge conflicts incoming!

src/routes/landing/components/FilterableLocationSelector.js
```diff
@@ -25,10 +25,11 @@ class FilterableLocationSelector extends React.Component {
   render () {
     return (
       <div className={style.topSpacing}>
-        <Heading/>
+        <Heading />
```
maxammann: Avoid refactors in other modules. We have separate tasks for this :) No problem in this case as I did not start the landing task yet.

src/routes/landing/containers/LandingPage.js
```diff
@@ -5,32 +5,24 @@ import compose from 'lodash/fp/compose'
...
-        <Footer/>
-      </div>
-    )
+          locations={this.props.locations} />
   }
 }
 
 const mapStateToProps = (state) => ({
-  language: state.router.params && state.router.params.language ? state.router.params.language : 'de'
+  language: state.router.params.language || 'de'
```
maxammann: 👍

src/modules/app/Navigation.js
```diff
@@ -1,19 +1,17 @@
-// todo add logic for correct hrefs here
 class Navigation {
```
maxammann: I don't like this class very much, because it's mostly duplicate code in each function and is more some kind of configuration. Maybe we can make it come kind of json-compatible configuration (Or exported js-object). The configuration would be in each route then.

Possible problems: If you change something here the code in isCategoriesSelected() will fail.
Possible solution for this would be to test isCategoriesSelected() by using the paths form Navigation.js

src/modules/app/components/GeneralHeader.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+import locationIcon from '../assets/location-icon.svg'
+import Header from 'modules/layout/components/Header'
+
+class GeneralHeader extends React.Component {
```
maxammann: I think we don't really need separate components here. They just hold a small configuration.
We can pass this directly when using it: `<Header actionItems={[{href: '/', iconSrc: locationIcon}]} />`
Less components -> less tests :)

src/modules/app/containers/RouterFragment.js
Schedulaar: hmm... this should do the same as `/augsburg` I think?

src/modules/app/createReduxStore.js
Schedulaar: I've already merged that in. I could refactor Steffs dependency to use mockStore

src/modules/layout/__tests__/HeaderDropDown.spec.js
Schedulaar: Do you have any idea on how to test that?

src/modules/layout/__tests__/HeaderDropDown.spec.js
Schedulaar: I think I tried it and it didn't work

src/modules/layout/components/Header.js
Schedulaar: What do you mean by verify?
I think we should create a task if we should continue using Headroom or write it ourselves in a better way, but I'd prefer having 10ms delay and that feature over not having 10ms delay and not having that feature :D

src/routes/disclaimer/containers/__tests__/DisclaimerPage.spec.js
Schedulaar: This was part of the merge conflict 🤔

src/routes/landing/components/FilterableLocationSelector.js
Schedulaar: I couldn't hold it back anymore :(

src/modules/app/Navigation.js
Schedulaar: Hmm. yeah I don't like the class either... Was thinking about how to avoid it :D I think we need some better abstraction for all of our routes, so we can parse links more easily without such a class

src/modules/app/components/GeneralHeader.js
Schedulaar: The test for that component will really just be a snapshot. 
The thing is: I'd like to have that every route determines its layout itself. Therefore I'd like to strip the RouterFragment.js in app module and remove the layout wrappers because they don't really belong there imo. But that way we would have these wrapper components in the routePages again. Opinions on that? :D

src/modules/endpoint/hocs/__tests__/withFetcher.spec.js
Schedulaar: You're not even consistent with your `Wuschh` aka `Wuschhh`

src/modules/layout/components/HeaderActionBar.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
```
klinzo: I think the naming of HeaderActionBar and HeaderMenuBar is not quite clear, maybe we should think of different names

src/modules/app/components/GeneralHeader.js
maxammann: I don't see the connection between the GeneralHeader and the Layout/Wrapper components thing :/

We could add a HOC which builds a layout around a components. I think that would avoid the WrapperComponents

src/modules/app/Navigation.js
maxammann: yeah, should we create a separate refactor task for this?

src/modules/app/createReduxStore.js
maxammann: @klinzo s task does not work with a mockStore

src/modules/app/Navigation.js
klinzo: we could do this in WEBAPP-90

src/modules/layout/__tests__/HeaderDropDown.spec.js
maxammann: you can use find() to look for a HOC with a specific displayName. Like `find('locationFetcher')`

src/modules/layout/__tests__/HeaderDropDown.spec.js
maxammann: k, already though that :P But I don't know why :/

src/modules/layout/components/Header.js
maxammann: Just wanted to remind that we still have the performance bug. My guess is that it's because of Headroom.

src/routes/disclaimer/containers/__tests__/DisclaimerPage.spec.js
maxammann: See above. Maybe @klinzo knows more. But I'm pretty sure this will fail.

src/modules/app/Navigation.js
```diff
@@ -1,19 +1,17 @@
...
   constructor (location, language) {
     this._location = location
     this._language = language
   }
 
-  get home () {
-    if (!this._language || !this._location) {
-      return '/'
+  get categories () {
+    if (!this._location) {
```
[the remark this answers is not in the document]
maxammann: This if actually does nothing :P

src/modules/app/containers/LanguageSelector.js
```diff
@@ -74,4 +49,4 @@ const mapStateToProps = (state) => ({
 export default compose(
   connect(mapStateToProps),
```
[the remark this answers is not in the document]
maxammann: 👍 but we have to decide in each case: If a parent is already connected to state then it's probably better to pass it down (If that path is not too long). If not then a state update would trigger a lot more render-updates.

src/modules/app/containers/LocationHeader.js
```diff
@@ -0,0 +1,74 @@
...
+const mapStateToProps = (state) => ({
+  route: state.router.route,
+  navigation: new Navigation(state.router.params.location, state.router.params.language)
+})
+
+export default compose(
+  connect(mapStateToProps),
+  withFetcher(LOCATIONS_ENDPOINT, true, true),
+  translate('app')
+)(LocationHeader)
```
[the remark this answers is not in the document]
maxammann: So yeah seems like it would make more sense to fetch once in LocationLayout and pass it down.

src/modules/app/containers/LocationLayout.js
```diff
@@ -0,0 +1,42 @@
...
+class LocationLayout extends React.Component {
+  static propTypes = {
+    location: PropTypes.string,
+    locations: PropTypes.arrayOf(PropTypes.instanceOf(LocationModel))
+  }
+
+  getCurrentLocation = () => this.props.locations.find((location) => location.code === this.props.location)
+
+  render () {
+    if (!this.getCurrentLocation()) {
```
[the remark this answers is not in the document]
maxammann: What do you mean with "when we map to a route"?

src/modules/app/containers/RouterFragment.js
maxammann: Yeah its shorter and really makes sense. Just use the language of the user in this case.

src/modules/app/createReduxStore.js
maxammann: @klinzo k if it works ok, but wasn't this useful for the test with the async stuff and real mock?

src/modules/layout/components/HeaderActionBar.js
maxammann: yeah kind of true :)

src/modules/app/containers/LocationLayout.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
```
maxammann: LocationLayout is part of the location route. It is ony used there.
I know this is problematic because of RouteFragments but we should refactor this then probably.

src/modules/app/containers/RouterFragment.js
klinzo: so we also need /augsburg/disclaimer and /augsburg/event and so on?

src/modules/app/createReduxStore.js
klinzo: it also works with a mock store, and we have a separate task for this (WEBAPP-167), but we should talk about this today on ts

src/modules/app/Navigation.js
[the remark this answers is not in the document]
klinzo: My question was kind of the same as the question regarding the /augsburg route, why we don't need a language there

src/modules/app/containers/LocationLayout.js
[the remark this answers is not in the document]
klinzo: mapping to a route in RouteFragment, I would also discuss this today on ts

src/modules/app/containers/LocationLayout.js
klinzo: no it's also used for events, disclaimer and search

klinzo: but again maybe the name is not quite clear

src/modules/layout/components/HeaderActionBar.js
klinzo: I think HeaderActionBar is good, for HeaderMenuBar I would maybe suggest HeaderNavigationBar or something like that

src/modules/layout/__tests__/HeaderDropDown.spec.js
Schedulaar: Hmm... this seems so buggy... I've worked on that for an hour now and it just won't work. The find(...) returns an empty string and when I call `wrappedComponent.name()` I get `HeaderDropDown` instead of `OnClickOutside(HeaderDropDown)` what I would expect.
I would skip that test now, since it's partially covered by the 'should match snapshot' test, because we need to dive there^^

src/routes/disclaimer/containers/__tests__/DisclaimerPage.spec.js
klinzo: yep

src/routes/main-disclaimer/containers/MainDisclaimerPage.js
```diff
@@ -5,10 +5,10 @@ class MainDisclaimerPage extends React.Component {
     return (
```
Schedulaar: Oops. Why is that there ? 🤔

Schedulaar:
Still missing:
Test for 
* I18nProvider.js
* createReduxStore.js
* createHistory.js

Navigation and RouterFragment will not be tested.

src/modules/layout/components/HeaderActionBar.js
Schedulaar: Done.

src/__mocks__/react-i18next.js
```diff
@@ -1,4 +1,8 @@
 import React from 'react'
 
 // this mock makes sure any components using the translate HoC receive the t function as a prop
-export const translate = (key) => Component => props => <Component t={() => key} {...props} />
+export const translate = (namespace) => Component => {
+  const Translated = props => <Component t={key => `${namespace}:${key}`} {...props} />
+  Translated.displayName = `Translate(${Component.displayName || Component.name})`
```
maxammann: I think it's maybe good to do a check for production here. Not sure what exactly the benefit is but major libraries do it:
https://github.com/acdlite/recompose/blob/c830f184123eebc8ea6b8b712840fdc86a680fa2/src/packages/recompose/getContext.js#L15

src/modules/app/containers/I18nProvider.js
```diff
@@ -7,67 +7,70 @@ import WebFont from 'webfontloader'
...
-  constructor () {
-    super()
-
-    // Transform locale resources to the structure: languageCode -> namespace -> key:value
-    // And not: namespace -> languageCode -> key:value
-    const i18nextResources = reduce(resources, (accumulator, namespace, namespaceName) => {
+  /**
+   * Transform locale resources to the structure: languageCode -> namespace -> key:value
+   * And not: namespace -> languageCode -> key:value
+   */
```
maxammann: Return type and params are missing

src/__mocks__/react-i18next.js
Schedulaar: I don't see why we should do that.

maxammann: k, me neither :D
You can refactor this though to:
```
setDisplayName(wrapDisplayName(Component, 'Translate'))(Translated)
```

Not sure if we want to use this library that heavily though.

PR #76, opened 2018-01-04
WEBAPP-146: Add test for landing route

maxammann:


src/routes/categories/containers/__tests__/__snapshots__/CategoriesPage.spec.js.snap
```diff
@@ -7,27 +7,27 @@ exports[`CategoriesPage should render CategoryTiles if the path is the root cate
       Array [
         LocationModel {
           "_code": "augsburg",
-          "_eventsEnabled": undefined,
-          "_extrasEnabled": undefined,
-          "_live": undefined,
+          "_eventsEnabled": false,
+          "_extrasEnabled": false,
+          "_live": false,
```
klinzo: does not really matter but should be true

src/routes/landing/components/Heading.js
```diff
@@ -4,14 +4,19 @@ import LocationBig from '../assets/LocationBig.svg'
...
 import Caption from 'modules/common/components/Caption'
+import PropTypes from 'prop-types'
 
 class Heading extends React.Component {
+  static propTypes = {
+    t: PropTypes.func.isRequired
+  }
+
   render () {
     const {t} = this.props
```
klinzo: you could inline this

src/setupMocks.js
```diff
@@ -0,0 +1,9 @@
+// Contains mocks which are used before any test runs
+// todo: We don't want to have these mocks. So this will be fixed in WEBAPP-170
```
klinzo: you should also move the rest of the models used in setupTest to setupMocks then

src/routes/landing/containers/__tests__/LandingPage.spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+      const tree = mount(
+        <Provider store={store.redux}>
+          <ConnectedLandingPage />
+        </Provider>
+      )
+
+      const landingPageProps = tree.find(LandingPage).props()
+
+      expect(landingPageProps).toEqual({
```
klinzo: you could also do a snapshot test here

src/routes/landing/components/Heading.js
maxammann: Inline how? (Give always the example in such cases :P)

src/setupMocks.js
maxammann: setupTest mocks will be gone with WEBAPP-170 :) #79

src/routes/categories/containers/__tests__/__snapshots__/CategoriesPage.spec.js.snap
maxammann: These are now the default values defined in CategoryModel. I think undefined makes never sense so I changed it to some boolean.

src/routes/landing/components/Heading.js
klinzo: this.props.t('where') instead of t('where')

src/setupMocks.js
klinzo: yeah but why create  an extra class setupMocks then?

src/routes/categories/containers/__tests__/__snapshots__/CategoriesPage.spec.js.snap
klinzo: as I said does not really matter, but categories always have to be live (otherwise they are filtered by categories endpoint), so maybe for specification reasons which our tests should be it should be true here

src/setupMocks.js
maxammann: Because I want to import the mocks from somewhere else without running the code in setupTest.js

src/routes/categories/containers/__tests__/__snapshots__/CategoriesPage.spec.js.snap
maxammann: I think false is a more reasonable, more protective default. It is true that a non-live location usually won't be rendered. I could override the default value in CategoriesPage.spec.js. What do you think?

klinzo: in this specific case I would prefer true as default value (we could even remove the live attribut completely, because we don't have any non-live categories in our whole app, except in the endpoint, so we could filter categories from the endpoint before mapping them to CategoryModel). eventsEnabled and extrasEnabled should stay false, of course

src/setupMocks.js
klinzo: ok sorry my bad

src/routes/landing/containers/__tests__/LandingPage.spec.js
klinzo: but I think it's better like this

src/routes/categories/containers/__tests__/__snapshots__/CategoriesPage.spec.js.snap
maxammann: Uh uh uh, I like removing!

src/routes/landing/containers/__tests__/LandingPage.spec.js
maxammann: yeah not sure. I won't touch it again, but we need a real reason when to use snapshots and when just a deep equal :/

src/routes/categories/containers/__tests__/__snapshots__/CategoriesPage.spec.js.snap
maxammann: You mean Location model?

klinzo: right, my bad, than leave it :D

src/routes/landing/containers/__tests__/LandingPage.spec.js
klinzo: jop

src/routes/categories/containers/__tests__/__snapshots__/CategoriesPage.spec.js.snap
klinzo: but I would change it here to true anyway

maxammann: @Schedulaar what do you think would be a sane default? In this case is really does not matter if its true or false

maxammann: Another option would be that we remove all default values from models and have to specify every value.
If the backend does not deliver some key-value pairs then our models would have default values. This is not an expected behaviour. You would expect an exception.

package.json
```diff
@@ -160,7 +160,8 @@
       "react/prop-types": "warn",
       "jsx-quotes": "warn",
       "react/jsx-tag-spacing": "warn",
-      "react/no-unused-prop-types": "warn"
+      "react/no-unused-prop-types": "warn",
+      "prefer-const": "warn"
```
Schedulaar: Can we have a linter config, so that we cannot merge into dev unless there's no linter warning?

maxammann: Change warn to error here. That's the reason we have warn and error :P

src/routes/categories/containers/__tests__/__snapshots__/CategoriesPage.spec.js.snap
Schedulaar: I actually think we shouldn't have defaults there :D

Schedulaar: So yes, rather raise an exception

package.json
Schedulaar: Why would we ever want to have warnings in dev? 🤔

src/routes/categories/containers/__tests__/__snapshots__/CategoriesPage.spec.js.snap
maxammann: I created task: https://integreat.atlassian.net/browse/WEBAPP-173

package.json
Schedulaar: And do lint errors prevent building / starting / testing?

maxammann: They shouldn't. At least I never experienced that it would prevent it.

Schedulaar: Hmm... so should we make everything an error ? :D

maxammann: If thats your definition of less annoying :P
Let's move this discusison to Slack :D

PR #77, opened 2018-01-04
WEBAPP-147: Add main disclaimer test

maxammann:


maxammann:
@klinzo done

Schedulaar:
@maxammann @klinzo Didn't we have a "strict" naming convention, that we should put all Pages into routes/page-name/containers/PageName.js ?
Not sure, if it's a good one though :D

maxammann:
@Schedulaar yes, but I think it doesnt really make sense

PR #78, opened 2018-01-05
WEBAPP-142: Test and refactor module common

klinzo:


src/modules/endpoint/models/DateModel.js
```diff
@@ -25,6 +25,32 @@ class DateModel {
   get allDay () {
     return this._allDay
   }
+
+  toLocaleString (locale) {
```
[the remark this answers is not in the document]
Schedulaar: Does it work for all locales? Have you tested it with ar and fa?

src/modules/endpoint/models/__tests__/DateModel.spec.js
```diff
@@ -0,0 +1,95 @@
+import DateModel from '../DateModel'
+import moment from 'moment'
+
+describe('DateModel', () => {
```
Schedulaar: Maybe we should include tests for other locales as well

src/modules/common/containers/Error.js
```diff
@@ -18,17 +20,23 @@ class Error extends React.Component {
...
 
   render () {
-    const {t} = this.props
     return <div>
-      <div className={style.centerText}>{t(this.props.error)}</div>
-      <div className={style.centerText}><FontAwesome name='frown-o' size="5x"/></div>
-      <a className={style.centerText} href="/" onClick={this.goBack}>Go back</a>
+      <div className={style.centerText}>{this.props.t(this.props.error)}</div>
+      <div className={style.centerText}><FontAwesome name='frown-o' size='5x' /></div>
+      <a className={style.centerText} href='/' onClick={this.goBack}>Go back</a>
```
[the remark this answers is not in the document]
Schedulaar: Yes, he does.

src/modules/endpoint/models/DateModel.js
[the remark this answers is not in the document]
klinzo: yeah forgot to comment this, nope but I could add this, maybe we need somebody to check if it's correct, but should be

src/modules/endpoint/models/__tests__/DateModel.spec.js
klinzo: will be much effort, especially because I'm not able to speak arabic/farsi

src/modules/common/containers/Error.js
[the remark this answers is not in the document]
klinzo: sure he does :P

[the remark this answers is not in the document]
klinzo: @Schedulaar ups falsch verstanden von wegen Falscher Task, dachte du machst das in deinem auch :D

src/modules/endpoint/models/__tests__/DateModel.spec.js
Schedulaar: Maybe u can do sth lik 
```javascript
locales = ['de', 'fa', 'ru', 'ar', 'en']
expect(locales.map(locale => dateModel.toLocaleString(locale))).toMatchSnapshot()
```
and then just verify real quick with Google Translate or similar (not even necessary imo). Import thing is that it actually does something in other languages than German, and if it's doing the right format in German, other languages should be correct I think.

klinzo: yeah sounds good, will be done

maxammann: `dateModel.toLocaleString` no longer exists. This is now part of TimeSpan

maxammann: Is there a reason why `toLocaleString` is again part of DateModel and Timespan is gone?
toLocaleString() *renders* a combination of Dates so why not make a component of it?

This is a lot of unnecessary complexity for a model.

src/modules/common/components/Error.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { translate } from 'react-i18next'
+import FontAwesome from 'react-fontawesome'
+
+import style from './Error.css'
+import { Link } from 'redux-little-router'
+
+export class Error extends React.Component {
```
maxammann: I renamed this to Failure because of a conflict with an ES6 class.

klinzo: I don't like the name, would rather call it ErrorComponent or something like that

src/modules/endpoint/models/__tests__/DateModel.spec.js
klinzo: if you wish I could move it to a class TimeSpan again, just thought as it is only one function and it fits logically into DateModel it would be ok, so maybe we should rather rename DateModel to TimeSpan because it's not really a DateModel

src/modules/common/components/Error.js
maxammann: pls any argument? :P
ErrorComponent doesn't follow our implicit naming convention. No other component has Component in its name.

src/modules/endpoint/models/__tests__/DateModel.spec.js
maxammann: Yeah, TimespanModel for the model and Timespan for the comoponent (or view). Sounds reasonable 👍

src/modules/common/components/Error.js
klinzo: the argument is that I don't like the name Failure and it's always Error :P

klinzo: but yeah sure both soultions aren't very good

src/modules/endpoint/models/__tests__/DateModel.spec.js
klinzo: I removed DateModel completely and added it's getter/setter functionality to EventModel, and created a TimeSpan component for the toTimeSpanString() functionality

src/modules/common/components/SearchInput.js
```diff
@@ -5,20 +5,24 @@ import { translate } from 'react-i18next'
...
+    onFilterTextChange: PropTypes.func.isRequired,
+    spaceSearch: PropTypes.bool,
+    t: PropTypes.func.isRequired
   }
 
   render () {
-    const {t} = this.props
     return (
-      <div className={this.props.className}>
+      <div className={this.props.spaceSearch ? style.searchSpacing : undefined}>
```
klinzo: Do I need undefined or an empty string ''? or something different?

maxammann: classnames supports a more advanced syntax: https://github.com/JedWatson/classnames

```
cx({ [style.searchSpacing]: this.props.spaceSearch });
```

klinzo: nice, thx

src/modules/common/components/__tests__/Failure.spec.js
```diff
@@ -0,0 +1,16 @@
...
+  const mockTranslate = jest.fn((msg) => msg)
+
+  test('should match snapshot', () => {
+    const wrapper = shallow(
+      <Failure error='Error Message' t={mockTranslate} />
+    )
+
+    expect(wrapper).toMatchSnapshot()
+  })
+})
```
Schedulaar: I think, we miss the test, that if we `import Failure from '../Failure'` that the translate(...) has been wrapped around it, right?

src/modules/common/components/__tests__/SearchInput.spec.js
```diff
@@ -0,0 +1,20 @@
...
+  })
+
+  test('should render and space search', () => {
+    const mockTranslate = jest.fn((text) => text)
+    expect(shallow(<SearchInput filterText={'Test'}
+                                onFilterTextChange={() => {}}
+                                t={mockTranslate}
+                                spaceSearch />)).toMatchSnapshot()
+  })
+})
```
Schedulaar: Same as above.

src/modules/common/containers/LanguageSelector.js
```diff
@@ -1,40 +1,64 @@
...
+    const preventLink = (event) => {
+      event.preventDefault()
+      this.props.closeDropDownCallback()
+    }
+
+    return this.props.languages.map(language =>
+      <Link
+        key={language.code}
+        className={cx(style.element, this.props.language === language.code ? style.elementActive : '')}
+        onClick={language.code === this.props.language ? preventLink : this.props.closeDropDownCallback}
```
Schedulaar: Why do we need to prevent the link? We shouldn't do that imo.
We should make closeDropDownCallback not required and check, if closeDropDownCallback is defined, so we can better reuse LanguageSelector

src/modules/common/components/RemoteContent.js
```diff
@@ -10,12 +10,13 @@ class RemoteContent extends React.Component {
       PropTypes.shape({
         __html: PropTypes.string.isRequired
       }).isRequired
-    ).isRequired
+    ).isRequired,
+    isCentered: PropTypes.bool
```
Schedulaar: I think a more common name would be just `centered` instead of `isCentered` here.

src/modules/common/components/__tests__/Failure.spec.js
maxammann: Our translate mock and the snapshot testing should catch these mistakes

Schedulaar: But there's no snapshot for that, is there?

src/modules/common/containers/LanguageSelector.js
klinzo: what sense does it make to change the language to the same language you're currently at? this does just reload the page, and, because the languageChangeUrls are undefined for the current language even redirects to the root categories page, which definitely shouldn't be the case

Schedulaar: If you see it like that, then there shouldn't be a <Link /> for the current language in the first place. Maybe you can replace it with a <span />?

src/modules/common/components/__tests__/Failure.spec.js
maxammann: It creates a snapshot which has the translated key in it. Uhmmm....
To what does our mock translate keys?

So yeah it probably does not test for the translate HOC atm. But do we really need it? I mean we still don't know if the translation exists etc.

Schedulaar: No, but we know that this.props.t doesnt raise an exception because it isnt there
And There are translated keys, because steff mocked t himself in Failure.spec.js

src/modules/common/containers/LanguageSelector.js
maxammann: I think we wanted the links for SEO.
When you click the language your currently having it should just close the popup and do nothing else. 
For formatting move some of the stuff out of the jsx.

Also reformat to this:
className={cx(style.element, {[style.elementActive]: this.props.language === language.code}}

src/modules/common/components/__tests__/Failure.spec.js
maxammann: So yeah the test for ConnectedFailure is missing 🔥

src/modules/common/containers/LanguageSelector.js
Schedulaar: Links for other languages are alright, thats good for SEO, but a Link for the current language is not good, right? As steff said, the links for the current language are more or less broken, then a link rather confuses Search engines.

maxammann: Shouldn't we fix the link for the current language then? Or is it not important? If that is the case then we can remove the anchor tag for the current language. The current language acts then as a close button.

Schedulaar: I think it's not important. And that's what I meant to say :D Replace `<Link />` for the current language with `<span />`.

src/modules/common/components/__tests__/Failure.spec.js
klinzo: will be added

src/modules/common/containers/LanguageSelector.js
klinzo: fixing the link is easy, we could just take the pathname from router. So we don't close the drop down menu?

src/modules/common/components/__tests__/Failure.spec.js
klinzo: do you think a snapshot is enough?

src/modules/common/components/RemoteContent.js
klinzo: done

src/modules/common/containers/LanguageSelector.js
klinzo: makes sense that the current language is  not clickable 👍

src/modules/common/components/__tests__/Failure.spec.js
maxammann: Or lets call it TranslatedFailure. So you only need to render the translated component.

src/modules/common/containers/LanguageSelector.js
maxammann: So we just want to use a span which has no href. And when you click it it closes the popup.

klinzo: maybe we shouldn't even close the popup

maxammann: But then the user wouldn't get any feedback

klinzo: right, done

klinzo: @Schedulaar we don't need to check if closeDropDownCallback is defined, because if it's undefined nothing happens which is what we want

PR #80, opened 2018-01-07
WEBAPP-150: Test SerachPage

maxammann:


src/routes/search/containers/__tests__/SearchPage.spec.js
```diff
@@ -0,0 +1,193 @@
...
+        categories,
+        languages,
+        setLanguageChangeUrls: expect.any(Function)
+      })
+    })
+
+    test('should map dispatch to props', () => {
+      const store = createReduxStore(createHistory, {
+        categories: new Payload(false),
+        languages: new Payload(false),
```
klinzo: you don't have to initialize categories and languages

PR #81, opened 2018-01-15
WEBAPP-132: CSS - Care Super Slightly... oder so

Schedulaar:
Tests incoming someday...
I'm abusing this issue for several smaller things like:
* Reduce size of categoryTiles on mobile
* Improve landing page: The key of grouped locations stick at the top while scrolling as well as the search bar. And clicking / tapping on the search bar scrolls down to the list
* Fix different kind of css things
* Reduce font size slightly (still bigger than native app) and increase line-height a bit (opinions?)
* Reduce amount of fonts


maxammann:
Didn't look thorugh it yet, but updated tests are missing :)

Schedulaar:
@maxammann As mentioned in the PR description :P

src/modules/common/components/ScrollingSearchContainer.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react'
```
klinzo: the naming is maybe not optimal, because it would be a container in the components folder

src/modules/common/components/ScrollingSearchContainer.js
```diff
@@ -0,0 +1,58 @@
...
+    if (scrollTop < this.state.top) {
+      this.scroll()
+    }
+  }
+
+  scroll = () => {
+    animateScroll.scrollTo(this.state.top, {duration: 500})
+  }
+
+  onChange = value => {
```
klinzo: rename to onFilterTextChange

src/routes/landing/components/LocationSelector.js
```diff
@@ -7,18 +7,6 @@ import { Link } from 'redux-little-router'
...
-  }
-
-  render () {
-    return (
-      <div className={style.languageListParent}>{this.props.name}</div>
-    )
-  }
-}
-
 class LocationEntry extends React.Component {
```
klinzo: move to a seperate file

src/routes/landing/components/LocationSelector.js
```diff
@@ -37,7 +25,12 @@ class LocationSelector extends React.Component {
   static propTypes = {
     locations: PropTypes.arrayOf(PropTypes.instanceOf(LocationModel)),
     filterText: PropTypes.string.isRequired,
-    language: PropTypes.string
+    language: PropTypes.string,
```
klinzo: language and locations should be required

src/routes/landing/components/LocationSelector.js
```diff
@@ -7,18 +7,6 @@ import { Link } from 'redux-little-router'
...
-  render () {
-    return (
-      <div className={style.languageListParent}>{this.props.name}</div>
-    )
-  }
-}
-
 class LocationEntry extends React.Component {
   static propTypes = {
     language: PropTypes.string,
```
klinzo: should be required

src/routes/landing/components/FilterableLocationSelector.js
```diff
@@ -16,19 +16,19 @@ class FilterableLocationSelector extends React.Component {
```
klinzo: language and locations should be required

src/modules/common/components/ScrollingSearchContainer.js
```diff
@@ -0,0 +1,58 @@
...
+    if (node) {
+      const {top, height} = node.getBoundingClientRect()
+      this.setState(Object.assign({}, this.state, {top, height}))
+    }
+  }
+
+  render () {
+    return <React.Fragment>
+      <div ref={this.setDimensions} className={style.searchBar}
+           style={{'top': this.props.stickyTop + 'px'}}>
```
klinzo: another usecase for styled components, I don't like mixing inline styling and styling in separate css sheets, but right now there is no better solution, right?

klinzo:
image
last row doesn't look very good

klinzo:
if you scroll slow on the landing page, you can location names flickering between the search input and the current key, which isn't a big issue though.
I like your improvements 👍

src/modules/common/components/ScrollingSearchContainer.css
```diff
@@ -0,0 +1,12 @@
+@import url('../../common/constants/colors.css');
+
+/* noinspection CssOverwrittenProperties */
+.search-bar {
```
klinzo: little top-padding would be nice imo

klinzo: while the search input is sticky

Schedulaar:
@klinzo You're probably using Firefox where the flickering exists... This seems like a glitch of firefox's renderer... But not really too annoying for me :D
And I'm not really sure about aligning the category tiles... It felt right for me, to have them all centered, because the whole layout is a "centered" layout, aligning them to the left somehow breaks my system there :D ... But that's some design question, maybe we should ask #design-marketing

src/modules/common/components/ScrollingSearchContainer.js
Schedulaar: Hmm.. yeah... suggestions ? :P

src/modules/common/components/ScrollingSearchContainer.js
```diff
@@ -0,0 +1,58 @@
...
+  scroll = () => {
+    animateScroll.scrollTo(this.state.top, {duration: 500})
+  }
+
+  onFilterTextChange = value => {
+    this.props.onFilterTextChange(value)
+    this.scroll()
+  }
+
+  setDimensions = node => {
```
maxammann: I think we should stick to es6 methods in classes and no lambda members.
So `setDimensions(node) {`
When you are at it you can look arount in the project for these kind of definions.

src/modules/common/components/ScrollingSearchContainer.js
klinzo: maybe just ScrollingSearchBar?

src/modules/common/components/SearchInput.js
```diff
@@ -11,19 +10,21 @@ export class SearchInput extends React.Component {
...
     onFilterTextChange: PropTypes.func.isRequired,
     spaceSearch: PropTypes.bool,
+    onClick: PropTypes.func,
     t: PropTypes.func.isRequired
   }
 
   render () {
     return (
-      <div className={cx({ [style.searchSpacing]: this.props.spaceSearch })}>
+      <div className={this.props.spaceSearch ? style.searchSpacing : ''}>
```
maxammann: I prefere passing an object instead of an empty string :/

src/routes/categories/components/CategoryTiles.js
```diff
@@ -32,11 +32,11 @@ class CategoryTiles extends React.Component {
     return (
       <div>
         <Caption title={this.getLocationName(this.props.title)} />
-        <Row>
+        <div className={style.categoryTiles}>
```
maxammann: Do we use the Row still? If not we can remove react-flexbox-grid.

src/modules/app/containers/I18nProvider.js
```diff
@@ -73,9 +73,9 @@ export class I18nProvider extends React.Component {
 
   loadFonts (language) {
     // Lateef for arabic ui and content, Open Sans for latin text in arabic text, Raleway for latin ui
-    const arabicFonts = ['Lateef:400', 'Raleway:300,400,400i,600,700,700i', 'Open+Sans:400']
+    const arabicFonts = ['Lateef:400', 'Raleway:400,700', 'Open+Sans:400,700']
     // We do not need an arabic font
-    const latinFonts = ['Raleway:300,400,400i,600,700,700i', 'Open+Sans:400']
```
maxammann: Maybe we should load some more Open Sans fonts. The cms can deliver italic and fonts with different font weights so we need to support them.

src/modules/common/components/ScrollingSearchContainer.js
Schedulaar: But these have to be bound explicitly (when used with event handling) which is quite annoying

src/modules/common/components/SearchInput.js
Schedulaar: empty object? as a classname?

maxammann: I mean the syntax I used before, but I'm also ok with the trinary

src/modules/common/components/ScrollingSearchBox.js
```diff
@@ -0,0 +1,60 @@
...
+
+  setReference (node) {
+    if (node) {
+      this._node = node
+    }
+  }
+
+  render () {
+    return <React.Fragment>
+      <div ref={node => this.setReference(node)} className={style.searchBar}
```
maxammann: Refs are not that good to use in react. I don't see an other option here. The whole scrolling stuff is also a little bit overpowered maybe here :)

src/modules/common/components/ScrollingSearchBox.js
```diff
@@ -0,0 +1,60 @@
...
+           style={{'top': this.props.stickyTop + 'px'}}>
+        <SearchInput filterText={this.props.filterText} onFilterTextChange={value => this.onFilterTextChange(value)}
+                     onClick={() => this.onClick()}
+                     spaceSearch={this.props.spaceSearch} />
+      </div>
+      <div className={style.searching}>
+        {React.cloneElement(this.props.children, {stickyTop: this.props.stickyTop + 45})}
```
maxammann: Where does this 45 come from?

src/modules/common/components/ScrollingSearchBox.js
```diff
@@ -0,0 +1,60 @@
...
+  static propTypes = {
+    filterText: PropTypes.string.isRequired,
+    onFilterTextChange: PropTypes.func.isRequired,
+    stickyTop: PropTypes.number.isRequired,
+    spaceSearch: PropTypes.bool,
+    children: PropTypes.element.isRequired
+  }
+
+  static defaultProps = {
+    stickyTop: 0
```
maxammann: stickyProps is always default, why do we need it? looks more like a constant.

src/modules/common/components/ScrollingSearchBox.js
maxammann: If we don't have the stickyTop we can also remove the children from here and don't need to clone them.

src/modules/common/components/SearchInput.js
```diff
@@ -11,19 +10,21 @@ export class SearchInput extends React.Component {
     filterText: PropTypes.string.isRequired,
     onFilterTextChange: PropTypes.func.isRequired,
     spaceSearch: PropTypes.bool,
+    onClick: PropTypes.func,
```
maxammann: onClick makes you think it's the onClick for the whole componet. Maybe onClickInput would be better.

src/modules/common/components/ScrollingSearchBox.js
Schedulaar: StickyTop is the top position of the DOM-element where the element should be fixed when scrolling below. If you have multiple sticky elements (in this case SearchInput and the selector group initials), we need to tell our child components (which could also have a sticky element) what the most top position should be. (If you always set stickyTop: 0, then the groups would be hidden by the search input)

src/modules/common/components/ScrollingSearchBox.js
Schedulaar: Yeah, but we need ref here.

src/modules/common/components/ScrollingSearchBox.js
Schedulaar: the search input has a height of 45px, then the groups should stick below that.

src/modules/common/components/SearchInput.js
Schedulaar: Done

src/modules/common/components/ScrollingSearchBox.js
maxammann: Then we should define the 45 as a constant. There should be no magic values. Maybe even for the scroll time. Like FAST_SCROLL = 100, MEDIUM_SCROLL = 1000

Schedulaar: Yeah, I'll do it. 100 is too fast. I'll just add one with 500.

maxammann:
I think we are a bit inconsitent with our scrolling behaviour:
**LandingPage:** The SeachInput is always visible, when clicking it or typing in it it get's scrolled so you see the results better.
**Rest of the app:** The header get's hidden when scrolling, when scrolling up (becouse you remember there is something you need) you instantly see it.

I'd propose we use the second method all around our app. When you are on the landing page and scroll you probably don't want to search. If you decide: Fuck it that list is too long and want to search you scroll up and the Serach appears instantly.

Schedulaar:
@maxammann Hmm.. You're kind of right there :D so another use case for Headroom? :/

maxammann:
@Schedulaar yeah, but we should be sure that it doesn't add that much overhead to our app. :D
Maybe you can test before and after implementing it with Headroom?

Schedulaar:
@maxammann Also it's unclear then how the groups of the locationselector should be sticky... I think the only possibilty would be to have them stick to top: 0px; and then if you scroll up, the Headroom with the search input scrolls in and covers the groups, right?

maxammann:
@Schedulaar yeah :D I hope I unterstood it correctly but sounds right.
When scrolling up the SearchInput appears.

Schedulaar:
It was a bit tricky, because with Headroom, I have no control that the width of the input stays the same (because headroom adds a position: fixed to it). I solved this by adding the css-part of Headroom myself using position:sticky :D
I won't test this, since it may be refactored anyways in WEBAPP-169, but I made sure, it works in all supported browsers.
Cool side effect was, that the glitch that the input scrollled a little bit too much up (while scrolling up and headroom was in fixed position) disappeared with the sticky implementation

Schedulaar:
@maxammann Also I can't see any difference in terms of performance, so we should be good to go (I hope)

PR #84, opened 2018-01-17
WEBAPP-89: Fetch events in german

klinzo:
until our cms automatically duplicates not translated events, we always load events in 'de', no matter what language the user has selected


Schedulaar:
Hmmm... There are some instances that translate their events. These would be disappointed if they dont See their events in the correct language. We should leave this issue to the CMS imo.

klinzo:
@Schedulaar we agreed in the coworking space after you left to always use german, because there are nearly no instances translating their events, and it would be too much effort to compare the events in two languages.

PR #87, opened 2018-02-05
WEBAPP-90: Make route configuration more centralized, add automatic language redirect

maxammann:


src/modules/app/containers/RouterFragment.js
```diff
@@ -13,54 +13,80 @@ import PdfFetcherPage from 'routes/pdf-fetcher/containers/PdfFetcherPage'
...
             </Fragment>
             {/* Matches /augsburg/de/events* -> Events */}
             <Fragment forRoute='/events(/:id)'>
-              <LocationLayout><EventsPage /></LocationLayout>
+              <EventsPage />
             </Fragment>
             {/* Matches /augsburg/de/fetch-pdf/* -> Redirect */}
             <Fragment forRoute='/fetch-pdf'>
-              <Layout><PdfFetcherPage /></Layout>
+              <PdfFetcherPage />
```
maxammann: Uses now LocationLayout

src/modules/app/containers/RouterFragment.js
```diff
@@ -13,54 +13,80 @@ import PdfFetcherPage from 'routes/pdf-fetcher/containers/PdfFetcherPage'
...
+    routerConfig: PropTypes.instanceOf(RouteConfig).isRequired
+  }
+
+  static isLanguageCode (language) {
+    return language && language.length === 2
+  }
+
   render () {
+    const {routerConfig} = this.props
+    const matchRoute = (id) => routerConfig.matchRoute(id)
```
Schedulaar: Just a hint for pureComponents: Lambda's defined inside the render() method also trigger a re-render of all pure-components receiving them as prop. (Doesn't matter in this case, since RouterFragment shouldn't re-render a lot itself, since its property doesn't change.)

src/modules/app/containers/RouterFragment.js
```diff
@@ -13,54 +13,80 @@ import PdfFetcherPage from 'routes/pdf-fetcher/containers/PdfFetcherPage'
...
     return <Fragment forRoute='/'>
       {/* Routes */}
       <React.Fragment>
+        {/* No language was provided to redirect to a specific language (e.g. the browsers language) */}
+        <Fragment forRoute='/'>
+          <I18nRedirect />
+        </Fragment>
 
         {/* Matches two or more arguments like /augsburg/de */}
         <Fragment forRoute='/:location/:language(/*)'>
```
Schedulaar: As discussed in lecture today, I'd really like these routes to be somehow referenced via js, (not put the string in manually), to deduplicate the definition of the routes... But I am also unsure how to do this. :/

src/modules/app/containers/__tests__/__snapshots__/App.spec.js.snap
```diff
@@ -70,7 +70,298 @@ exports[`App should match snapshot 1`] = `
     }
   >
     <Connect(I18nProvider)>
-      <RouterFragment />
+      <RouterFragment
```
Schedulaar: I don't like this snapshot. Way too long and too much information.

src/modules/app/containers/I18nRedirect.js
```diff
@@ -0,0 +1,35 @@
...
+import { connect } from 'react-redux'
+
+export class I18nRedirect extends React.Component {
+  static propTypes = {
+    redirect: PropTypes.func.isRequired,
+    currentPath: PropTypes.string.isRequired
+  }
+
+  static contextTypes = {
+    i18n: PropTypes.object.isRequired
```
Schedulaar: We shouldn't access the context directly. Is there any other method to do that?

Schedulaar: Maybe we could add the i18nLanguage into the store in our i18nextProvider?

src/modules/endpoint/Payload.js
```diff
@@ -12,7 +12,7 @@ class Payload {
    * @param error {string|null} The error message which has occurred or null
    * @param requestUrl {string|null} The url from which to fetch or null if we're not sure yet
    * @param fetchDate {number|null} The date when the last fetch has been started
-   * @throws {Error} If you supply data and error or the url is invalid.
+   * @throws {RouterRedirect} If you supply data and error or the url is invalid.
```
Schedulaar: throws RotuerRedirect?

src/modules/app/RouteConfig.js
```diff
@@ -0,0 +1,38 @@
...
+    this.routes = routes.map(route => {
+      if (isPlainObject(route)) {
+        return new Route(route.id, route.path)
+      }
+
+      throw new Error('Value if routes config is no object!')
+    })
+  }
+
+  getRouterConfig () {
```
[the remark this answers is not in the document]
klinzo: it just returns an object with all routes with the path as name right? but documentation would be very nice :D

src/modules/app/containers/App.js
```diff
@@ -10,18 +9,68 @@ import languagesEndpoint from '../../endpoint/endpoints/languages'
...
+      {
+        id: CategoriesPage,
+        path: '/:location/:language(/*)'
+      },
+      {
+        id: MainDisclaimerPage,
+        path: '/disclaimer'
+      },
+      {
+        path: '/:unknown(/)'
```
klinzo: why do we need this? and why doesn't it have any id?

src/modules/app/containers/App.js
```diff
@@ -10,18 +9,68 @@ import languagesEndpoint from '../../endpoint/endpoints/languages'
...
+      },
+      {
+        id: LandingPage,
+        path: '/:language(/)'
+      }
+    ]
```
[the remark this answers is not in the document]
klinzo: 👍

src/modules/app/containers/App.js
```diff
@@ -10,18 +9,68 @@ import languagesEndpoint from '../../endpoint/endpoints/languages'
...
+import RouteConfig from '../RouteConfig'
+import RouterFragment from './RouterFragment'
 
 class App extends React.Component {
+  store
+  routeConfig
+
   componentWillMount () {
-    this._store = createReduxStore(createHistory)
+    const routes = [
```
klinzo: couldn't we create Route objects here?

src/modules/app/containers/App.js
```diff
@@ -10,18 +9,68 @@ import languagesEndpoint from '../../endpoint/endpoints/languages'
...
+      {
+        path: '/'
```
klinzo: id: LandingPage would be nice I think

src/modules/app/containers/App.js
```diff
@@ -10,18 +9,68 @@ import languagesEndpoint from '../../endpoint/endpoints/languages'
...
   }
 
   render () {
-    return <Provider store={this._store}>
+    return <Provider store={this.store}>
       <EndpointProvider
         endpoints={[languagesEndpoint, locationEndpoint, categoriesEndpoint, disclaimerEndpoint, eventsEndpoint]}>
         <I18nProvider>
-          <RouterFragment />
+          <RouterFragment routerConfig={this.routeConfig} />
```
klinzo: do we have a routeConfig or a routerConfig? naming should be consistent :D

src/modules/app/containers/I18nRedirect.js
klinzo: change PropTypes.object to PropTypes.instanceOf(i18n)

src/modules/app/containers/RouterFragment.js
```diff
@@ -13,54 +13,80 @@ import PdfFetcherPage from 'routes/pdf-fetcher/containers/PdfFetcherPage'
...
  * todo: Test and document in WEBAPP-90
  * todo: Layouts should be set in each route
  */
 class RouterFragment extends React.Component {
+  static propTypes = {
+    routerConfig: PropTypes.instanceOf(RouteConfig).isRequired
+  }
+
+  static isLanguageCode (language) {
+    return language && language.length === 2
```
klinzo: can't we also simplify this to return language.length === 2?

src/modules/app/containers/I18nRedirect.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { replace } from 'redux-little-router'
+import normalizeUrl from 'normalize-url'
+
+import { connect } from 'react-redux'
+
+export class I18nRedirect extends React.Component {
```
klinzo: please write a short documentation what your classes are used for, I18nRedirect for example is just used to redirect if we don't have a language in our path right?

src/modules/app/containers/__tests__/__snapshots__/App.spec.js.snap
klinzo: @maxammann just changed it so that the pr looks a lot bigger than it is :P 300 lines extra 👍

src/modules/layout/components/LocationFooter.js
```diff
@@ -2,19 +2,22 @@ import React from 'react'
...
-    navigation: PropTypes.instanceOf(Navigation).isRequired,
+    matchRoute: PropTypes.func.isRequired,
+    currentParams: PropTypes.object.isRequired,
     t: PropTypes.func.isRequired
   }
 
   render () {
     return <Footer>
-      <Link href={this.props.navigation.disclaimer}>{this.props.t('imprintAndContact')}</Link>
+      <Link href={this.props.matchRoute(DisclaimerPage).stringify(this.props.currentParams)}>
```
klinzo: why do you need the whole DisclaimerPage here to check for the route? wouln't 'DisclaimerPage' be sufficient?

src/modules/layout/components/LocationHeader.js
```diff
@@ -11,47 +10,62 @@ import LocationModel from 'modules/endpoint/models/LocationModel'
...
-
-  isEventsSelected = () => this.props.route === '/:location/:language/events(/:id)'
-
   getActionItems () {
-    const {navigation} = this.props
+    const {matchRoute, currentParams} = this.props
     return [
-      new HeaderActionItem({href: navigation.search, iconSrc: searchIcon}),
-      new HeaderActionItem({href: navigation.locationSelector, iconSrc: locationIcon}),
+      new HeaderActionItem({href: matchRoute(SearchPage).stringify(currentParams), iconSrc: searchIcon}),
```
klinzo: same as in LocationFooter

src/modules/layout/containers/LocationLayout.js
```diff
@@ -30,18 +34,20 @@ export class LocationLayout extends React.Component {
       return <Layout header={<GeneralHeader />} footer={<GeneralFooter />}>{this.props.children}</Layout>
     }
 
-    const {route, navigation} = this.props
-    return <Layout header={<LocationHeader location={location} route={route} navigation={navigation} />}
-                   footer={<LocationFooter navigation={navigation} />}>
+    const {route, matchRoute} = this.props
+    return <Layout header={<LocationHeader location={location}
+                                           route={route} matchRoute={matchRoute}
+                                           currentParams={this.props.currentParams} />}
```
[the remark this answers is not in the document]
klinzo: also makes the code more readable imo, because you now what params are meant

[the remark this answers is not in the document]
klinzo: you are already passing location separately

src/modules/app/RouteConfig.js
```diff
@@ -0,0 +1,38 @@
...
+      return accumulator
+    }, {})
+  }
+
+  /**
+   * @param id The id of the route to look for
+   * @return {Route} The matched route
+   */
+  matchRoute (id) {
+    const route = this.routes.find(route => route.id && route.id === id)
```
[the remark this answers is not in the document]
klinzo: getRoute() would be a better name imo

src/modules/app/RouteConfig.js
```diff
@@ -0,0 +1,38 @@
...
+import { isPlainObject } from 'lodash/lang'
+
+export class RouteConfig {
+  routes = []
+
+  constructor (routes = []) {
+    this.routes = routes.map(route => {
+      if (isPlainObject(route)) {
+        return new Route(route.id, route.path)
+      }
```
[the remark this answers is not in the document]
maxammann: Because I want to force the type of routes here :) it's an array of object

src/modules/app/RouteConfig.js
```diff
@@ -0,0 +1,38 @@
...
+
+      throw new Error('Value if routes config is no object!')
```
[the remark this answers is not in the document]
maxammann: haha intellij refactoring made this :D

src/modules/app/containers/App.js
maxammann: LandingPage is not used right now for linking. We can add it later if we want to.

src/modules/app/containers/App.js
maxammann: This has to do with the I18nRedirect. See RouterFragment for this.

src/modules/app/containers/I18nProvider.js
```diff
@@ -51,13 +53,19 @@ export class I18nProvider extends React.Component {
   }
 
   setLanguage (language) {
-    const targetLanguage = language || this._i18n.languages[0]
+    const targetLanguage = language || this.i18n.languages[0]
+
+    if (targetLanguage.length !== 2) {
```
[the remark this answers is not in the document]
maxammann: It is string :D

src/modules/app/containers/I18nRedirect.js
maxammann: Type of i18n is actually object. instanceOf does not work.

@Schedulaar why shouldn't we access the context? This is how the translate hoc works.

src/modules/app/containers/RouterFragment.js
maxammann: No because language can be undefined.

src/modules/app/containers/RouterFragment.js
maxammann: It's not possible without loosing a lot of flexibility.

maxammann: Came to this conclusion after several hours at least :D

src/modules/app/containers/RouterFragment.js
maxammann: It shouldn't but, but it would make things a lot more complicated. I think we don't really need to care here.

src/modules/app/containers/__tests__/__snapshots__/App.spec.js.snap
maxammann: :P would need to replace this with a normal test then.

src/modules/layout/components/LocationFooter.js
maxammann: The ids are actual class references here. So a string wouln't be enough.

src/modules/app/containers/App.js
```diff
@@ -10,18 +9,68 @@ import languagesEndpoint from '../../endpoint/endpoints/languages'
...
+      {
+        id: MainDisclaimerPage,
+        path: '/disclaimer'
+      },
+      {
+        path: '/:unknown(/)'
+      },
+      {
+        id: LandingPage,
+        path: '/:language(/)'
```
klinzo: can we unify this with `{path: '/'}` as `{id: LandingPage, path: '/(:language(/))}`, because '/' is also the path to the landing page

src/modules/layout/containers/LocationLayout.js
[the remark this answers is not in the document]
maxammann: Done

src/modules/layout/containers/LocationLayout.js
```diff
@@ -15,13 +14,18 @@ import LocationFooter from '../components/LocationFooter'
...
+    matchRoute: PropTypes.func.isRequired,
     location: PropTypes.string.isRequired,
     locations: PropTypes.arrayOf(PropTypes.instanceOf(LocationModel)).isRequired,
     route: PropTypes.string.isRequired,
+    currentParams: PropTypes.object.isRequired,
     children: PropTypes.node
   }
 
+  componentWillUnmount () {
+    console.log('LocationLayout unmounted')
```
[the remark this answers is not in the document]
maxammann: Done

src/modules/app/RouteConfig.js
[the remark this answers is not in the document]
maxammann: Done

src/modules/app/RouteConfig.js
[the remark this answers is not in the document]
maxammann: Done

src/modules/app/containers/RouterFragment.js
maxammann: Done

src/modules/endpoint/Payload.js
maxammann: Done

src/modules/app/RouteConfig.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  /**
+   * @param id The id of the route to look for
+   * @return {Route} The matched route
+   */
+  matchRoute (id) {
+    const route = this.routes.find(route => route.id && route.id === id)
+    if (!route) {
+      throw Error(`Route ${id} was not found in RouteConfig!`)
```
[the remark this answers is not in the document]
maxammann: Done

src/modules/app/containers/App.js
maxammann: Done

src/modules/app/containers/App.js
maxammann: Those are two different paths. One shows the I18nRedirect and one is LandingPage.

maxammann:
Actually I'm not that sure if we should merge this PR. It adds a lot of complexit without any real benefit imo.

PR #88, opened 2018-02-06
WEBAPP-169: Replace Headroom

Schedulaar:
I still need to test the changes, but I want you to check it out (so I don't do unnecessary testing :D)


src/modules/common/components/Headroom.js
```diff
@@ -0,0 +1,107 @@
...
+    // this.lastKnownScrollY: Last ScrollTop (to detect change in scroll direction)
+    // this.currentScrollY: Current ScrollTop
+    // this.direction: 'up' || 'down'
+
+    this.currentScrollY = this.getScrollY()
+    if (this.lastKnownScrollY < this.currentScrollY) { // We're moving up
+      if (this.direction !== 'up') { // We've changed direction from down to up!
+        this.lastTransform = this.state.transform
+        this.lastScrollY = this.lastKnownScrollY
+        this.direction = 'down'
```
klinzo: shouldn't this be: `this.direction = 'up'`?

src/modules/common/components/Headroom.js
```diff
@@ -0,0 +1,107 @@
...
+      }
+    } else if (this.lastKnownScrollY > this.currentScrollY) { // We're moving down
+      if (this.direction !== 'down') { // We've changed direction from up to down!
+        this.lastTransform = this.state.transform
+        this.lastScrollY = this.lastKnownScrollY
+        this.direction = 'up'
```
klinzo: same as above, shouldn't the new direction be 'down'?

src/modules/common/components/Headroom.js
```diff
@@ -0,0 +1,107 @@
...
+    if (transform + currentScrollY < this.props.pinStart) {
+      return 0
+    }
+    return Math.min(0, Math.max(-this.props.scrollHeight, transform))
+  }
+
+  update () {
+    // this.lastScrollY: ScrollTop when the scroll direction last changed
+    // this.lastTransform: Transform when the scroll direction last changed
+    // this.lastKnownScrollY: Last ScrollTop (to detect change in scroll direction)
```
klinzo: what is the difference between lastScrollY and lastKnownScrollY?

src/modules/common/components/Headroom.js
```diff
@@ -0,0 +1,107 @@
...
+    if (window.pageYOffset !== undefined) {
+      return window.pageYOffset
+    } else if (window.scrollTop !== undefined) {
+      return window.scrollTop
+    } else {
+      return (document.documentElement || document.body.parentNode || document.body).scrollTop
+    }
+  }
+
+  truncate (transform, currentScrollY) {
```
klinzo: documentation would be nice :D

src/modules/common/components/Headroom.js
klinzo: and what is transform?

src/modules/common/components/Headroom.js
```diff
@@ -0,0 +1,107 @@
...
+  constructor (props) {
+    super(props)
+    this.handleEvent = this.handleEvent.bind(this)
+    this.update = this.update.bind(this)
+    this.ticking = false
+    this.lastScrollY = 0
+    this.lastTransform = 0
+    this.lastKnownScrollY = 0
+    this.currentScrollY = 0
+    this.direction = 'down'
```
maxammann: Magic number, introduce some kind of constant for this

src/modules/common/components/Headroom.js
```diff
@@ -0,0 +1,107 @@
...
+    height: PropTypes.number
+  }
+
+  static defaultProps = {
+    pinStart: 0
+  }
+
+  constructor (props) {
+    super(props)
+    this.handleEvent = this.handleEvent.bind(this)
```
maxammann: Should we go with bind or lambdas? I'm pro lambdas.

maxammann: We also have a linter rule which restricts bind in jsx

package.json
```diff
@@ -34,6 +34,7 @@
     "redux-little-router": "~14.2.1",
     "redux-logger": "~3.0.6",
     "redux-thunk": "~2.2.0",
+    "redux-responsive": "^4.3.5",
```
maxammann: Did you measure performance when resizing the window? :D

Schedulaar: Haha nope, but I tried resizing and I couldnt notice any performance impacts.
Other than that I don't care about performance when resizing :P

src/modules/common/components/Headroom.js
Schedulaar: No, but my comment(s) are wrong ^^

src/modules/common/components/Headroom.js
Schedulaar: 0 is not a magic number imo

src/modules/common/components/Headroom.js
Schedulaar: Yup, documentation and testing will be added. I created the pull request, so that you can check the header on your devices. I want to see how you like it.

src/modules/common/components/Headroom.js
maxammann: but if `this.direction !== 'up'` is true then direction is already down

src/modules/common/components/Headroom.js
maxammann: k, let's call it magic string 'down' :D

src/modules/common/components/Headroom.js
Schedulaar: Ok that's funny :D why does it work then? :D

src/modules/layout/containers/Header.js
```diff
@@ -1,23 +1,28 @@
...
-import HeaderActionBar from './HeaderActionBar'
+import HeaderNavigationBar from '../components/HeaderNavigationBar'
+import HeaderActionBar from '../components/HeaderActionBar'
 import HeaderActionItem from '../HeaderActionItem'
 import HeaderNavigationItem from '../HeaderNavigationItem'
+import Headroom from '../../common/components/Headroom'
+import { connect } from 'react-redux'
+
+const HEADER_HEIGHT_LARGE = 85
+const HALF_HEADER_HEIGHT_SMALL = 55
```
klinzo: why did you name this HALF_...?

package.json
maxammann: does not cause performance impacts as it only pushes to the store when the viewport changes from small to big. Not for every pixel change 👍

src/modules/common/components/Headroom.js
```diff
@@ -0,0 +1,141 @@
...
+
+const UPWARDS = 'up'
+const DOWNWARDS = 'down'
+
+class Headroom extends React.PureComponent {
+  static propTypes = {
+    children: PropTypes.any.isRequired, // The child node to be displayed as a header
+    scrollHeight: PropTypes.number.isRequired, // The maximum amount of px the header should move up when scrolling
+    pinStart: PropTypes.number.isRequired, // The minimum scrollTop position where the transform should start
+    stickyAncestor: PropTypes.node, // Gets rendered with a corresponding stickyTop prop as an ancestor
```
maxammann: I'm ok with that formatting as I could not find a JSDoc example for documenting values of plain objects.

src/modules/common/components/Headroom.js
```diff
@@ -0,0 +1,141 @@
...
+
+    this.setState({transform, stickyTop})
+
+    this.lastKnownScrollTop = currentScrollTop
+  }
+
+  handleEvent () {
+    if (!this.ticking) {
+      this.ticking = true
+      raf(() => { // Request animation frame to optimize performance for dom changes
```
maxammann: raf is only a devDependencies and no dependency. Not sure if we really need the polyfil though. It is only included currently because of a bug of jest.

Schedulaar:
@maxammann Right, that was a little bit annoying, I've changed the behavior, so it's not continuous, but at least it doesn't glitch now :D Pls tell me if you like it, afterwards I'll add the test :P

maxammann:
glitch

src/modules/common/components/Headroom.js
```diff
@@ -0,0 +1,133 @@
...
+      return this.props.pinStart + this.props.scrollHeight >= scrollTop
+    } else {
+      return this.props.pinStart >= scrollTop
+    }
+  }
+
+  /**
+   *
+   * @param scrollTop
+   * @param direction
```
maxammann: missing types

src/modules/common/components/Headroom.js
```diff
@@ -0,0 +1,133 @@
...
+    } else {
+      return (document.documentElement || document.body.parentNode || document.body).scrollTop
+    }
+  }
+
+  constructor (props) {
+    super(props)
+    this.handleEvent = this.handleEvent.bind(this)
+    this.update = this.update.bind(this)
+    this.state = {mode: STATIC, transition: false}
```
maxammann: declare these variables

PR #91, opened 2018-02-20
WEBAPP-181: Show better error message for Lüneburg

Schedulaar:


src/modules/common/components/Failure.js
```diff
@@ -16,12 +16,13 @@ export class Failure extends React.Component {
   }
 
   render () {
+    const {t, error} = this.props
     return <div>
-      <div className={style.centerText}>{this.props.t(this.props.error)}</div>
+      <div className={style.centerText}>{t(error)}</div>
       <div className={style.centerText}><FontAwesome name='frown-o' size='5x' /></div>
-      <Link className={style.centerText} href={'/'}>Go Back</Link>
+      <Link className={style.centerText} href={'/'}>{t('goBack')}</Link>
```
maxammann: This should be called goToStart or something like this because it doesn't go back in history, but leads to the root.

PR #92, opened 2018-02-20
WEBAPP-64: Implement extras

klinzo:
This task is still in progress. Currently we have to wait for
1) the delivering of icons and names for the extras by our cms
2) a solution to access the sprungbrett api, which is currently not possible because access control allow origin is disabled, so either this has to be enabled or we have to use a proxy

If you have the time you can suggest improvements :D


maxammann:
I'm not sure if reviewing is good if the task is not done yet. I'd suggest only to review if there are blocking code questions which need to be cleared. 

Starting reviewing now just adds a lot of comments which would be resolved anyway probably later on. At least thats what I saw on my last PR :P

maxammann:
I'd suggest to close it as long as this is not finished. So we have a place to view the code.

PR #93, opened 2018-02-21
WEBAPP-190: Add git-hooks to runl linters tests

maxammann:
What do you think about this? Does this help or should every dev decide for them selfes? Push and checkout times increase with this PR.


Schedulaar:
Im not sure about these changes, we could try them though. Imo it's not too bad, if there are broken builds pushed to feature branches, since only one should work on them at a time (and we should avoid branch dependencies as well anyway).

maxammann:
The question is just if this will bug us less than broken builds because of linting  :P
This probably will avoid commmits like 'Fix linting'. I also think we probably shouldn't run the tests. And running yarn on each checkout may take a while too.

PR #95, opened 2018-02-21
WEBAPP-175: Add tests for createHistory and RouterFragment

maxammann:


src/modules/app/__tests_/createHistory.spec.js
```diff
@@ -0,0 +1,30 @@
+import createHistory from '../createHistory'
+import { createMemoryHistory } from 'history'
+
+describe('createHistory', () => {
+  test('should create history to create a browser history by default', () => {
+    expect(() => createHistory()).not.toThrow()
```
Schedulaar: Do you need to wrap that in a lambda? 🤔

maxammann: No, but I think this is more expressive. Else you would maybe no know what this should test for.

PR #97, opened 2018-02-21
WEBAPP-18: Display language selector when language is not available

Schedulaar:


src/routes/categories/containers/LanguageFailure.js
```diff
@@ -0,0 +1,69 @@
...
+  getTitle () {
+    return this.props.locations.find(location => location.code === this.props.location).name
+  }
+
+  render () {
+    return <React.Fragment>
+      <Caption title={this.getTitle()} />
+      <p style={{
+        textAlign: 'center',
+        margin: '25px 0'
```
klinzo: do we want to inline styles?

src/modules/common/containers/LanguageSelector.css
```diff
@@ -11,6 +11,15 @@
   text-align: center;
 }
 
+.vertical {
```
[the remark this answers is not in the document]
Schedulaar: When the language is not available, they are shown vertically inside the layout. (see LanguageFailure)

src/routes/categories/containers/LanguageFailure.js
Schedulaar: I think it doesn't matter here. If it's just two / three lines of css. Especially complex css-code like for the header should be extracted into classes to improve readabilty imo.

src/modules/common/containers/LanguageSelector.css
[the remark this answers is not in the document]
klinzo: yeah sorry, I had an outdated version

src/modules/common/containers/__tests__/LanguageSelector.spec.js
```diff
@@ -24,9 +24,9 @@ describe('LanguageSelector', () => {
     ar: 'test/url/ar'
   }
 
-  test('should match snapshot', () => {
-    const mockCloseDropDownCallback = jest.fn()
+  const mockCloseDropDownCallback = jest.fn()
```
maxammann: jest.fn() functions shouldn't be outside of test() environments. You probably expect a fresh mock when using it in different tests. Also if you don't expect anything from it you can just pass an empty lambda.

src/modules/endpoint/hocs/withFetcher.js
```diff
@@ -24,15 +23,15 @@ const contextTypes = {
  * @param endpointName {string} The name of the endpoint to fetch from
  * @param hideError {boolean} If you want to hide errors in the render() method
  * @param hideSpinner {boolean} If you want to hide a loading spinner in the render() method
+ * @param FailureComponent {*} the component which is rendered in error-case if hideError is false
  * @return {buildHOC} Returns a HOC which renders the supplied component as soon as the fetcher succeeded
  */
-export function withFetcher (endpointName, hideError = false, hideSpinner = false) {
+export function withFetcher (endpointName, hideError = false, hideSpinner = false, FailureComponent = Failure) {
```
maxammann: Should we combine the hideError and FailureComponent parameters? If you do not set a FailureComponent it does not get shown.

maxammann: Or if you set it explicitly to null.

src/routes/categories/containers/__tests__/LanguageFailure.spec.js
```diff
@@ -0,0 +1,147 @@
...
+        location: location,
+        setLanguageChangeUrls: expect.any(Function),
+        t: expect.any(Function),
+        locations: locations,
+        languages: languages
+      })
+    })
+
+    test('should map dispatch to props', () => {
+      const store = createReduxStore(createHistory, {
```
maxammann: I think a mockstore should be enough here because you are only checking if the setLanguageChangeUrls action has been called.

src/routes/categories/containers/LanguageFailure.js
maxammann: I think we shouldnt even inline simple ones. We should stay consistent accross all styles. We already have some 2-3 line styles.

src/routes/categories/containers/__tests__/LanguageFailure.spec.js
```diff
@@ -0,0 +1,147 @@
...
+
+      const languageFailure = mount(
+        <Provider store={store}>
+          <EndpointProvider endpoints={[locationsEndpoint, languagesEndpoint]}>
+            <ConnectedLanguageFailure />
+          </EndpointProvider>
+        </Provider>
+      ).find(LanguageFailure)
+
+      languageFailure.props().setLanguageChangeUrls(mapLanguageToUrl, languages)
```
maxammann: I think setLanguageChangeUrls should already get called by mount()

src/modules/endpoint/hocs/withFetcher.js
Schedulaar: Yup makes sense with explicitly setting it to null I think.

src/routes/categories/containers/__tests__/LanguageFailure.spec.js
Schedulaar: Right

src/routes/categories/containers/LanguageFailure.js
Schedulaar: Agreed. :P

src/routes/categories/containers/__tests__/LanguageFailure.spec.js
Schedulaar: Since LanguageFailure has two withFetcher prepended, the standard mockStore doesn't work here.

src/modules/common/containers/__tests__/LanguageSelector.spec.js
Schedulaar: done

PR #100, opened 2018-02-22
WEBAPP-172: Validate all todos, replace withStateMapper with template strings

klinzo:


src/modules/app/containers/__tests__/I18nRedirect.spec.js
```diff
@@ -12,7 +12,7 @@ describe('I18nRedirect', () => {
       <I18nRedirect
         currentPath={'/augsburg/'}
         redirect={redirect} />,
-      {context: {i18n: {language: 'de'}}, childContextTypes: {i18n: PropTypes.object.isRequired}} // fixme why can't we use I18nProvider here?
+      {context: {i18n: {language: 'de'}}, childContextTypes: {i18n: PropTypes.object.isRequired}}
```
klinzo: this is okay, see the official doc: https://github.com/airbnb/enzyme/blob/master/docs/api/mount.md#arguments

src/modules/layout/containers/__tests__/LocationLayout.spec.js
```diff
@@ -42,28 +41,34 @@ describe('LocationLayout', () => {
...
-      // todo: add locations
-      expect(tree.find(ConnectedLocationLayout).childAt(0).props()).toMatchSnapshot()
+      ).find(LocationLayout)
+
+      expect(locationLayout.props()).toEqual({
+        path: path,
+        location: location,
+        language: language,
+        locations: locations,
+        dispatch: expect.any(Function)
```
klinzo: does anyone now why there is this dispatch prop in the props?

src/modules/app/containers/__tests__/I18nRedirect.spec.js
maxammann: Yes it works, but it still makes no sense that wrapping it in a I18nProvider does not work here.

src/modules/endpoint/Endpoint.js
```diff
@@ -113,16 +109,15 @@ class Endpoint {
...
       }
 
-      const formattedURL = format(this.url, urlParams)
-      /*
-       todo: check if there are any paramters left in the url: formattedURL.match(/{(.*)?}/)
-       currently this does not work as unused paramaters are just removed from the url
-       */
+      const formattedURL = this.mapRouterToUrl(router)
+
+      if (formattedURL.includes('undefined')) {
```
maxammann: Do we need to use a string here? `[undefined].includes(undefined)` is true for example.

src/modules/endpoint/EndpointBuilder.js
```diff
@@ -20,18 +18,17 @@ class EndpointBuilder {
...
-    this._stateMapperBuilder = new StateMapperBuilder(this)
 
     this._refetchLogic = (urlParams, nextUrlParams) => !isEqual(urlParams, nextUrlParams)
   }
 
   /**
-   * Adds a url to the builder
-   * @param url The url
+   * Adds a router to url mapper to the builder
+   * @param routerToUrlMapper THe routerToUrlMapper which is mapping the router to a url
```
maxammann: Minor typo in 'THe'

src/modules/endpoint/endpoints/categories.js
```diff
@@ -4,10 +4,10 @@ import CategoryModel from '../models/CategoryModel'
 import CategoriesMapModel from '../models/CategoriesMapModel'
 
 export default new EndpointBuilder('categories')
-  .withUrl('https://cms.integreat-app.de/{location}/{language}/wp-json/extensions/v0/modified_content/pages?since=1970-01-01T00:00:00Z')
-  .withStateMapper().fromArray(['location', 'language'], (state, paramName) => state.router.params[paramName])
-  .withMapper((json, urlParams) => {
-    const baseUrl = `/${urlParams.location}/${urlParams.language}`
+  .withRouterToUrlMapper(router => `https://cms.integreat-app.de/${router.params.location}/${router.params.language}` +
```
maxammann: :+1:

src/modules/endpoint/endpoints/locations.js
```diff
@@ -15,15 +15,15 @@ function stripSlashes (path) {
...
+  .withRouterToUrlMapper(() => 'https://cms.integreat-app.de/wp-json/extensions/v1/multisites')
   .withMapper(json => {
     const locations = json
       .map((location) => new LocationModel({
         name: location.name,
         code: stripSlashes(location.path),
         live: location.live,
         eventsEnabled: location['ige-evts'] === '1',
-        extrasEnabled: true // todo: Adjust this in WEBAPP-64
+        extrasEnabled: true
```
maxammann: This todo is still valid until WEBAPP-64 is done, right?

src/modules/layout/containers/__tests__/LocationLayout.spec.js
maxammann: Every connected component gets the dispatch prop

src/modules/endpoint/Endpoint.js
```diff
@@ -98,12 +94,12 @@ class Endpoint {
     return `${this.stateName}Payload`
   }
 
-  requestAction (urlParams = {}) {
+  requestAction (router) {
```
maxammann: Why did you change this to router? All implementations of endpoints just use the params object in that 'router'.
I don't really want a dependency from the router to the endpoint API because the router is likely to change in the future.

maxammann: We could introduce an ParamContext or LocationContext which holds the params and offers methods for accessing them.

maxammann: Or we change 'router' to 'state' and just pass the whole state. This would also allow to include the state in the url. Which is basically what this feature would be for. So we would have a mapStateToUrl function.

src/modules/app/containers/__tests__/I18nRedirect.spec.js
klinzo: yep, I tried, but couldn't get it to work...should I create a task for this?

src/modules/endpoint/Endpoint.js
klinzo: which is okay because we won't ever have an endpoint where the request url is undefined. Do you have a idea how to solve this more elegant anyway?

src/modules/endpoint/endpoints/locations.js
klinzo: Yeah I'm also currently working on WEBAPP-64, so I thought I could just remove it here.

src/modules/endpoint/Endpoint.js
klinzo: you're right, so only params or the whole state....which would you prefer? the state could be big...but on the other hand for the extras I need the state

src/modules/app/containers/__tests__/I18nRedirect.spec.js
maxammann: Yep, this "fixme" is not fixed yet :P Maybe there are reasons why this does not work. Or it's a but of enzyme

src/modules/endpoint/Endpoint.js
maxammann: yep, but you can just write `if (formattedURL.includes(undefined)) {`

src/modules/app/containers/__tests__/I18nRedirect.spec.js
klinzo: done

PR #101, opened 2018-02-22
WEBAPP-180: Translate Datenschutz in footers

klinzo:


maxammann:
I changed the base so the diff is cleaner

PR #103, opened 2018-02-23
WEBAPP-191: Add link to categories home

maxammann:


src/modules/layout/containers/GeneralHeader.js
```diff
@@ -1,11 +1,11 @@
 import React from 'react'
```
klinzo: why did you move GeneralHeader to the containers package? It does not connect to the store nor does it use withFetcher or translations

klinzo: Okay, but then LocationHeader and GeneralHeader should also be containers because Header is connecting to the state

Schedulaar: Hmm.. yap... Do you think, we need a better definition of when to call a ReactComponent a container and when to call it a component? Or does this distinction still make sense in our project at all? 🤔
At least I like, that we have some distinction between complex ReactComponents, that can modify the state, vs dumb / just visual components...

maxammann:
@Schedulaar @klinzo my solution is now to make Header a component. So I added Headroom to LocationLayout and RouterFragment which are already containers

src/modules/layout/components/Header.js
```diff
@@ -2,27 +2,23 @@ import React from 'react'
...
  * of the Header.
  * Uses Headroom to save space when scrolling.
  */
-export class Header extends React.Component {
+class Header extends React.Component {
   static propTypes = {
     navigationItems: PropTypes.arrayOf(PropTypes.instanceOf(HeaderNavigationItem)).isRequired,
     actionItems: PropTypes.arrayOf(PropTypes.instanceOf(HeaderActionItem)).isRequired,
-    smallViewport: PropTypes.bool.isRequired
+    logoHref: PropTypes.string.isRequired
```
Schedulaar: Add viewport here

PR #104, opened 2018-02-26
WEBAPP-199: Fix false-positive tests and review withFetcher

maxammann:


src/modules/endpoint/Endpoint.js
```diff
@@ -104,18 +91,19 @@ class Endpoint {
...
+      const payload = state[this.stateName]
+      if (payload.isFetching) {
         return new StoreResponse(false)
       }
 
       const formattedURL = this.mapStateToUrl(state)
 
-      if (formattedURL.includes(undefined)) {
-        throw new Error(`Some necessary params in the state were undefined:${formattedURL}`)
+      if (formattedURL.includes('undefined')) {
```
klinzo: :D

src/modules/endpoint/__tests__/Endpoint.spec.js
```diff
@@ -85,22 +85,26 @@ describe('Endpoint', () => {
   describe('Actions', () => {
     let clock
     const mockedTime = 0
+    let prevError
 
     beforeEach(() => {
       clock = lolex.install({now: mockedTime, toFake: []})
+      prevError = console.error
+      console.error = error => console.log(`Some expected error was thrown: ${error}`)
```
Schedulaar: (Why) Do we need a console.log ?

src/modules/endpoint/hocs/__tests__/withFetcher.spec.js
```diff
@@ -120,57 +118,59 @@ describe('withFetcher', () => {
...
 
-    hoc.setProps({...hoc.props(), ...otherProps}) // No change in props
+    hoc.setProps({...hoc.props(), state}) // No change in props
 
     // mockRequestAction has been called once in componentWillMount but it shouldn't have been a second time
     expect(mockRequestAction.mock.calls).not.toHaveLength(2)
 
-    const newProps = {[endpoint.payloadName]: new Payload(false)} // newProps !== otherProps (identity)
-    hoc.setProps({...hoc.props(), ...newProps})
+    const newState = {[endpoint.stateName]: new Payload(false)} // neState !== state (identity)
```
klinzo: typo, change to newState

Schedulaar: typo in comment

src/modules/endpoint/hocs/withFetcher.js
```diff
@@ -31,59 +33,47 @@ export function withFetcher (endpointName, FailureComponent = Failure, hideSpinn
...
       componentWillReceiveProps (nextProps) {
         // Dispatch new requestAction to ask the endpoint whether data is available, if:
-        // (a) the Fetcher urlParams prop changed or
-        // (b) the Fetcher endpoint.payloadName prop changed because of new data in the store (e.g. because a payload has been fetched)
-        const endpoint = this.endpoint
-        if (endpoint.shouldRefetch(this.props.state, nextProps.state) ||
-          this.props[endpoint.payloadName] !== nextProps[endpoint.payloadName]) {
-          this.fetch(nextProps.state)
+        // (a) the Fetcher endpoint.payloadName prop changed because of new data in the store (e.g. because a payload has been fetched)
+        if (!isEqual(this.props.state, nextProps.state) || this.getStatePayload(this.props) !== this.getStatePayload(nextProps)) {
```
klinzo: still deep compare on whole state :/

src/modules/endpoint/hocs/withFetcher.js
```diff
@@ -93,13 +83,13 @@ export function withFetcher (endpointName, FailureComponent = Failure, hideSpinn
           }
         }
 
-        if (this.errorVisible()) {
+        if (this.getStatePayload().error) {
```
klinzo: change to `payload.error`

src/modules/endpoint/hocs/withFetcher.js
```diff
@@ -93,13 +83,13 @@ export function withFetcher (endpointName, FailureComponent = Failure, hideSpinn
...
         }
 
-        const allProps = ({...this.props, [this.endpoint.stateName]: payload.data})
+        const allProps = ({...this.props, [this.props.endpoint.stateName]: payload.data})
         // Strip all internal data
-        delete allProps[this.endpoint.payloadName]
+        delete allProps.endpoint
         delete allProps.getEndpoint
         delete allProps.state
         delete allProps.requestAction
```
klinzo: why don't we just pass only
```js
[this.props.endpoint.stateName]: payload.data
```
as props to the WrappedComponent instead of adding all props to allProps and deleting them again?

src/modules/endpoint/__tests__/Endpoint.spec.js
maxammann: We want to let the tests fail if console.error is executed. So we need to overwrite console.error before doing something which is supposed to output an error so the test does not fail.

Schedulaar: Yep, thats clear. Why did you choose to overwrite with a console.log? Why not a noop? For better debugging?

src/modules/endpoint/hocs/withFetcher.js
Schedulaar: Because we *want* to pass all props to the WrappedComponent:
If Component1 needs someInfo as a prop and needs the events payload.data then we want to use it like so:
`const FetchingComponent1 = withFetcher('events')(Component1)`
`<FetchingComponent1 someInfo='someInfo:)' />` and expect that `someInfo` gets passed through the `<Fetcher />` into Component1 along with the events data.

klinzo: do we use this?

maxammann: Yeah we use this at some points. It failed when not passing all the props.

PR #105, opened 2018-02-26
WEBAPP-200: Fix missing viewport prop

klinzo:


src/modules/layout/components/GeneralHeader.js
```diff
@@ -5,7 +5,7 @@ import HeaderActionItem from '../HeaderActionItem'
 
 class GeneralHeader extends React.Component {
   render () {
-    return <Header logoHref={'/'} actionItems={[new HeaderActionItem({href: '/', iconSrc: locationIcon})]} />
+    return <Header viewportSmall logoHref={'/'} actionItems={[new HeaderActionItem({href: '/', iconSrc: locationIcon})]} />
```
Schedulaar: viewportSmall should only be true, when state.viewport.is.small is true.

PR #109, opened 2018-03-05
WEBAPP-206: Sort search results and highlight query

klinzo:


package.json
```diff
@@ -23,6 +23,7 @@
     "react-dom": "~16.2.0",
     "react-flexbox-grid": "~2.0.0",
     "react-fontawesome": "~1.6.1",
+    "react-highlighter": "^0.4.2",
```
maxammann: small dependency which should be fine.

src/routes/search/containers/SearchPage.js
```diff
@@ -32,15 +32,27 @@ export class SearchPage extends React.Component {
...
+    return title.includes(filterText) || content.includes(filterText)
   }
 
   findCategories () {
+    const filterText = this.state.filterText.toLowerCase()
+
     return this.props.categories.toArray()
       .filter(category => this.acceptCategory(category))
+      // sort results so that categories including the filter text in the title on top
+      .sort((category1, category2) => {
```
maxammann: Should we maybe also sort alphabetically? Also I trust that code that it sorts correctly but it's rater complicated imo. Maybe some comments would be nice.

src/routes/search/containers/SearchPage.js
```diff
@@ -32,15 +32,27 @@ export class SearchPage extends React.Component {
   }
```
maxammann: The sorting should be tested (A few examples for sorting should be added in the spec)

src/routes/search/containers/SearchPage.js
klinzo: we could also filter for a match in title and a match in content separately and concat the results afterwards if that increases readability? alphabetical sorting is okay for me, but adds more complexity

package.json
klinzo: hope so, 15.000 downloads a week should be enough :D anyway, is the style of the highlighting ok or should I change something there?

src/routes/search/containers/SearchPage.js
klinzo: done

src/routes/search/containers/SearchPage.js
maxammann: We can sort first alphabetically and afterwards by an occurance by title.

src/routes/categories/components/CategoryListItem.css
```diff
@@ -43,3 +43,7 @@
     border-bottom: 1px solid var(--theme-color);
   }
 }
+
+.highlight {
+  background-color: var(--theme-color);
```
maxammann: Should we use here a lighter color? or maybe no color at all and just bolden it?

src/routes/search/containers/SearchPage.js
Schedulaar: I wouldn't bother alphabetical sorting... Doesn't make sense in a search function and nobody expects alphabetical sorting there...
Maybe a better-readable version would be:
```javascript
prioritizeCategory (category) {
  const filterText = this.state.filterText.toLowerCase()
  if (category.title.toLowerCase().includes(filterText)) {
    return 2
  } else if (category.content.toLowerCase().includes(filterText)) {
    return 1
  } else {
    return 0
  }
}

findCategories () {
  return this.props.categories.toArray()
    .map(category => ({priority: prioritizeCategory(category), category})
    .filter(({priority, category}) => priority !== 0)
    .sort((wrapper1, wrapper2) => wrapper1.priority - wrapper2.priority)
    .map(wrapper => ({model: wrapper.category, children: []))
}
```

maxammann: I think alphabetical sorting makes sense because imo it would be nice if the search is deterministic.

src/routes/categories/components/CategoryListItem.css
klinzo: @Schedulaar

src/routes/search/containers/SearchPage.js
klinzo: I also like alphabetical sorting :D

klinzo: @Schedulaar your suggestion does only work if we don't want to sort alphabetically

klinzo: what do you think of `findCategories2 () {
    const filterText = this.state.filterText.toLowerCase()

    const categoriesWithTitle = this.props.categories.toArray()
      .filter(category => category.title.toLowerCase().includes(filterText))
      .sort((category1, category2) => category1.title < category2.title ? -1 : category1.title > category2.title ? 1 : 0)

    const categoriesWithContent = this.props.categories.toArray()
      .filter(category => !category.title.toLowerCase().includes(filterText))
      .filter(category => category.content.toLowerCase().includes(filterText))
      .sort((category1, category2) => category1.title < category2.title ? -1 : category1.title > category2.title ? 1 : 0)

    return categoriesWithTitle
      .concat(categoriesWithContent)
      .map(category => ({model: category, children: []}))
  }`?

klinzo: what do you think of
 `findCategories () {
    const filterText = this.state.filterText.toLowerCase()

    const categoriesWithTitle = this.props.categories.toArray()
      .filter(category => category.title.toLowerCase().includes(filterText))
      .sort((category1, category2) => category1.title < category2.title ? -1 : category1.title > category2.title ? 1 : 0)

    const categoriesWithContent = this.props.categories.toArray()
      .filter(category => !category.title.toLowerCase().includes(filterText))
      .filter(category => category.content.toLowerCase().includes(filterText))
      .sort((category1, category2) => category1.title < category2.title ? -1 : category1.title > category2.title ? 1 : 0)

    return categoriesWithTitle
      .concat(categoriesWithContent)
      .map(category => ({model: category, children: []}))
  }`?

klinzo: with some comments :D

Schedulaar: My solution can be easily adjusted: 
```js
findCategories () {
  return this.props.categories.toArray()
    .map(category => ({priority: prioritizeCategory(category), category})
    .filter(({priority, category}) => priority !== 0)
    .sort((wrapper1, wrapper2) => {
      if(wrapper1.priority !== wrapper2.priority) {
         return wrapper1.priority - wrapper2.priority)
      } else {
        return wrapper1.category.title.localeCompare(wrapper2.category.title)
      }
     })
    .map(wrapper => ({model: wrapper.category, children: []))
}
```
https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

src/routes/categories/components/CategoryListItem.css
Schedulaar: Not sure, I think just bold font would be enough...

src/routes/search/containers/SearchPage.js
maxammann: I'm fine with both :) Let the PR-OP decide. 

I'd propose to da a function which compares two strings and ignores case.

klinzo: I don't really care which one we use, both is not beautiful :D I like the localeCompare though, so we have now: ```findCategories () {
    const filterText = this.state.filterText.toLowerCase()

    const categoriesWithTitle = this.props.categories.toArray()
      .filter(category => category.title.toLowerCase().includes(filterText))
      .sort((category1, category2) => category1.title.localeCompare(category2.title))

    const categoriesWithContent = this.props.categories.toArray()
      .filter(category => !category.title.toLowerCase().includes(filterText))
      .filter(category => category.content.toLowerCase().includes(filterText))
      .sort((category1, category2) => category1.title.localeCompare(category2.title))

    return categoriesWithTitle
      .concat(categoriesWithContent)
      .map(category => ({model: category, children: []}))
  }``` or @Schedulaar 's suggestion, @NKlug your opinion?

src/routes/categories/components/CategoryListItem.css
klinzo: Just bold font:
image
little bit unobstrusive imo

Schedulaar: I think it's obtrusive ( :D ) enough because the user knows what he has typed and is only interested in looking through the search results. With more highlighting the user concentrates more on the search key (which he already knows) and less on the content I think. 🤔

klinzo: sounds logic ^^

src/routes/search/containers/SearchPage.js
NKlug: I like @klinzo 's code more since it's easier to read and to understand.

src/routes/categories/components/CategoryList.js
```diff
@@ -18,7 +18,9 @@ class CategoryList extends React.Component {
       children: PropTypes.arrayOf(PropTypes.instanceOf(CategoryModel)).isRequired
     })).isRequired,
     title: PropTypes.string,
-    content: PropTypes.string
+    content: PropTypes.string,
+    /** A search query to highlight in the categories titles **/
```
Schedulaar: I think you only need one `*` to close the comment tag for js-doc style 🤔 (`/** A search query to highlight in the categories titles */`)

klinzo: right.... thanks

PR #110, opened 2018-03-05
WEBAPP-68: Add flow type checking for categories route

klinzo:


maxammann:
If we move the prop-types to type definitions right now you wont't get any warnings about missing prop types in the IDE. Also flow does not throw errors when they are missing. We maybe need this plugin to check them/add them in compile time: https://www.npmjs.com/package/babel-plugin-flow-react-proptypes

maxammann:
You also need to set a flow binary in the intellij config. For me that is: `/home/max/projects/integreat-webapp/node_modules/flow-bin`

We need to add a guide to the readme. Ofc it would ne nice if it worked out of the box.

Also some stuff is red in the flow enabled files.

klinzo:
@maxammann you can go to: `Settings | Languages & Frameworks | JavaScript` and choose 'Flow' as JS Language Version, which also supports ECMAScript6 and JSX, set `/<project-directory>/node_modules/flow-bin` as Flow Package or Executable and check the checkbox for `Use Flow server for: Type checking`. This enables flow checking for erros during coding, which also shows missing props (even errors). For me there is nothing marked red in flow enabled files, could you maybe provide more information on this? A guide would maybe nice yes :D

klinzo:
image

klinzo:
@NKlug did everything work for you?@markl @NKlug what did you have to adjust in order to get it to work?

maxammann:
@klinzo setting the binary is enough. As far as I understand flow you should have it globally installed because a Server runs in the background.

There should be also some scripts in the package.json so run it manually and restart the flow server.

klinzo:
@maxammann the official doc says; Flow works best when installed per-project with explicit versioning rather than globally, see: https://flow.org/en/docs/install/
you don't need scripts in package.json, just run yarn flow

maxammann:
@klinzo ok then the only thing which is missing is documentation in README.md

klinzo:
@Schedulaar yeah right, question mark means optional, everything else is required, so undefined is not accepted

klinzo:
@Schedulaar so is this okay or do you have something to request changes for?

Schedulaar:
@klinzo Go for it :D

PR #111, opened 2018-03-06
WEBAPP-157: Prevent news button from being clicked if there are none

NKlug:
Show 'News' grayed out and unclickable if there are no news and display  'No News' as a tooltip


src/modules/layout/components/HeaderNavigationBar.css
```diff
@@ -21,7 +21,12 @@
   text-transform: uppercase;
 }
 
-.navigation-item.active-navigation-item {
+.navigation-item.inactive-navigation-item {
+  opacity: 0.5;
```
Schedulaar: With a small viewport in Landkreis Alzey-Worms (who don't have news): When you open the LanguageDropDown, the title NEWS is shown above the languages. (Whyever this is not the case when you don't set opacity: 0.5; 🤔 )
You can fix this by switching the order of <HeaderNavigationBar /> and <HeaderActionBar /> in Header.js

src/modules/layout/components/HeaderNavigationBar.css
```diff
@@ -21,7 +21,12 @@
...
+  cursor: default;
```
Schedulaar: You could also set user-select: none; in the Header.css in .header: This prevents that the user can select text (and therefore the cursor will be displayed as default).

src/modules/layout/components/HeaderNavigationBar.css
NKlug: What do you mean by "switching the order"?

Schedulaar: the order in the html markup

src/modules/layout/components/HeaderNavigationBar.css
NKlug: Done (/)

src/modules/layout/components/HeaderNavigationBar.css
Schedulaar: So the result would look like this in Header.js: 
``` jsx
    return (
      <Headroom scrollHeight={scrollHeight}>
        <header className={style.header}>
          <div className={style.logoWide}>
            <Link href={this.props.logoHref}>
              <img src={logoWide} />
            </Link>
          </div>
          <HeaderNavigationBar className={style.navigationBar} items={this.props.navigationItems} />
          <HeaderActionBar className={style.actionBar} items={this.props.actionItems} />
        </header>
      </Headroom>
    )
```

NKlug: Thanks, fixed it.

src/modules/layout/components/LocationHeader.js
```diff
@@ -54,25 +55,33 @@ class LocationHeader extends React.Component {
...
+
     const extras = isExtrasEnabled() &&
       new HeaderNavigationItem({
         href: '/',
-        active: isExtrasSelected(),
-        text: t('extras')
+        selected: isExtrasSelected(),
+        text: t('extras'),
+        active: true,
+        tooltip: ''
```
klinzo: remove `tooltip: ''`, this is not needed

src/modules/layout/components/LocationHeader.js
```diff
@@ -54,25 +55,33 @@ class LocationHeader extends React.Component {
...
 
     const categories = isCategoriesEnabled() &&
       new HeaderNavigationItem({
         href: matchRoute(CategoriesPage).stringify(currentParams),
-        active: isCategoriesSelected(),
-        text: t('categories')
+        selected: isCategoriesSelected(),
+        text: t('categories'),
+        active: true,
+        tooltip: ''
```
klinzo: same as above

src/modules/layout/components/HeaderNavigationBar.js
```diff
@@ -21,10 +22,15 @@ class HeaderMenuBar extends React.Component {
     return <div
       className={cx(this.props.className, style.navigationBar, isEmpty(this.props.items) ? style.hidden : '')}>
       {this.props.items.map(item => (
-        <Link key={item.text}
-              className={cx(style.navigationItem, item.active ? style.activeNavigationItem : '')}
-              href={item.href}>{item.text}</Link>
+        <ConditionalLink active={item.active} data-tip={!item.active ? item.tooltip : ''}
```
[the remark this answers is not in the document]
NKlug: I thought about that, too, but im my opionion conditionally rendering a (Disabled)Link would be a lot of duplicate code, especially regarding the props that are passed. Furthermore, if there are similar cases in the future, it is way easier to use a ConditionalLink than implementing an if-clause every time we need it. Also, if you know what the ConditionalLink does, it is very easy to use. Maybe we need some documentation here.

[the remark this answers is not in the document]
klinzo: agreed if you add some documentation :P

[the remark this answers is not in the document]
NKlug: Done

src/modules/app/containers/ConditionalLink.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
```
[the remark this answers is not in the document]
NKlug: Done :)

src/modules/common/components/ConditionalLink.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import { Link } from 'redux-little-router'
+import { branch, mapProps, renderComponent } from 'recompose'
+import { omit } from 'lodash/object'
+
+export class DisabledLink extends React.Component {
```
Schedulaar: I think we still mix up the words `dis-/enabled` and `(in)active`, don't we?
I think in this case DisabledLink is the right word (when a link is not even clickable).
An active Link is (for me) a link that indicates (visually) that the current site is the target of the link

NKlug: The problem about this is that we call "Extras" and "News" *enabled* if the locations really has it enabled via the cms. In my opinion that is the right word for it, too. But then if we also called the "News" Link *enabled/disabled* if there are news we would be ambiguous and have naming problems in HeaderNavigationItem. So atm, I call a HeaderNavigationItem active, if it should be shown normal (not grayed out), selected the target matches the current site and enabled if it is enabled via the cms. I agree that we mix the words, but maybe we should then call DisabledLink *InactiveLink*.

Schedulaar: Yes, that's a possible solution.
If you want you can also rename the ExtrasEnabled and NewsEnabled thingy to sth like ExtrasVisible and NewsVisible, maybe then we'll don't have ambigouities in any context (even though enabled may sound better than visible in the cms-context)... But that's up to you :D

NKlug: @klinzo @maxammann What is your opinion on that? :)

klinzo: I think extras/events enabled and selected should stay the way they are, enabled and selected are definitely the right terms for this. accessible would maybe be good imo for whether there are events to show or not

NKlug: I think I'm gonna stick to *active* then and rename DisabledLink

PR #112, opened 2018-03-06
WEBAPP-173: no default values for endpoint models

klinzo:
I don't really know what exactly to do here, as mentioned in #app-web, I don't think throwing an exception if invalid values are passed is necessary or reasonable since the endpoints shouldn't change. Do you have any suggestions? Does using flow make sense here? imo yes because we will benefit whenever we use a withFetcher because we have static types there...but it adds lots of boilerplate to the models in the endpoints package.... feel free to suggest or criticise anything that comes to your mind


maxammann:
Flow should be added incrementally and only if there are no resulting merge conflicts.

Im OK with not validating the data.

src/modules/endpoint/models/CategoryModel.js
```diff
@@ -1,53 +1,77 @@
...
+  title: string,
+  parentId: number,
+  parentUrl: string,
+  content: string,
+  thumbnail: string,
+  order: number,
+  availableLanguages: any
+}
+
 class CategoryModel {
```
Schedulaar: Do we actually need these classes? I think type definition was the sole purpose of them... 🤔

klinzo: nope, good suggestion

src/modules/endpoint/models/EventModel.js
```diff
@@ -1,63 +1,93 @@
+// @flow
+
+import moment from 'moment'
+
+type eventModel = {
```
maxammann: This here looks a bit like duplicate code. So basically eventModel is a EventModel without methods? Looks redundant to me.

src/modules/endpoint/models/EventModel.js
```diff
@@ -1,63 +1,93 @@
...
     return this._allDay
   }
 
-  get excerpt () {
+  get excerpt (): string {
     return this._excerpt
   }
 
-  get availableLanguages () {
+  get availableLanguages (): any {
```
maxammann: Type any shouldn't be used.

src/modules/endpoint/endpoints/__tests__/categories.spec.js
```diff
@@ -48,34 +47,39 @@ describe('categories', () => {
...
-      parentId: 0,
-      parentUrl: '/augsburg/de',
-      order: 11,
-      availableLanguages: {
-        en: 4804, ar: 4819, fa: 4827
-      },
-      thumbnail: 'https://cms.integreat-ap…03/Beratung-150x150.png'
-    }),
-    new CategoryModel({id: 0, url: '/augsburg/de', title: 'augsburg'})
+  const categoryModels = [{
```
maxammann: Types get lost here. Would be nice to have them when initially creating this array.

src/modules/endpoint/models/EventModel.js
klinzo: already in work :D

src/modules/endpoint/endpoints/__tests__/categories.spec.js
klinzo: is it necessary to have types in tests?

maxammann: Tests have to be stable too so I'd say yes. @Schedulaar @NKlug what do you think?

Schedulaar: Hmm... I'm not sure... In this case: If categoryModels was 'misstyped', flow would (should) complain in line 95 (https://github.com/Integreat/integreat-webapp/pull/112/files/231674fb44fb954fe0c7d41676b50ad5b757ce72#diff-d86c55c5be132abd301ff00fb6eee652R95), that categoryModels can't be inserted into `new CategoriesMapModel()`. So I guess we're good to go

src/modules/endpoint/models/EventModel.js
klinzo: so we would need to know all possible locales and optionally add them to availableLanguages type :/

klinzo: removed, although we now have all possible languages hard coded

src/modules/endpoint/endpoints/__tests__/categories.spec.js
klinzo: I could create a task to add flow to all tests @maxammann ?

src/modules/endpoint/types.js
```diff
@@ -0,0 +1,46 @@
...
+  ku?: string,
+  ru?: string,
+  ti?: string,
+  sr?: string,
+  am?: string,
+  per?: string,
+  pl?: string
+}
+
+export type EventType = {
```
maxammann: I think the models should stay classes. I'm not that sure what the real difference is between a Type Alias and a Class Type. Maybe our models will have (or had) methods which are not possible to add with flow types.

None the less each type/class should belong to a separate file. Also we should name this Event or EventModel. You would not name it an EventClass.

src/modules/layout/containers/LocationLayout.js
```diff
@@ -13,19 +11,22 @@ import GeneralFooter from '../components/GeneralFooter'
...
-    locations: PropTypes.arrayOf(PropTypes.instanceOf(LocationModel)).isRequired,
-    currentPath: PropTypes.string.isRequired,
-    viewportSmall: PropTypes.bool.isRequired,
-    children: PropTypes.node,
-    events: PropTypes.arrayOf(PropTypes.instanceOf(EventModel)).isRequired
-  }
+import type { EventType } from '../../endpoint/types'
+import Route from '../../app/Route'
+
+type Props = {
```
maxammann: As we no longer use prop-types, does it check in runtime that all props are passed? Or is this Props thing a compile-time feature?

src/modules/endpoint/models/CategoriesMapModel.js
```diff
@@ -1,49 +1,62 @@
...
 
   /**
    * Returns the category with the given id
-   * @param id The url
+   * @param id The id
    * @return {CategoryModel | undefined} The category
    */
-  getCategoryById (id) {
-    return this.toArray().find(category => category.id === Number(id))
+  getCategoryById (id: number): ?CategoryModel {
```
Schedulaar: Return type is not optional

src/modules/layout/containers/__tests__/LocationLayout.spec.js
```diff
@@ -17,22 +16,21 @@ describe('LocationLayout', () => {
...
-      allDay: true
-    }),
-    new EventModel({
-      id: 2,
-      title: 'second Event',
-      startDate: moment.tz('2017-11-18 09:30:00', 'UTC'),
-      endDate: moment.tz('2017-11-18 19:30:00', 'UTC'),
-      allDay: true
-    })]
+  const events = [{
```
Schedulaar: Why don't we use constructor EventModel here?

src/routes/events/components/__tests__/EventDetail.spec.js
```diff
@@ -2,18 +2,17 @@ import React from 'react'
 import { shallow } from 'enzyme'
 import moment from 'moment-timezone'
 
-import EventModel from 'modules/endpoint/models/EventModel'
 import EventDetail from '../EventDetail'
 
 describe('EventDetail', () => {
-  const event = new EventModel({
+  const event = {
```
Schedulaar: Same here

src/routes/events/components/__tests__/EventList.spec.js
```diff
@@ -2,34 +2,33 @@ import React from 'react'
 import { shallow } from 'enzyme'
 import moment from 'moment-timezone'
 
-import EventModel from 'modules/endpoint/models/EventModel'
 import EventList from '../EventList'
 
 describe('EventList', () => {
   const events = [
```
Schedulaar: And here

src/routes/events/components/__tests__/EventListElement.spec.js
```diff
@@ -2,18 +2,18 @@ import React from 'react'
 import { shallow } from 'enzyme'
 import moment from 'moment-timezone'
 
-import EventModel from 'modules/endpoint/models/EventModel'
 import EventListElement from '../EventListElement'
 
 describe('EventListElement', () => {
-  const event = new EventModel({
+  const event = {
```
Schedulaar: and here

src/routes/events/containers/__tests__/EventsPage.spec.js
```diff
@@ -9,34 +9,33 @@ import EndpointBuilder from 'modules/endpoint/EndpointBuilder'
 import EndpointProvider from 'modules/endpoint/EndpointProvider'
 
 import ConnectedEventsPage, { EventsPage } from '../EventsPage'
-import EventModel from 'modules/endpoint/models/EventModel'
 import LanguageModel from 'modules/endpoint/models/LanguageModel'
 
 describe('EventsPage', () => {
   const events = [
```
Schedulaar: and here

src/routes/search/containers/__tests__/SearchPage.spec.js
```diff
@@ -21,12 +20,16 @@ describe('SearchPage', () => {
   ]
 
   const categoryModels = [
```
Schedulaar: and here

src/routes/search/containers/__tests__/SearchPage.spec.js
```diff
@@ -130,33 +131,33 @@ describe('SearchPage', () => {
   it('should sort correctly', () => {
     const categoryModels = [
```
Schedulaar: and here

PR #113, opened 2018-03-07
WEBAPP-198: Make tests fail if there is a missing prop

maxammann:


src/modules/endpoint/__tests__/Endpoint.spec.js
```diff
@@ -89,7 +89,7 @@ describe('Endpoint', () => {
 
     beforeEach(() => {
       clock = lolex.install({now: mockedTime, toFake: []})
-      prevError = console.error
+      prevError = console.error // todo: Find better way of allowing console.error
```
maxammann: For now this should be enough. We should keep an eye on this if we really want that or if there is a better way.

src/modules/layout/components/__tests__/LocationHeader.spec.js
```diff
@@ -96,7 +105,10 @@ describe('LocationHeader', () => {
     const component = shallow(<LocationHeader matchRoute={matchRoute} language={language}
                                               locationModel={createLocation(true, true)}
                                               currentPath=''
-                                              viewportSmall />)
+                                              viewportSmall
+                                              eventCount={0} />)
     expect(component.dive()).toMatchSnapshot()
   })
+
+  // fixme: Test the events enabled functionality. Especially isEventsActive()
```
maxammann: @NKlug eventCount was missing in this test. Some more tests are needed here to keep the coverage up :)

NKlug: I'll have a look at it :)

PR #114, opened 2018-03-08
WEBAPP-224: Fix Language Failure

Schedulaar:
Now the EventsNavigationItem fetches data itself, so we also avoid that the screen is flashing on language change (Because the whole header had to wait for the events data)


src/modules/layout/containers/EventsNavigationItem.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  render () {
+    const {href, events, text, tooltip, selected} = this.props
+    return <HeaderNavigationItem href={href} selected={selected} text={text} tooltip={tooltip}
+                                 active={!isEmpty(events)} />
+  }
+}
+
+export default withFetcher('events', EventsNavigationItem, EventsNavigationItem)(EventsNavigationItem)
```
klinzo: Why do you display an EventsNavigationItem even if there is an error?

src/modules/layout/components/HeaderNavigationItem.js
```diff
@@ -0,0 +1,40 @@
...
+    /** the message to be displayed when the item is hovered */
+    tooltip: PropTypes.string
+  }
+
+  render () {
+    const {active, text, tooltip, selected, href} = this.props
+    return <React.Fragment>
+      <ConditionalLink active={active} data-tip={!active ? tooltip : ''}
+                       key={text}
+                       className={cx(style.navigationItem, !active ? style.inactiveNavigationItem : (selected ? style.selectedNavigationItem : ''))}
```
klinzo: I think you could change this to 
```js
className={cx({style.navigationItem, [style.inactiveNavigationItem]: !active, [style.selectedNavigationItem]: active && selected})}
```
to stay consistent

src/modules/layout/components/HeaderNavigationBar.js
```diff
@@ -1,38 +1,26 @@
...
-        <ConditionalLink active={item.active} data-tip={!item.active ? item.tooltip : ''}
-                         key={item.text}
-                         className={cx(style.navigationItem, !item.active ? style.inactiveNavigationItem : (item.selected ? style.selectedNavigationItem : ''))}
-                         href={item.href}>{item.text}
-        </ConditionalLink>
-      ))}
-
-      <ReactTooltip place='top' type='dark' effect='solid' />
-
+      className={cx(this.props.className, style.navigationBar, isEmpty(this.props.children) ? style.hidden : '')}>
```
klinzo: you could change this to 
```js
className={cx({this.props.className, style.navigationBar, [style.hidden]: isEmpty(this.props.children})}

File: src/modules/layout/components/Header.js [hunk 4]
```diff
@@ -39,7 +38,7 @@ class Header extends React.Component {
               <img src={logoWide} />
             </Link>
           </div>
-          <HeaderNavigationBar className={style.navigationBar} items={this.props.navigationItems} />
+          <HeaderNavigationBar className={style.navigationBar}>{this.props.navigationItems}</HeaderNavigationBar>
```
klinzo: Is it necessary to pass a custom className?

maxammann:
I think some parts of this will conflict with #115. Discuss with @klinzo how we should proceed here.

src/modules/layout/containers/EventsNavigationItem.js
Schedulaar: It's a general  issue of how we should proveed if an error occurs... But since Extras and Categories NavigationItems are also still visible, it would be consistent (in terms of user interface) if the Events are also still visible....

src/modules/layout/components/Header.js
Schedulaar: Yes, because these are styles that are applied to position the navigationBar and the actionBar inside the header and therefore belongs to the header and not to the bars themselves

Schedulaar:
@maxammann @klinzo I don't think we've made not mergeable changes and the conflicts will be resolved by the last to merge.

PR #115, opened 2018-03-10
WEBAPP-223: Refactor withFetcher

klinzo:
I removed the old withFetcher, endpoints, builders and so on and created a imo very easy new withFetcher, without any depencies to the redux state. I had to refactor the RouterFragment and split it, so that the fetched data can be passed to all routes and the locationLayout.
Currently there are still some problems:
1) /augsburg redirects to /augsburg/de, which is fine, but for some reason also the Failure component gets displayed, which is the forNoMatch fragment in RouterFragment
2) I can't distinguish between /:location and /:language (location should be a valid location name, language should be a language code, which means a string with two or three chars, because of this the LocationSelectionButton in the Header is not working
3) LanuageChange in deeper levels of categories are not working yet, also because the routes are not working the way I want them to
4) Some tests are missing and the test of LocationLayout > connect is failing and I have no idea why

Because of problems 1 to 3 it would maybe really make sense to switch from redux-little-router to redux-first-router (also because RouterFragment, LanguageFragment and LocationFragment are quite complicated right now)


maxammann:
Why does switching to redux-first-router fix the problems 1 to 3?
Also should we already start commenting on the changes? :)

src/modules/app/containers/LanguageFragment.js
```diff
@@ -0,0 +1,110 @@
...
+    routeConfig: PropTypes.instanceOf(RouteConfig).isRequired,
+    location: PropTypes.string.isRequired,
+    language: PropTypes.string.isRequired
+  }
+
+  render () {
+    const {locations, languages, location, language, routeConfig} = this.props
+    const params = {location: location, language: language}
+
+    const LanguageFragmentWithFetchers = compose(
```
maxammann: HOCs should not be used within render(). This is not supported by react and can cause problems.

package.json
```diff
@@ -80,7 +80,6 @@
     "git-repo-version": "~1.0.1",
     "highlight.js": "~9.12.0",
     "html-loader": "~0.5.1",
-    "husky": "^0.14.3",
```
maxammann: Why did you remove this? Was it too annoying in your experience? We could change it so it runs on prepush.

src/modules/app/containers/LocationFragment.js
```diff
@@ -0,0 +1,61 @@
...
+      </Fragment>
+
+      <Fragment forNoMatch>
+        <LanguageFailure languages={languages} locations={locations} />
+      </Fragment>
+    </React.Fragment>
+  }
+}
+
+class Container extends React.Component {
```
maxammann: There should be only one class per file.

src/modules/endpoint/hocs/withFetcher.js
```diff
@@ -1,127 +1,40 @@
...
-        const storeResponse = this.props.requestAction()
-        this.setState({isDataAvailable: storeResponse.dataAvailable})
-      }
-
-      render () {
-        const payload = this.getStatePayload()
+const withFetcher = (name, urlMapper, mapper, params, FailureComponent = Failure) => WrappedComponent =>
+  class Fetcher extends React.Component {
+    constructor (props) {
+      super(props)
```
maxammann: You don't need to pass the props afaik. You can just remove them from the super and constructor.

src/modules/endpoint/mappers/__tests__/categories.spec.js
```diff
@@ -1,4 +1,4 @@
-import categories from '../categories'
+import categoriesMapper from '../categories'
```
maxammann: Naming is a bit off here. is it a 'categories' or a 'categoriesMapper'?

src/modules/endpoint/mappers/categories.js
```diff
@@ -0,0 +1,36 @@
+import CategoryModel from '../models/CategoryModel'
+import CategoriesMapModel from '../models/CategoriesMapModel'
+
+function categoriesMapper (json, params) {
```
maxammann: We try to avoid function declarations. Use a lambda instead. I think I forgot to add a linter rule for this :see_no_evil:

src/modules/endpoint/EndpointProvider.js
```diff
@@ -1,40 +0,0 @@
-import React from 'react'
-import PropTypes from 'prop-types'
-import Endpoint from './Endpoint'
-
-class EndpointProvider extends React.Component {
```
maxammann: How did you solve the mocking of endpoints without the EndpointProvider?

src/modules/app/containers/LanguageFragment.js
```diff
@@ -0,0 +1,110 @@
+import { Fragment } from 'redux-little-router'
```
maxammann: This file will need tests now because we need to test the withFetcher calls. the fetch(...) call needs to get mocked then.

maxammann:
I just submitted my review. Good work so far! Looks nice :)

I just think some tests are missing. Also do you want to do all of this with Flow? I think we should use flow from now on. Especially in citical parts like the Fetcher.

maxammann:
I found these writeups some time ago: https://gist.github.com/Vermeille/6660f2371bd272ce85872c3e7012e4fb

I think we are quite on the same track just with HOCs which is much nicer :)

src/modules/app/createReduxStore.js
```diff
@@ -33,7 +32,6 @@ const createReduxStore = (createHistory, initialState = {}, routes = new RouteCo
   }
 
   const reducer = combineReducers({
-    ...endpointReducers,
```
maxammann: Do we have any additional reducers/actions we need now because the models are no longer in the redux-state?

maxammann:
Maybe this is also interresting for some endpoints: https://medium.com/stashaway-engineering/react-redux-tips-better-way-to-handle-loading-flags-in-your-reducers-afda42a804c6

package.json
klinzo: yep, especially during merging... I didn't want to commit the changes though. Prepush would be fine

src/modules/endpoint/mappers/categories.js
klinzo: I thought we try to avoid lambdas :D will be changed

src/modules/endpoint/EndpointProvider.js
klinzo: I haven't tested the endpoints yet, so I didn't have to mock anything.

src/modules/app/containers/LanguageFragment.js
klinzo: yep sure, as I said tests are missing and this will require some more work. I wouldn't include this in our release.

klinzo:
With redux-first-router you can really decide which 'route' you want to use next and differ more, because it's handled by actions. I think I will close this pr until everything is working, I just wanted to know if this in general is the right direction to go with the fetching.

package.json
Schedulaar: Im fine with this removal, too :D

src/modules/endpoint/hocs/withFetcher.js
```diff
@@ -1,127 +1,40 @@
...
-       * Tries to trigger a request by dispatching a requestAction
-       */
-      tryRequest () {
-        const storeResponse = this.props.requestAction()
-        this.setState({isDataAvailable: storeResponse.dataAvailable})
-      }
-
-      render () {
-        const payload = this.getStatePayload()
+const withFetcher = (name, urlMapper, mapper, params, FailureComponent = Failure) => WrappedComponent =>
```
Schedulaar: I dislike that you discarded the idea of the Endpoint class: Having the name, urlMapper and jsonMapper in one place is a benefit imo. Also the params shouldn't be part of the withFetcher declaration, since these are likely to change and HOCs shouldn't be called too often (especially not in the render method).
We could think about extracting the params into the props of the Fetcher, sth like  `fetcherParams` and the usage would then look like this:
```jsx
class SomeComponent extends React.Component {
 // Need some fetched data
}
const SomeFetchingComponent = withFetcher(categoriesEndpoint)

class SomeWrapperComponent {
  render () {
    return <SomeFetchingComponent fetcherParams={{location: 'augsburg', language: 'de'}} />
  }
}
```
If we want to connect that with our littleRouter, we could then also write
```jsx
class SomeComponent extends React.Component {
  // ...
}
export default compose(
  connect(state => ({ fetcherParams: { location: state.router.params.location, language: state.router.params.language }}),  
  withFetcher(categoriesEndpoint)
)
```

Schedulaar:
Generally I think, before we apply changes to the way we fetch our data, we should think about the upcoming cache-feature... Reducing the complexity now doesn't make sense when we need to add it later again anyways.
Also (since I'm currently traveling with Deutsche Bahn and experencing (very) very bad wifi connection) I would still prefer that we have some place (maybe in an endpoint class itself and not in the state) where we can store the data we have already fetched, so we can have multiple fetchers in parallel but only fetch the data once.

maxammann:
@Schedulaar This probably won't geht merged before the next release so we can test how much caching we need. But we should not optimze pre-maturely. I think not much will get refetched during normal usage of the app.

If we cache then it probably should go into the state because of e.g. the concept of time-travel-debugging.

The discussion about the offljne-functionality in cordova and the "HTML to jsx" is also important when talking about caching.

maxammann:
@schedulaar I think it makes more sense to just use the whole props instead of the fetcherProps. Using the props makes the hoc easier to integreat with other hocs and libs.

PR #116, opened 2018-03-13
WEBAPP-224: Add Caption to ExtrasPage

Schedulaar:


src/routes/extras/containers/ExtrasPage.js
```diff
@@ -85,6 +92,7 @@ const mapDispatchToProps = dispatch => ({
 })
 
 export default compose(
+  translate('layout'),
```
maxammann: Change the i18n namespace to extras here.

PR #121, opened 2018-03-14
WEBAPP-227: Fix a bug in .flowconfig

maxammann:


src/routes/extras/components/SprungbrettList.js
```diff
@@ -1,23 +1,25 @@
...
 import SprungbrettJobModel from 'modules/endpoint/models/SprungbrettJobModel'
 
 import SprungbrettListItem from './SprungbrettListItem'
 import style from './SprungbrettList.css'
 
-class SprungbrettList extends React.Component {
-  static propTypes = {
-    jobs: PropTypes.arrayOf(PropTypes.instanceOf(SprungbrettJobModel)).isRequired
-  }
+type Prop = {
```
klinzo: shouldn't we stay consistent and name this Props? or do we want to use Prop if there is only one prop? I'm pro props :D

PR #122, opened 2018-03-15
WEBAPP-49: Redesign PDF

Schedulaar:

~**I had to make the layout HOCs, to get the right behavior with the stickyTop thingy :/ Also each route now determines its layout itself (I think that makes sense) - that also solved the "problem" with the PDF-FetcherPage... Tests still missing, but I (and Nick) will wait for your first feedback. Fun fact:**~

~matchRoute was previously given the xxxxPage as input, but when you need to render a link in, lets say, CategoriesPage targeting itself, then, inside the render() of CategoriesPage, you would want to say matchRoute(CategoriesPage), but cannot, since you're currently defining CategoriesPage^^ So I changed the identification to strings~


src/locales.json
```diff
@@ -70,6 +70,18 @@
       }
     }
   },
+  "categories": {
+    "ar": {
+    },
+    "de": {
+      "createPdf": "Erstelle PDF"
```
NKlug: Is "categories" the right place for this? We maybe want to display the toolbar with the PDFButton on other pages, too, not only on the categories page.

src/modules/layout/components/Toolbar.css
```diff
@@ -0,0 +1,28 @@
...
+  box-sizing: border-box;
+  flex-direction: column;
+  align-items: center;
+  padding: 15px 0;
+}
+
+.toolbar > * {
+  opacity: 0.5;
+  font-size: 1.5rem;
+  transition: 0.2s opacity;
```
NKlug: In my opinion opacity 0.5 looks a bit too much "inactive", just like the News Button when there are no news. Maybe we should also use the same value for the transition time and the tooltip delay-show.

src/modules/layout/components/Toolbar.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+import React from 'react'
+import type { Node } from 'react'
+import cx from 'classnames'
+import style from './Toolbar.css'
+
+type Props = {
+  className?: string,
+  children?: Node
```
NKlug: Shouldn't this be `Array<Node>`?

src/modules/app/createRouteConfig.js
```diff
@@ -1,47 +1,39 @@
...
-import EventsPage from 'routes/events/containers/EventsPage'
 import Route from './Route'
 
 const createRouteConfig = () => [
   new Route({
     path: '/'
   }),
   new Route({
-    id: LandingPage,
+    id: 'landing',
```
NKlug: Is there no better way than hard-coding the id as a string? If not, we should maybe define the ids in a central place in some kind of map and import it whenever we need it.

src/routes/disclaimer/containers/DisclaimerPage.js
```diff
@@ -44,6 +45,7 @@ const mapDispatchToProps = dispatch => ({
 })
 
 export default compose(
+  withLocationLayout(null),
```
maxammann: What is `null` in this case? Couldn't get it easily by looking at the implementation.

src/routes/categories/containers/CategoriesToolbar.js
```diff
@@ -0,0 +1,52 @@
...
+  getPdfFetchPath () {
+    return `/${this.props.location}/${this.props.language}/fetch-pdf?url=${this.props.path}`
+  }
+
+  render () {
+    try {
+      this.props.categories.getCategoryByUrl(this.props.path)
+      return <Toolbar>
+        <ToolbarItem name='file-pdf-o' text={this.props.t('createPdf')} href={this.getPdfFetchPath()} />
+        {/* todo: Add these functionalities:
```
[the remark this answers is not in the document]
Schedulaar: I won't add favorites, share and text-to-speech in this task :P

src/modules/layout/hocs/withLayout.js
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * The standard Layout, used for any view in this app as a container.
+ * If a footer is supplied and there's not enough content (in header and children) to fill the viewbox, the footer will
+ * always stick to the bottom of the viewbox.
+ */
+const withLayout = (
+  Header: ComponentType<{ onStickyTopChanged: number => void }>,
+  Toolbar: ?ComponentType<{}>,
+  Footer: ComponentType<{}>
```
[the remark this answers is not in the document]
Schedulaar: Not sure what you mean by that, could you add a demo usage example? 🤔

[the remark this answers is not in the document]
maxammann: So we could add a
* withHeader() which adds a header
* withFooter() which adds a footer
* withToolbar() which adds a toolbar

~~Not sure if that is possible but the withLocationLayout call should be documented as it's not easy to see what the first parameter does because of the recompose library. I'm not sure if you could see it easily without it.~~

src/routes/disclaimer/containers/DisclaimerPage.js
maxammann: Ok, got it now :P

src/modules/app/createRouteConfig.js
maxammann: Is there a reason for this change? By using the class as identifier you avoid typos.

Schedulaar: see pr description

Schedulaar: But defining these somewhere as constants makes sense to me

maxammann: ok damn need to read the damn description :P

ok makes sense because we used the connected component as id.
Maybe we can think about using flow enums here.

maxammann: See https://flow.org/en/docs/types/utilities/#toc-keys (last example)

src/modules/layout/hocs/withLayout.js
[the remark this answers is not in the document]
Schedulaar: But the layout is important to stick these three things together with the correct styling...
Your idea would be
```jsx
const CategoriesPage = ....
export default compose(
  withToolbar(CategoriesToolbar),
  withHeader(LocationHeader),
  withFooter(LocationFooter),
  withFetcher(...),
  ...
)
```
? I don't know how we could even implement this...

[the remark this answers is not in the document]
maxammann: So if those three depend on each other then ofc leave it as one Layout. But it would be cool if they wouldn't. This would remove a lot of if-else-then code. But we don't need to do this,

Schedulaar:
~Missing: Some tests,
Readd stickabilty to toolbar~
wrong pr :D

maxammann:
@Schedulaar Is this still blocked by the toolbar or is this ready for review?

Schedulaar:
@maxammann is ready for review now

src/modules/app/createRouteConfig.js
NKlug: @Schedulaar Can you do something about this?

Schedulaar: This only affected the old router config, which has been removed with the last changes of @klinzo :D

klinzo:
I really like your changes, good work. One thing though: If you're scrolling down, the header vanishes and the toolbar is visible. If you scroll back up a little bit, the header pops back in and hides the toolbar, which will only be visible again if you scroll back to the top of the page.

Schedulaar:
@klinzo Woops, indeed... Seems i broke it again :D ... will check that

Schedulaar:
@klinzo should work again, I also added a test :D

src/modules/layout/components/Toolbar.js
maxammann: I think Node also can be an array of nodes. But not sure.

src/modules/layout/components/Toolbar.css
maxammann: Still open for discussion.

.eslintrc.json
```diff
@@ -80,7 +80,8 @@
             "ignore": [
               -1,
               0,
-              1
+              1,
+              2
```
maxammann: :+1:

src/modules/app/constants/dimensions.css
```diff
@@ -1,6 +1,9 @@
 @custom-media --small-viewport (max-width: 750px);
+@custom-media --min-max-width screen and (min-width: 1100px);
```
maxammann: Not related to this task, but are there any official recommendations for the min-width values. If not then these are also magic values.

src/modules/app/constants/dimensions.js
```diff
@@ -1,5 +1,8 @@
 export default {
   smallViewport: `(max-width: 750px)`,
-  headerHeight: '85px',
-  headerHeightSmall: '55px'
+  minMaxWidth: `screen and (min-width: 1100px)`,
```
maxammann: Is this validated by styled-components?

src/modules/layout/components/Layout.js
```diff
@@ -1,33 +1,45 @@
...
         <div>
-          {this.props.header}
-          <main className={style.layout}>
-            <div className={cx(style.content)}>
-              {this.props.children}
-            </div>
-          </main>
+          {header}
+          <Body>
+            <Aside style={{top: `${asideStickyTop}px`}}>
```
maxammann: :+1:

src/modules/app/constants/dimensions.css
Schedulaar: Well, comes down to how `official` you want :D Theres no number in the html/css spec, but there are probably some suggested values for them in guides by google or other companies...
I think we should set them on how they best fit our design...

src/modules/app/constants/dimensions.js
Schedulaar: Hmm... the template-string itself not no, could be validated once its injected in styled-components-stylesheets 🤔 not sure though

src/modules/app/constants/dimensions.css
maxammann: Yeah I'm thinking more about good recommendations :D

PR #123, opened 2018-03-16
WEBAPP-223: Remove withFetcher, change to Redux-First-Router

klinzo:

Most of the work should be done now (if you exclude testing of course :D). I don't know of any bugs left, but there are some for sure, so please try and play with it.
I hope this makes some things way more easy. 

Known issues:
1) Tests are missing (@maxammann) and failing
2) Displaying of errors can be improved, there are just dummies for now. (Will be added in a other task)

Regarding the tests: I would appreciate it if you could give me a rough feedback, before testing the whole stuff.
Also some naming/folder structure could be bad. :D


klinzo:
The landing route is now /landing/:language. Are there better suggestions?

src/routes/extras/containers/ExtrasPage.js
```diff
@@ -60,19 +42,21 @@ export class ExtrasPage extends React.Component {
...
-    if (this.props.extra === SPRUNGBRETT_EXTRA && sprungbrett) {
-      return <SprungbrettPage title={sprungbrett.name} />
-    } else if (this.props.extra) {
+    const {extraAlias, extras, sprungbrettJobs} = this.props
+
+    if (extraAlias) {
+      const extra = extras.find(_extra => _extra.alias === extraAlias)
+
+      if (extra && extraAlias === SPRUNGBRETT_EXTRA) {
+        return sprungbrettJobs ? <SprungbrettList title={extra.name} jobs={sprungbrettJobs} /> : <LoadingSpinner />
```
Schedulaar: What happens if there occurs an error while fetching? 🤔

src/routes/landing/components/LocationSelector.js
```diff
@@ -21,7 +21,7 @@ class LocationSelector extends React.PureComponent {
 
   filter () {
```
Schedulaar: Maybe you could just apply the following refactor:
```jsx
filter () {
  const filterText = this.props.filterText.toLowerCase()
  if (filterText === 'wirschaffendas') {
    return locations.filter(location => !location.live)
  } else {
    return locations.filter(location => location.live)
      .filter(location => location.name.toLowerCase().includes(filterText))
  }
}
```

src/modules/app/containers/I18nProvider.js
```diff
@@ -24,35 +21,11 @@ export class I18nProvider extends React.Component {
...
-   */
-  static transformResources (resources) {
-    return reduce(resources, (accumulator, namespace, namespaceName) => {
-      forEach(namespace, (language, languageCode) => {
-        accumulator[languageCode] = {...accumulator[languageCode], [namespaceName]: language}
-      })
-      return accumulator
-    }, {})
-  }
-
```
Schedulaar: Fonts don't load for me

src/modules/app/containers/Switcher.js
```diff
@@ -0,0 +1,130 @@
...
+          return <SearchPage />
+        }
+        break
+      case PDF_FETCHER_ROUTE:
+        if (categories) {
+          return <PdfFetcherPage />
+        }
+        break
+    }
+    return <LoadingSpinner />
```
Schedulaar: This doesn't deal with errors that might occur while loading.
(And the spinner isn't displayed correctly)
Also I'd like it more if the routePages themselves can decide what kind of resources they need

src/modules/app/routes/mainDisclaimer.js
```diff
@@ -0,0 +1,24 @@
...
+ * MainDisclaimerRoute, matches /disclaimer
+ * @type {{path: string, thunk: function(Dispatch, GetState)}}
+ */
+export const mainDisclaimerRoute = {
+  path: '/disclaimer',
+  thunk: (dispatch: Dispatch, getState: GetState) => {
+    const prev = getState().location.prev
+
+    if (prev.payload.city) {
+      clearStoreOnCityChange(dispatch, getState)
```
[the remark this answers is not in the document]
NKlug: +1

src/modules/common/containers/LanguageSelector.js
```diff
@@ -1,24 +1,31 @@
-import React from 'react'
+import * as React from 'react'
```
NKlug: Is the wildcard import necessary?

src/modules/endpoint/models/__tests__/LocationModel.spec.js
```diff
@@ -1,4 +1,4 @@
-import LocationModel from '../LocationModel'
+import LocationModel from '../CityModel'
```
NKlug: Should be `import CityModel `, right?

src/modules/layout/containers/LocationLayout.js
```diff
@@ -1,63 +1,68 @@
...
+    const cities = this.props.cities
+    return cities && cities.find(_city => _city.code === this.props.city)
   }
 
   render () {
-    const {language, location, currentPath, matchRoute, viewportSmall, children} = this.props
-    const locationModel = this.getCurrentLocation()
+    const {language, city, currentRoute, viewportSmall, children, events, languages} = this.props
+    const cityModel = this.getCurrentCity()
+    const isEventsActive = events ? events.length > 0 : false
```
NKlug: `const isEventsActive = events.length > 0`

src/routes/categories/containers/LanguageFailure.js
```diff
@@ -4,45 +4,22 @@ import React, { Fragment } from 'react'
 import { translate } from 'react-i18next'
 import compose from 'lodash/fp/compose'
 import { connect } from 'react-redux'
-import withFetcher from 'modules/endpoint/hocs/withFetcher'
 
 import LanguageSelector from 'modules/common/containers/LanguageSelector'
-import setLanguageChangeUrls from 'modules/language/actions/setLanguageChangeUrls'
 import LanguageModel from 'modules/endpoint/models/LanguageModel'
-import LocationModel from 'modules/endpoint/models/LocationModel'
+import LocationModel from 'modules/endpoint/models/CityModel'
```
NKlug: `CityModel`

src/routes/categories/containers/__tests__/CategoriesPage.spec.js
```diff
@@ -4,11 +4,9 @@ import { Provider } from 'react-redux'
 
 import createHistory from 'modules/app/createHistory'
 import createReduxStore from 'modules/app/createReduxStore'
-import EndpointBuilder from 'modules/endpoint/EndpointBuilder'
-import EndpointProvider from 'modules/endpoint/EndpointProvider'
 
 import ConnectedCategoriesPage, { CategoriesPage } from '../CategoriesPage'
-import LocationModel from 'modules/endpoint/models/LocationModel'
+import LocationModel from 'modules/endpoint/models/CityModel'
```
NKlug: `CityModel` again :)

src/routes/categories/containers/__tests__/LanguageFailure.spec.js
```diff
@@ -4,10 +4,8 @@ import { Provider } from 'react-redux'
 
 import createHistory from 'modules/app/createHistory'
 import createReduxStore from 'modules/app/createReduxStore'
-import EndpointBuilder from 'modules/endpoint/EndpointBuilder'
-import EndpointProvider from 'modules/endpoint/EndpointProvider'
 
-import LocationModel from 'modules/endpoint/models/LocationModel'
+import LocationModel from 'modules/endpoint/models/CityModel'
```
NKlug: `CityModel` :)

src/routes/landing/components/FilterableLocationSelector.js
```diff
@@ -4,13 +4,13 @@ import PropTypes from 'prop-types'
 import Heading from './Heading'
 import ScrollingSearchBox from 'modules/common/components/ScrollingSearchBox'
 import LocationSelector from './LocationSelector'
-import LocationModel from 'modules/endpoint/models/LocationModel'
+import LocationModel from 'modules/endpoint/models/CityModel'
```
NKlug: `CityModel`

src/routes/landing/components/LocationSelector.js
```diff
@@ -4,12 +4,12 @@ import PropTypes from 'prop-types'
 import style from './LocationSelector.css'
 import { transform } from 'lodash/object'
 import { groupBy, filter } from 'lodash/collection'
-import LocationModel from 'modules/endpoint/models/LocationModel'
+import LocationModel from 'modules/endpoint/models/CityModel'
```
NKlug: `CityModel`

src/routes/landing/containers/LandingPage.js
```diff
@@ -1,30 +1,26 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
-import compose from 'lodash/fp/compose'
 
 import FilterableLocationSelector from 'routes/landing/components/FilterableLocationSelector'
-import withFetcher from 'modules/endpoint/hocs/withFetcher'
-import LocationModel from 'modules/endpoint/models/LocationModel'
+import LocationModel from 'modules/endpoint/models/CityModel'
```
NKlug: .

src/modules/app/constants.js
```diff
@@ -1 +1,2 @@
-export const LANGUAGE_CODE_LENGTH = 2
+export const MIN_LANGUAGE_CODE_LENGTH = 2
```
maxammann: I think we could try to export whole enums: https://flow.org/en/docs/types/utilities/#toc-keys
Should we create a task for this?

src/modules/app/onBackNext.js
```diff
@@ -0,0 +1,33 @@
...
+import { categoriesFetcher } from '../endpoint/fetchers'
+
+/**
+ * This handles the use of the back/ next button of the browser. We have to remove and refetch the categories in the old
+ * language, because the stopover of CategoriesRedirect, during which the refetch normally happens, is not executed again
+ * when using the back/ next button.
+ * @param dispatch
+ * @param getState
+ * @param bag
+ * @return {Promise<void>}
```
maxammann: You can remove this params returns or add docu :P

src/modules/app/routes/categories.js
```diff
@@ -0,0 +1,39 @@
+// @flow
```
maxammann: Can we have the categories.js, dislaimer.js somehow local in the routes/ directory? we could maybe put these in a Route.js in each package of a route.

src/modules/app/routes/categories.js
```diff
@@ -0,0 +1,39 @@
...
+
+    if (prev.payload.language && prev.payload.language !== language) {
+      clearStoreOnLanguageChange(dispatch, getState)
+    }
+
+    if (prev.payload.city && prev.payload.city !== city) {
+      clearStoreOnCityChange(dispatch, getState)
+    }
+
+    await locationLayoutFetcher(dispatch, getState)
```
maxammann: I think we shouldn't load there the data which is neccassary for the location layout. Can we maybe have this a bit more local to the definition of <LocationLayout>?

src/modules/endpoint/fetchers/categories.js
```diff
@@ -0,0 +1,71 @@
+// @flow
```
maxammann: I really liked the Endpoint.js and Endpoint.js builder to give it more structure. There's a lot of freedom you have now and not a real schema. For example you have to call `.json()` a lot and have to come up with a way of mapping urls with `urlMapper`

src/modules/endpoint/fetchers/extras.js
```diff
@@ -0,0 +1,30 @@
...
+    .then(result => result.json())
+    .then(json => json
+      .map(extra => new ExtraModel({
+        alias: extra.alias,
+        name: extra.name,
+        path: extra.url,
+        thumbnail: extra.thumbnail
+      })))
+    .then(extras => {
+      dispatch(saveExtras(extras))
```
maxammann: I think this is more complicated as it needs to be. It's absolutly not clear which action triggers which other actions and results in which store diff.

src/modules/endpoint/fetchers/index.js
```diff
@@ -0,0 +1,8 @@
+export { default as categoriesFetcher } from './categories'
```
maxammann: I don't think we need a file to import and export files which are in the corrosponing directory. Does this make things easier?

src/modules/app/routes/landing.js
```diff
@@ -0,0 +1,30 @@
...
+export const LANDING_ROUTE = 'LANDING'
+
+export const goToLanding = (language: string) => createAction(LANDING_ROUTE)({language})
+
+/**
+ * LandingRoute, matches /landing/de
+ * @type {{path: string, thunk: function(Dispatch, GetState)}}
+ */
+export const landingRoute = {
+  path: '/landing/:language',
```
maxammann: So is '/' no longer the landing page? I'm also a bit agains introducting a new route if we already have one. We should keep the set of routes minimal.

maxammann:
Also this diff is way too big. It's really difficult to do a review on this.

src/modules/common/containers/LanguageSelector.js
```diff
@@ -1,24 +1,31 @@
...
 export class LanguageSelector extends React.Component {
   static propTypes = {
     closeDropDownCallback: PropTypes.func,
-    languages: PropTypes.arrayOf(PropTypes.instanceOf(LanguageModel)).isRequired,
-    language: PropTypes.string.isRequired,
-    location: PropTypes.string.isRequired,
-    languageChangeUrls: PropTypes.object.isRequired,
-    verticalLayout: PropTypes.bool
+    languages: PropTypes.arrayOf(PropTypes.instanceOf(LanguageModel)),
+    location: PropTypes.object.isRequired,
```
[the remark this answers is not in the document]
klinzo: should be done when adding flow

src/modules/layout/components/HeaderNavigationItem.js
```diff
@@ -14,7 +14,7 @@ class HeaderNavigationItem extends React.PureComponent {
     /** text to be displayed */
     text: PropTypes.string.isRequired,
     /** link to the page that should be shown when the item is clicked */
-    href: PropTypes.string.isRequired,
+    href: PropTypes.object.isRequired,
```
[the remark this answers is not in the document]
klinzo: because redux-first-router links are using actions to link to different pages, see https://github.com/faceyspacey/redux-first-router-link. Should also be refactored to type: Action when flow is introduced here

src/modules/layout/components/LocationHeader.js
```diff
@@ -0,0 +1,84 @@
...
+import LanguageModel from '../../endpoint/models/LanguageModel'
+import { EXTRAS_ROUTE, goToExtras } from '../../app/routes/extras'
+import { CATEGORIES_ROUTE, goToCategories } from '../../app/routes/categories'
+import { EVENTS_ROUTE, goToEvents } from '../../app/routes/events'
+import { goToSearch } from '../../app/routes/search'
+import { goToLanding } from '../../app/routes/landing'
+
+class LocationHeader extends React.Component {
+  static propTypes = {
+    languages: PropTypes.arrayOf(PropTypes.instanceOf(LanguageModel)),
```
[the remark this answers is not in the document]
klinzo: I think the header should be visible as early as possible, which means even before the languages have been fetched.

src/modules/app/i18n.js
```diff
@@ -0,0 +1,37 @@
...
+
+i18n
+  .use(LanguageDetector)
+  .init({
+    resources: transformResources(localesResources),
+    fallbackLng: FALLBACK_LANGUAGE,
+    load: 'languageOnly',
+    // eslint-disable-next-line no-undef
+    debug: __DEV__
+  })
```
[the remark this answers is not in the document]
klinzo: this is not working, I would appreciate it if we could discuss and fix this on wednesday

src/routes/extras/containers/ExtrasPage.js
klinzo: error handling is still missing

src/modules/app/containers/Switcher.js
klinzo: I would do this whole error handling topic in WEBAPP-203, this task is already large enough imo

klinzo: but I agree

src/routes/extras/containers/ExtrasPage.js
klinzo: I would do this whole error handling topic in WEBAPP-203, this task is already large enough imo

src/modules/app/containers/I18nProvider.js
klinzo: yes, for me they load sometimes but sometimes not. Could we also discuss/fix this on wednesday?

src/modules/common/containers/LanguageSelector.js
klinzo: nope, removed

src/modules/endpoint/models/__tests__/LocationModel.spec.js
klinzo: tests must  be refactored anyway

src/modules/app/constants.js
klinzo: yes, I would appreciate it

src/modules/app/onBackNext.js
klinzo: removed :D

src/modules/app/routes/categories.js
klinzo: I don't understand this, sorry :D

src/modules/endpoint/fetchers/categories.js
klinzo: imo Endpoint and Endpoint builder made everything way more complex than it had to be, I didn't like it. We could discuss this though and maybe readd them with another task. This one is already big enough.

src/modules/endpoint/fetchers/extras.js
klinzo: Suggestions? There is only one dispatch here and this is saving the extras to the store

src/modules/app/routes/landing.js
klinzo: it also wasn't the landing page before, / was the i18nRedirect

klinzo: the landing was added because it's easier to differ between a city name and a language code, because there are also city names which are only three digits long (we don't have one right now though)

src/modules/endpoint/fetchers/index.js
klinzo: not really

klinzo:
@maxammann So you don't think this is a step forward? The change from redux-little-router to redux-first-router is big, I didn't see a way to reduce the differences without breaking the whole app. I had to refactor all the components in order to keep the app working.

maxammann:
@klinzo hm yes I don't really see a stop forward. We have now the new and better router but at the cost of broken error handling and a very complex fetching and store integration.

Imo this is very complex for just fetching a few json documents. It already was complicated before but the goal should be to reduce it. What do the others think? @Schedulaar @klinzo

klinzo:
@maxammann why is the error handling now broken?

klinzo:
also if the dependency to the redux store is a problem it's not that difficult to remove it

src/modules/app/routes/mainDisclaimer.js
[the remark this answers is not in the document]
klinzo: done

src/modules/common/containers/LanguageSelector.js
```diff
@@ -1,24 +1,31 @@
...
  */
 export class LanguageSelector extends React.Component {
   static propTypes = {
     closeDropDownCallback: PropTypes.func,
-    languages: PropTypes.arrayOf(PropTypes.instanceOf(LanguageModel)).isRequired,
-    language: PropTypes.string.isRequired,
-    location: PropTypes.string.isRequired,
-    languageChangeUrls: PropTypes.object.isRequired,
-    verticalLayout: PropTypes.bool
+    languages: PropTypes.arrayOf(PropTypes.instanceOf(LanguageModel)),
```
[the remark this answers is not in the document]
klinzo: Changed back to required

src/modules/app/routes/mainDisclaimer.js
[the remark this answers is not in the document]
klinzo: it is not really required, because we don't have a link in our app directing to /disclaimer

src/modules/app/routes/categories.js
klinzo: okay now I do, yes of course we could do that

klinzo: do we want that?

src/modules/app/routes/categories.js
klinzo: done

maxammann:
@klinzo you said you'd do the error handling in WEBAPP-203 :P

klinzo:
@maxammann yeah but the error handling isn't broken because of this? :D so right now we have three different types of error: invalid language or city, which is also handled right now. error during the fetching process, which can easily be added (WEBPP-203), and invalid category, event or extra, which should be handled by the corresponding route. The first one (invalid language or city) was nearly impossible to handle with the withFetcher design we had before. So I think error handling is easier than before

klinzo:
@maxammann I also think that this is a step forward regarding routing and stuff, only because of the declaring of the routes in one place and easy linking to different pages. The task maybe does not achieve the goal of making the fetching process fully independent of the redux store, so maybe this PR should be called WEBAPP-174: Switch to redux-first-router and the saving of the data in the redux store should be removed in a different task. We should discuss first whether we really want to remove the fetched data from the store, imo this would not be a real benefit. But a real rest api could be a problem

src/modules/app/containers/Switcher.js
```diff
@@ -29,102 +24,88 @@ import { DISCLAIMER_ROUTE } from '../routes/disclaimer'
...
+        return Switcher.getDisplayComponent(<SearchPage />, categoriesPayload)
       case PDF_FETCHER_ROUTE:
-        if (categories) {
-          return <PdfFetcherPage />
-        }
-        break
+        return Switcher.getDisplayComponent(<PdfFetcherPage />, categoriesPayload)
     }
-    return <LoadingSpinner />
+    return null
```
Schedulaar: We could think about displaying an error here 🤔, I would also prefer to put that into a `default` case

src/modules/app/createReduxStore.js
```diff
@@ -34,7 +34,7 @@ const createReduxStore = (createHistory, initialState = {}, routesMap = defaultR
   if (__DEV__) {
     middlewares.push(createLogger()) // Logs all state changes in console
   }
-
+  console.log(endpointReducers)
```
Schedulaar: remove

src/modules/app/onBeforeChange.js
```diff
@@ -2,53 +2,32 @@
...
 
-    if (!state.languages) {
-      await languagesFetcher(dispatch, params)
-    }
+    await languagesEndpoint.loadData(dispatch, state.languages, params)
 
-    if (!state.events) {
-      await eventsFetcher(dispatch, params)
-    }
+    await eventsEndpoint.loadData(dispatch, state.events, params)
```
Schedulaar: We should load them all in parallel, I think this should work:
```js
  await Promise.all([
    citiesEndpoint.loadData(dispatch, state.cities, params),
    languagesEndpoint.loadData(dispatch, state.languages, params),
    eventsEndpoint.loadData(dispatch, state.events, params)
  ])
```

src/modules/app/routes/categories.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-import categoriesFetcher from '../../endpoint/fetchers/categories'
+import categoriesFetcher from '../../endpoint/endpoints/categories'
```
Schedulaar: categoriesEndpoint

src/modules/app/routes/disclaimer.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-import disclaimerFetcher from '../../endpoint/fetchers/disclaimer'
+import disclaimerFetcher from '../../endpoint/endpoints/disclaimer'
```
Schedulaar: disclaimerEndpoint

src/routes/categories/containers/CategoriesPage.js
```diff
@@ -147,37 +76,29 @@ export class CategoriesPage extends React.Component<Props> {
...
-      const category = this.props.categories.getCategoryByUrl(this.props.path)
-      return <div>
-        <Breadcrumbs
-          parents={this.props.categories.getAncestors(category)}
-          locationName={this.getLocationName(this.props.location)} />
-        {this.getContent(category)}
-        <PdfButton href={this.getPdfFetchPath()} />
-      </div>
-    } catch (e) {
+    const categoryModel = this.props.categories.getCategoryByUrl(this.props.categoryPath)
```
Schedulaar: Maybe we should rename `getCategoryByUrl` to `findCategoryByUrl` since it can return undefined...

src/modules/app/routes/categoriesRedirect.js
```diff
@@ -22,10 +22,14 @@ export const categoriesRedirectRoute = {
     const state = getState()
     const {city, language, categoryId} = state.location.payload
 
-    const categories = await categoriesFetcher(dispatch, {city, language})
+    const categoriesPayload = await categoriesFetcher.loadData(dispatch, state.categories, {city, language})
+
+    if (!categoriesPayload.data) {
+      // todo error handling
```
Schedulaar: todo 🤔

src/modules/app/routes/categoriesRedirect.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-import categoriesFetcher from '../../endpoint/fetchers/categories'
+import categoriesFetcher from '../../endpoint/endpoints/categories'
```
Schedulaar: categoriesEndpoint

src/modules/app/routes/i18nRedirect.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-import citiesFetcher from '../../endpoint/fetchers/cities'
+import citiesFetcher from '../../endpoint/endpoints/cities'
```
Schedulaar: citiesEndpoint

src/modules/app/routes/i18nRedirect.js
```diff
@@ -26,7 +26,11 @@ export const i18nRedirectRoute = {
     const state = getState()
     const param = state.location.payload.param
 
-    const cities = state.cities || await citiesFetcher(dispatch, {})
+    const citiesPayload = await citiesFetcher.loadData(dispatch, state.cities)
+
+    if (!citiesPayload.data) {
+      // todo error handling
```
Schedulaar: todo

src/modules/endpoint/Endpoint.js
```diff
@@ -0,0 +1,146 @@
...
+      return payload
+    }
+
+    const payload = await this.fetchData(formattedURL, params)
+    dispatch(this.finishFetchAction(payload))
+    return payload
+  }
+
+  async fetchData (formattedUrl, params) {
+    return fetch(formattedUrl)
```
Schedulaar: should this function be async?
I think you could have it either non `async`, or have it `async` and do the following refactoring (I hope that works): 
```js
async fetchData (formattedUrl, params) {
  try {
    const response = await fetch(formattedUrl)
    if (!response.ok) {
      throw new Error(endpointLoadingErrorMessage)
    }
    const json = await response.json()
    const fetchedData = this.mapResponse(json, params)
    return Promise.resolve(new Payload(false, fetchedData, null, formattedUrl))
  } catch (e) {
    console.error(`Failed to fetch data for the endpoint: ${this.stateName}`)
    console.error(e)
    return new Payload(false, null, 'endpoint:page.loadingFailed', formattedUrl)
  }

File: src/routes/pdf-fetcher/containers/PdfFetcherPage.js [hunk 39]
```diff
@@ -30,9 +30,13 @@ class PdfFetcherPage extends React.Component {
   }
 
   componentWillMount () {
-    const category = this.props.categories.getCategoryByUrl(this.props.fetchUrl)
-    if (category) {
-      this.fetchUrl(category)
+    try {
+      const category = this.props.categories.getCategoryByUrl(this.props.fetchUrl)
```
Schedulaar: sometimes you have the try block and sometimes you dont with getCategoryByUrl 🤔

src/modules/app/containers/Switcher.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  render () {
+    const {viewportSmall, currentRoute} = this.props
+
+    if (currentRoute === LANDING_ROUTE) {
+      return <Layout footer={<GeneralFooter />}>
+        {this.getPage()}
+      </Layout>
+    } else if ([CATEGORIES_ROUTE, EVENTS_ROUTE, DISCLAIMER_ROUTE, EXTRAS_ROUTE, SEARCH_ROUTE].includes(currentRoute)) {
```
maxammann: I think it would be cool if this would be some kind of configuration like:
```
{
    render: () => <EventsPage />
    layout: LOCATION_LAYOUT
}
```

src/modules/app/containers/Switcher.js
```diff
@@ -0,0 +1,111 @@
...
+    } else if (payload.data) {
+      return page
+    } else if (payload.error) {
+      return <Failure error={payload.error} />
+    } else {
+      return null
+    }
+  }
+
+  getPage (): React.Node {
```
maxammann: this should be called renderPage() because executing this method is only valid inside of an render() function.

src/modules/app/containers/Switcher.js
```diff
@@ -0,0 +1,111 @@
...
+  eventsPayload: Payload,
+  extrasPayload: Payload,
+  disclaimerPayload: Payload
+}
+
+/**
+ * Renders different Pages depending on the current route. If the needed data is not available, a LoadingSpinner is rendered
+ */
+class Switcher extends React.Component<Props> {
+  static getDisplayComponent (page: React.Node, payload: Payload): React.Node {
```
maxammann: this should be called renderDisplayComponent() because executing this method is only valid inside of an render() function.

src/modules/app/containers/Switcher.js
```diff
@@ -0,0 +1,111 @@
...
+    const LoadingSpinner = () => <Spinner name='line-scale-party' />
+    if (payload.isFetching) {
+      return <LoadingSpinner />
```
maxammann: This is a bit problematic. If this case happens we already created `<CategoriesPage />` in getPage(). It would be better if this only gets created its actually used.
I think this is not only a minor optimisatin but also kind of should be done. If you instantiate a component by JSX you expect it to get rendered. This _could_ lead to an un expected behaviour.

src/modules/app/createReduxStore.js
```diff
@@ -1,54 +1,49 @@
...
-import setLanguageChangeUrlsReducer from '../language/reducers/setLanguageChangeUrls'
+import endpointReducers from '../endpoint/reducers'
 import { createResponsiveStateReducer, responsiveStoreEnhancer } from 'redux-responsive'
+import defaultRoutesMap from './routesMap'
 
-const createReduxStore = (createHistory, initialState = {}, routes = new RouteConfig()) => {
+import onBeforeChange from './onBeforeChange'
+
+/**
+ * Creates our Redux store
```
maxammann: I think we can remove this comment. The name createReduxStore is already expressive enough :)

src/modules/app/createReduxStore.js
```diff
@@ -1,54 +1,49 @@
...
     thunkMiddleware // Allows to return functions in actions
   ]
 
   // eslint-disable-next-line no-undef
   if (__DEV__) {
     middlewares.push(createLogger()) // Logs all state changes in console
   }
-
-  const reducer = combineReducers({
+  console.log(endpointReducers)
```
maxammann: _Ignore if WIP: Remove console.log_

src/modules/app/i18n.js
```diff
@@ -0,0 +1,37 @@
+import i18n from 'i18next'
+import LanguageDetector from 'i18next-browser-languagedetector'
+import localesResources from '../../locales'
+import { reduce, forEach } from 'lodash/collection'
+
+const FALLBACK_LANGUAGE = 'en'
+
+/**
+ * i18n is an instance of i18next which is responsible for language detection and change, translations and rtl vs ltr layout
```
maxammann: :warning: Warning, Warning! Comment in the wild!

Is the comment below suggesting to add back the I18n class? I think that would be good.

src/modules/app/onBeforeChange.js
```diff
@@ -0,0 +1,34 @@
...
+ * (onBeforeChange is executed before a change of the route)
+ */
+const onBeforeChange = async (dispatch: Dispatch, getState: GetState, bag: Bag) => {
+  const state = getState()
+  const {city, language} = bag.action.payload
+  const route = bag.action.type
+  const params = {city, language}
+
+  // in the following routes we have a location layout, so we need cities, languages and events
+  if ([CATEGORIES_ROUTE, EVENTS_ROUTE, EXTRAS_ROUTE, DISCLAIMER_ROUTE, SEARCH_ROUTE].includes(route)) {
```
maxammann: This is duplicated in the Switcher component. Can we use the configuration I mentioned?

src/modules/app/routes/categories.js
maxammann: I think this is no relevant anymore.

src/modules/app/routes/categories.js
```diff
@@ -0,0 +1,25 @@
...
+export const CATEGORIES_ROUTE = 'CATEGORIES'
+
+export const goToCategories = (city: string, language: string, categoryPath: ?string) =>
+  createAction(CATEGORIES_ROUTE)({city, language, categoryPath})
+
+/**
+ * CategoriesRoute, matches /augsburg/de*
+ * @type {{path: string, thunk: function(Dispatch, GetState)}}
+ */
+export const categoriesRoute = {
```
maxammann: Do we have a type for a route?

src/modules/app/routes/categoriesRedirect.js
```diff
@@ -0,0 +1,38 @@
...
+export const categoriesRedirectRoute = {
+  path: '/:city/:language/redirect/:categoryId',
+  thunk: async (dispatch: Dispatch, getState: GetState) => {
+    const state = getState()
+    const {city, language, categoryId} = state.location.payload
+
+    const categoriesPayload = await categoriesFetcher.loadData(dispatch, state.categories, {city, language})
+
+    if (!categoriesPayload.data) {
+      // todo error handling
```
maxammann: Does this belong to an other task?

src/modules/app/routesMap.js
```diff
@@ -0,0 +1,26 @@
+import { I18N_REDIRECT_ROUTE, i18nRedirectRoute } from './routes/i18nRedirect'
```
maxammann: We only want to export classes and functions basically. Can we create a simple immutable RouteConfig/RouteMap from this?

src/modules/common/containers/LanguageSelector.js
```diff
@@ -27,33 +36,71 @@ export class LanguageSelector extends React.Component {
...
+
+    switch (routeType) {
+      case CATEGORIES_ROUTE:
+        if (categories) {
+          try {
+            const category = categories.getCategoryByUrl(location.pathname)
+            if (category && category.id !== 0) {
+              return goToCategoriesRedirect(city, languageCode, `${category.availableLanguages[languageCode]}`)
+            }
+          } catch (e) {
```
maxammann: What exception does this catch? Can flow add types here?

maxammann: Apparently not :(, https://github.com/facebook/flow/issues/2470

maxammann: Where is this exception thrown? Exceptions should be used for the handling of unexpected errors. This more sounds like another return value.

src/routes/categories/containers/__tests__/LanguageFailure.spec.js
maxammann: Still relevant

src/modules/common/models/SelectorItemModel.js
```diff
@@ -1,8 +1,8 @@
 export default class SelectorItemModel {
```
maxammann: Not related to this PR, but SelectorItemModel is no model it's just a SelectorItem and should be in the root of the common/ module.

maxammann: Models are domain-specific objects which come from the backend.

src/modules/app/containers/Switcher.js
klinzo: shouldn't happen if routes are correctly added, but makes sense anyway

src/modules/endpoint/Endpoint.js
```diff
@@ -98,71 +84,62 @@ class Endpoint {
...
+      .then(response => {
+        if (response.ok) {
+          return response.json()
+        } else {
+          throw new Error(endpointLoadingErrorMessage)
+        }
+      })
+      .then(json => {
```
maxammann: We should stick to the async-await syntax and don't mix it with .then code.

src/modules/app/onBeforeChange.js
klinzo: good suggestion

src/routes/notFound/NotFoundPage.js
```diff
@@ -0,0 +1,46 @@
...
+  invalidCity: ?string
+}
+
+/**
+ * Our error component, but since the name Error collides with the ES6 class, we've called it Failure
+ */
+export class NotFoundPage extends React.Component<Props> {
+  getCityName (): ?string {
+    const {city, cities} = this.props
+    const cityModel = cities.find(_city => _city.code === city)
```
maxammann: I don't really like the _ code-style. I'd prefere real names here:
`const cityModel = cities.find(city => city.code === lookupCity)`

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
maxammann: I think getCategoryByUrl should now throw an exception.

klinzo: there is always a try catch block, except from tests (which have to be refactored) and the use inside CategoriesMapModel getAncestors(), where there shouldn't be an error since all parents are existing (we checked that in the categories endpoint)

src/routes/categories/containers/CategoriesPage.js
klinzo: nope, it throws, it does not return undefined

src/modules/app/routes/categoriesRedirect.js
klinzo: WEBAPP-203

src/modules/app/routes/i18nRedirect.js
klinzo: WEBAPP-203

src/modules/app/containers/Switcher.js
klinzo: What do you mean with render: () => <EventsPage />?

src/modules/app/containers/Switcher.js
klinzo: kind of contradictory  with the comment above :D

src/modules/app/containers/Switcher.js
klinzo: you're right, refactored

src/modules/app/i18n.js
klinzo: We have to fix the whole i18n thing, I would do this in a separate task

src/modules/app/routes/categoriesRedirect.js
klinzo: yep, I would do this in WEBAPP-203 along with all the other error handling stuff

src/modules/app/containers/Switcher.js
klinzo: I don't understand it :/

src/modules/app/onBeforeChange.js
klinzo: I used an array because I don't understand your configuration :D

src/modules/app/containers/Switcher.js
maxammann: The render function would return the node which needs to be renders for this route.

src/modules/app/containers/Switcher.js
maxammann: Both methods should be renamed.

src/routes/notFound/NotFoundPage.js
klinzo: I won't refactor all of this in this task. Apart from that, I like it :D

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
maxammann: @klinzo ok but then you should not catch them. This is not expected and we're fine if the app crashes if the backend returns wrong data. @Schedulaar @NKlug What are your opinions?

src/modules/app/routes/categories.js
maxammann: Currently the type is documented in a comment. This should be "documented" by using flow.

src/modules/common/models/SelectorItemModel.js
klinzo: okay

src/modules/app/routesMap.js
klinzo: what is the problem with the current?

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
Schedulaar: No the app should not crash: getCategoryByUrl would always return undefined, if the user navigates to a dead/invalid link where we should show an error.

src/modules/app/routesMap.js
maxammann: It's plain-old javascript. We have now cool classes. I think this would be the only file which exports an object (except for the new route definitions)

klinzo:
@maxammann you mean getCategoriesByUrl should not throw an exception? And as I said before, I will refactor tests with a different task, so the LocationModel thing in tests is not relevant right now :D

src/modules/common/containers/LanguageSelector.js
klinzo: in getCategoriesByUrl. We can remove the exception there, but getAncestors will be ugly than because of flow

src/routes/pdf-fetcher/containers/PdfFetcherPage.js
klinzo: we could also remove the exception from getCategoryByUrl, but getAncestors will be ugly than (because of flow)

src/modules/app/i18n.js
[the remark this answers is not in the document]
Schedulaar: did you fix this?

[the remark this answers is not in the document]
Schedulaar: Otherwise you could add it to the i18n task

src/routes/landing/components/LocationSelector.js
Schedulaar: Did you do it? :D

src/modules/app/containers/I18nProvider.js
Schedulaar: Did you fix this?
If not, you could add it to the i18n task

src/routes/landing/components/LocationSelector.js
klinzo: sure :P

klinzo: @Schedulaar I'm ALWAYS applying your suggestions :D

src/modules/app/containers/I18nProvider.js
klinzo: @Schedulaar fixed (by reverting the changes made to the I18nProvider and readding the I18nRedirectPage)

src/routes/categories/containers/__tests__/LanguageFailure.spec.js
klinzo: @maxammann Yep, changed in WEBAPP-234 :D

src/modules/app/constants.js
klinzo: @maxammann did you create a task?

src/modules/app/containers/Switcher.js
klinzo: @maxammann you could add a task

src/modules/app/containers/Switcher.js
klinzo: done

src/modules/app/containers/Switcher.js
klinzo: done

src/modules/app/routes/i18nRedirect.js
```diff
@@ -24,33 +18,7 @@ export const i18nRedirectRoute = {
...
-
-    // the param is probably a language code, so redirect to the landing route
-    if (param.length >= MIN_LANGUAGE_CODE_LENGTH || param.length <= MAX_LANGUAGE_CODE_LENGTH) {
-      dispatch(redirect(goToLanding(param)))
-      return
-    }
-
-    // param is neither a language code nor a city, so it is not a valid route
-    dispatch(redirect(goToNotFound(param)))
+    await citiesEndpoint.loadData(dispatch, state.cities, {})
```
Schedulaar: We don't need that any more right?

Schedulaar: We can leave it there for now.

klinzo: yes we need it, otherwise we would not have the cities in the I18nRedirectPage to check whether the param is a valid city

PR #124, opened 2018-03-21
WEBAPP-219: Styled Components and lib

Schedulaar:


src/modules/common/components/Caption.js
```diff
@@ -1,7 +1,17 @@
...
+
+const H1 = styled.h1`
+  margin: 25px 0;
+  font-size: 2rem;
+  text-align: center;
+  
+  @media ${props => props.theme.dimensions.smallViewport} {
+    margin: 10px 0;
+  }
+`
```
Schedulaar: Move these

src/modules/app/constants/helpers.js
```diff
@@ -0,0 +1,6 @@
+export default {
```
klinzo: could we rename this to something more declarative? for example removeLinkMarking..., because it is only one helper

PR #125, opened 2018-03-23
WEBAPP-232: Consider EndDate when filtering events

Schedulaar:

### DIFF
--- a/src/modules/endpoint/endpoints/events.js
+++ b/src/modules/endpoint/endpoints/events.js
@@ -10,21 +10,24 @@ export default new EndpointBuilder('events')
     `/${state.router.params.language}/wp-json/extensions/v0/modified_content/events?since=1970-01-01T00:00:00Z`)
   .withMapper(json => json
     .filter(event => event.status === 'publish')
-    .map(event => (new EventModel({
-      id: event.id,
-      title: event.title,
-      content: event.content,
-      thumbnail: event.thumbnail,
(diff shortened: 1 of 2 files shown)


src/modules/endpoint/endpoints/events.js
```diff
@@ -10,21 +10,24 @@ export default new EndpointBuilder('events')
...
+        startDate: moment(`${event.event.start_date} ${allDay ? '00:00:00' : event.event.start_time}`),
+        endDate: moment(`${event.event.end_date} ${allDay ? '23:59:59' : event.event.end_time}`),
+        allDay: allDay,
+        excerpt: event.excerpt,
+        availableLanguages: event.available_languages
+      })
+    })
     .filter(event => event.startDate.isValid())
-    .filter(event => event.startDate.isAfter(moment()))
+    .filter(event => event.startDate.isAfter(moment()) || (event.endDate.isValid() && event.endDate.isAfter(moment())))
```
klinzo: Do we need to filter for event.startDate.isAfter(moment()) and event.endDate.isAfter(moment())? I think filtering for event.endDate should be enough (+ startDate.isValid()), since the endDate should always be after startDate.

Schedulaar: If the endDate is not valid (e.g. because the backend serves an empty string for it), this implementation is still safe...

PR #126, opened 2018-03-23
WEBAPP-84: improve pdf generation

Schedulaar:

main.js is now ~half the size 🥇


src/routes/categories/containers/CategoriesPage.js
```diff
@@ -98,8 +99,12 @@ export class CategoriesPage extends React.Component<Props> {
     }
   }
 
-  getPdfFetchPath () {
-    return `/${this.props.location}/${this.props.language}/fetch-pdf?url=${this.props.path}`
+  getPdfUrl (category: CategoryModel) {
```
maxammann: Haha one function vs one bloated half megabyte class :P

Whooop!!

PR #127, opened 2018-03-25
WEBAPP-184 Upgrade npm packages

Schedulaar:

~We can't update webpack to v4 yet, because stylelint-webpack-plugin still lacking v4 support...~


klinzo:
Is the bable polyfill package beta stable enough?

maxammann:
If it works without the beta I would suggest to downgrade. For example we can not easily say why our builds fail right now. It could be a reason for it.

Also stuff like https://github.com/npm/npm/issues/19883 happens....

klinzo:
the bable polyfill was stable before this task so it can't be the reason for build failures

maxammann:
@klinzo yeah it was beta before this task and our builds failed before this task. I don't think this is the reason. Just a general thing to stick away from betas. Especially in the node world :P

klinzo:
nope it was in version 6.26.0 which isn't a beta, is it? Yet some other babel plugins and packages where also a beta before. Totally agreed, this is why I asked.

Schedulaar:
The travis bug was indeed induced by upgrading babel, but not because of babel, but because of me, since I forgot to update the `polyfills.js` with the updated import statement for babel/polyfill.
I get your point that we shouldn't use beta software, but in that case we already agreed on upgrading. Also it made life much easier when introducing styled-components so I wouldn't suggest the effort of downgrading it again, since babel7 has promised to release in the near future.

package.json
```diff
@@ -7,128 +7,128 @@
...
     "redux-first-router-link": "^1.4.2",
     "redux-logger": "~3.0.6",
-    "redux-responsive": "^4.3.5",
+    "redux-responsive": "^4.3.6",
     "redux-thunk": "~2.2.0",
-    "styled-components": "^3.2.0",
+    "styled-components": "^3.2.5",
     "url-pattern": "^1.0.3",
-    "url-search-params-polyfill": "~2.0.1",
+    "url-search-params-polyfill": "~3.0.0",
```
maxammann: Do we still use the search params? I think I only used it for the weird pdf endpoint.

tools/webpack.config.js
```diff
@@ -167,14 +168,6 @@ const config = {
 
 // Optimize the bundle in release (production) mode
 if (!isDebug) {
-  config.plugins.push(new webpack.optimize.UglifyJsPlugin({
```
maxammann: Is this applied automatically? If yes with what parameters?

package.json
```diff
@@ -7,128 +7,128 @@
...
-    "webpack": "~3.10.0",
-    "webpack-bundle-analyzer": "~2.9.1",
-    "webpack-dev-middleware": "~1.12.2",
-    "webpack-hot-middleware": "~2.21.0",
-    "webpack-node-externals": "^1.6.0"
+    "stylelint-order": "~0.8.1",
+    "stylelint-processor-styled-components": "^1.3.1",
+    "stylelint-webpack-plugin": "~0.10.4",
+    "url-loader": "~1.0.1",
+    "webpack": "~4.5.0",
```
maxammann: Are there any webpack plugins which are not updated yet? Like "assets-webpack-plugin"

tools/webpack.config.js
Schedulaar: Yes, that comes with the new `mode: production` in webpack 4... But the params are some defaults set by webpack... We could disable the default minifying of webpack4 though and put an uglifyjsPlugin in ourselves... What do you think?

package.json
Schedulaar: Woops, it seems assets-webpack-plugin didn't upgrade yet... Also it's last commit was Jan 20, 2017.
What do we need that plugin for ? 🙈

Schedulaar: @maxammann I suggest to merge it anyways now....

PR #131, opened 2018-03-28
WEBAPP-234: Fix all tests

klinzo:


src/modules/app/containers/I18nProvider.js
```diff
@@ -1,11 +1,14 @@
...
 import { I18nextProvider } from 'react-i18next'
+import { reduce, forEach } from 'lodash/collection'
 import WebFont from 'webfontloader'
 import PropTypes from 'prop-types'
+import LanguageDetector from 'i18next-browser-languagedetector'
 
-import { LANGUAGE_CODE_LENGTH } from '../constants'
-import i18n from '../i18n'
+import localesResources from 'locales.json'
+import { MAX_LANGUAGE_CODE_LENGTH, MIN_LANGUAGE_CODE_LENGTH } from '../constants'
```
Schedulaar: Do we still need them since have no ambigious routes any more?

src/modules/app/containers/I18nProvider.js
```diff
@@ -21,15 +24,39 @@ export class I18nProvider extends React.Component {
...
+      })
+      return accumulator
+    }, {})
+  }
+
   setLanguage (language) {
     const targetLanguage = language || this.i18n.languages[0]
 
-    if (targetLanguage.length !== LANGUAGE_CODE_LENGTH) {
+    if (targetLanguage.length < MIN_LANGUAGE_CODE_LENGTH || targetLanguage.length > MAX_LANGUAGE_CODE_LENGTH) {
```
Schedulaar: Do we stil lneed to check that? see above

src/modules/app/routes/categoriesRedirect.js
```diff
@@ -28,11 +29,13 @@ export const categoriesRedirectRoute = {
       // todo error handling
     }
 
-    try {
-      const category = categoriesPayload.data.getCategoryById(Number(categoryId))
-      dispatch(redirect(goToCategories(city, language, category.path)))
-    } catch (e) {
-      dispatch(redirect(goToNotFound(city, language)))
+    if (categoriesPayload.data instanceof CategoriesMapModel) {
```
Schedulaar: Why do you need to check that its of the right "type"?
And what should happen if it wasn't the right type?

src/modules/app/routes/extras.js
```diff
@@ -22,14 +23,16 @@ export const extrasRoute = {
...
-    if (!extrasPayload) {
+    if (!extrasPayload.data) {
       // todo error handling
     }
 
-    if (extraAlias === 'sprungbrett') {
-      const sprungbrettModel = extrasPayload.data.find(_extra => _extra.alias === extraAlias)
-      if (sprungbrettModel) {
-        await sprungbrettEndpoint.loadData(dispatch, state.sprungbrettJobs, {url: sprungbrettModel.path})
+    if (Array.isArray(extrasPayload.data)) {
```
Schedulaar: same as above

src/modules/app/routes/categoriesRedirect.js
klinzo: It is required by flow, otherwise we get an error here because we can't be sure that the payload data is a CategoriesMapModel. Will do this in WEBAPP-203, along with most of the other error handling stuff.

src/modules/app/containers/I18nProvider.js
klinzo: I think so, yes. Otherwise we would redirect to the landing page if you enter '/berlin' or '/some_random_stuff', because we have to decide what the meaning of this single param is (language, city, disclaimer, not_found...)

src/modules/app/routes/categoriesRedirect.js
klinzo: actually I don't think we need any error handling here at all. The data of the payload must not be something different than the corresponding model (or an array of it) and imo we don't have to handle the case that it is not the right data anyway, we should just not fetch or redirect than.

klinzo: on the other hand if the data is null, the payload has an error, which will be handled in the Switcher component (or a FailureComponent rendered in Switcher). Will do this in WEBAPP-203

src/modules/app/containers/I18nProvider.js
Schedulaar: Well... I think we should always redirect from '/' to '/landing/{i18nLanguage}'
and from '/asdf' to '/asdf/{i18nLanguage}' ?
So no need to decide if it's a language because it'll always be a city...

klinzo: I will remove this redirect, as discussed

src/routes/i18nRedirect/containers/I18nRedirect.js
```diff
@@ -0,0 +1,81 @@
...
+      return goToLanding(i18n.language)
+    }
+
+    if (param === 'disclaimer') {
+      return goToMainDisclaimer()
+    }
+
+    if (param === 'not_found') {
+      return goToNotFound()
+    }
```
Schedulaar: ⚠️ We shouldn't need to care about disclaimer or not_found here. I18nRedirect should only be rendered, when it's needed.

I think the main issue is, that the router still maps '/disclaimer' to the i18nRoute, and not to the mainDisclaimerRoute... This should not happen. I hope we can configure this in the router itself.

klinzo: yep you're right I also stumbled about this right now. As it was with ReduxLittleRouter, the order of the routes is important (in the RoutesMap). So if you put the disclaimer route before i18nRedirect, the disclaimer route is used when entering disclaimer

klinzo: should work again now, fixed the issue in WEBAPP-223

PR #132, opened 2018-03-28
WEBAPP-234: Readd tests

klinzo:

This PR reverts the last 4 of the WEBAPP-234: Fix tests PR.

### DIFF
--- a/src/modules/app/routes/__tests__/mainDisclaimer.spec.js
+++ b/src/modules/app/routes/__tests__/mainDisclaimer.spec.js
@@ -1,0 +1,13 @@
+import { goToMainDisclaimer, MAIN_DISCLAIMER_ROUTE, mainDisclaimerRoute } from '../mainDisclaimer'
+
+describe('mainDisclaimer route', () => {
+  it('should create the right action', () => {
+    expect(goToMainDisclaimer()).toEqual({
+      type: MAIN_DISCLAIMER_ROUTE
+    })
+  })
+
+  it('should have the right path', () => {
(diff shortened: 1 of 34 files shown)


src/modules/app/routes/__tests__/mainDisclaimer.spec.js
```diff
@@ -0,0 +1,13 @@
+import { goToMainDisclaimer, MAIN_DISCLAIMER_ROUTE, mainDisclaimerRoute } from '../mainDisclaimer'
+
+describe('mainDisclaimer route', () => {
```
maxammann: As a convention we usually write a name of a class or function here. What is a mainDisclaimer right now?

klinzo: Our Integreat Disclaimer, reachable with '/disclaimer'

klinzo:
@maxammann The testing of the thunks of the routes is missing, I'll add a JIRA task for this. I'm not quite sure how to test this though. Apart from that, the endpoints are not thoroughly tested (some branches missing), but this was also not tested before. All other newly added files should have tests.

src/modules/app/routes/__tests__/mainDisclaimer.spec.js
klinzo: The naming here is because we have some endpoint names which conflict with route names, i.e. extras, categories, events.... So we would have two tests with the same name testing different stuff

klinzo:
I did not know we had an caption in extras. Yep probably destroyed by a merge. I'll add it with my next task.

PR #134, opened 2018-04-07
WEBAPP-241: Disable position sticky for edge16

Schedulaar:


src/routes/landing/components/LocationEntry.js
```diff
@@ -1,9 +1,8 @@
 import PropTypes from 'prop-types'
```
klinzo: if you want you could rename this to CityEntry (also LocationSelector -> CitySelector, LocationListItem -> CityListItem and so on)

klinzo:
@Schedulaar and you could apply flow type checking :P

Schedulaar:
@klinzo won't add flow type checking, out of scope of this task

src/modules/common/constants.js
```diff
@@ -0,0 +1,4 @@
+import { detect } from 'detect-browser'
+const browser = detect()
+
+export const positionStickyDisabled = browser && browser.name === 'edge' && /^16\..*/.test(browser.version)
```
maxammann: Imporing this file can have unintendet side effects because detect() is called.

I would propose to have a Platform class which calls detect() in the constructor and has a function isPositionStickyDisabled

Schedulaar: @maxammann Hmm... should we have the platform class instance then somewhere in the state?
Otherwise we would create a platform instance in each stylesheet, and therefore call detect() many times, which could impact performance...

maxammann: Then it would make sense to add a PlatformProvider as react context. State doesn't really make sense beause its static data (also class instances shouldn't be in the state. Only obejcts).

Schedulaar: Done

PR #135, opened 2018-04-07
WEBAPP-104: Breadcrumbs scrolled to the end

Schedulaar:


src/routes/categories/containers/CategoriesPage.js
```diff
@@ -80,23 +83,33 @@ export class CategoriesPage extends React.Component<Props> {
                          content={category.content} />
   }
 
+  getBreadcrumbs (categoryModel: CategoryModel): Array<Node> {
```
klinzo: :+1:

src/modules/app/actions/setUIDirection.js
```diff
@@ -0,0 +1,10 @@
+// @flow
```
klinzo: Why did you put the uiDirection in the redux state? You could also access the context in the CategoriesPage and use 
```js
RTL_LANGUAGES.includes(targetLanguage) ? 'rtl' : 'ltr'
```
there

klinzo:
You could also add flow type checking to some more files

src/modules/app/actions/setUIDirection.js
Schedulaar: Hmm.. I think its up to our i18nProvider to decide which language is rtl 🤔

klinzo: Maybe a static method would do the job?

maxammann: I think its ok to have it like it is. But a i18n module would be nice maybe.

src/modules/common/components/Breadcrumbs.styles.js
```diff
@@ -0,0 +1,40 @@
+import styled from 'styled-components'
+
+const opposite = direction => direction === 'ltr' ? 'rtl' : 'ltr'
+export const Wrapper = styled.div`
+  margin: 10px 0;
+  text-align: end;
+  white-space: nowrap;
+  overflow-x: auto;
+  direction: ${props => opposite(props.direction)};
```
maxammann: Do we really need the `direction` here? Can we maybe use flex to reverse the order?

src/modules/common/components/__tests__/Breadcrumb.spec.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
```
maxammann: This file should be named Breadcrumb_s_.spec.js

src/modules/common/components/Breadcrumbs.styles.js
Schedulaar: No, as far as i know, this specific functionality only works with the `direction` property (so that the scrollStart is at the opposite direction)

src/modules/common/components/__tests__/Breadcrumb.spec.js
Schedulaar: Done

src/modules/app/actions/setUIDirection.js
Schedulaar: Hmm... yeah that's right... I'lll move I18nProvider along to the i18n package

Schedulaar: Done.

src/modules/app/actions/setUIDirection.js
```diff
@@ -0,0 +1,10 @@
...
+
+import { createAction } from 'redux-actions'
+import type { Action } from 'redux-first-router/dist/flow-types'
+
+export const setUiDirectionAction = 'SET_UI_DIRECTION'
+
+const setUiDirection = (direction: 'ltr' | 'rtl'): Action => createAction(setUiDirectionAction)(direction)
```
maxammann: Should be the same name as the file. so setUIDirection.

PR #136, opened 2018-04-08
WEBAPP-240: Fix language change and show message if there are no other available languages

klinzo:

This fixes the language change if a category is not available in all languages. If there are no other available languages than the current one, grey the languageIcon out and show a tooltip message that there are no other languages.


src/modules/layout/components/__tests__/__snapshots__/GeneralHeader.spec.js.snap
```diff
@@ -5,14 +5,14 @@ exports[`GeneralHeader should match snapshot 1`] = `
...
       HeaderActionItem {
-        "_dropDownNode": undefined,
         "_href": Object {
           "payload": Object {
             "param": undefined,
           },
           "type": "I18N_REDIRECT",
         },
         "_iconSrc": "test-file-stub",
+        "_node": undefined,
```
maxammann: Is here mock data missing?

PR #137, opened 2018-04-13
WEBAPP-230: Dynamic page titles

klinzo:

Make our page titles dynamic with react-helmet
Rename ExtraModel name to title
Rename I18nRedirect to I18nRedirectPage


src/locales.json
```diff
@@ -164,34 +164,35 @@
...
-      "creatingPdf": null,
-      "downloadPdfAt": null,
-      "page": {
-        "loadingFailed": null
-      }
+      "extras": "اضافات",
+      "news": "تازه​ها",
+      "landing": "تو کجایی؟",
+      "search": "جستجو",
+      "disclaimer": "مجوز ‎و اطلاعات تماس"
```
Schedulaar: Hmm... It's not clear from the name of these constants where they are used and what 'functionality' they have... Maybe you could rename them to sth like "disclaimerTitle" or similar...

src/modules/app/containers/Helmet.js
```diff
@@ -0,0 +1,112 @@
...
+          return t('extras')
+        }
+        break
+      case SEARCH_ROUTE:
+        return t('search')
+      case MAIN_DISCLAIMER_ROUTE:
+        return t('disclaimer')
+      case LANDING_ROUTE:
+        return t('landing')
+    }
```
Schedulaar: I don't like that we have another huge "switcher" here... Imo, the routes should decide on their title themselves.
Maybe we can expand the objects in /modules/app/routes/ for that purpose?
If we add an appropriate flow-type there, we should also be safe, that we haven't forgot to set a title to some route.
Also the tests wouldn't bloat a single file, like here Helmet.spec.js, but instead be spread to all of the routes...

src/routes/extras/containers/ExtrasPage.js
```diff
@@ -37,7 +37,7 @@ export class ExtrasPage extends React.Component<Props> {
   getTileModels (): Array<TileModel> {
     return this.props.extras.map(extra => new TileModel({
       id: extra.alias,
-      name: extra.name,
+      name: extra.title,
```
Schedulaar: then you could also rename `name` to `title` in the TileModel, right?

klinzo: Sure :D

PR #139, opened 2018-04-13
WEBAPP-220: Use v3 of locations endpoint

klinzo:


src/modules/endpoint/endpoints/cities.js
```diff
@@ -17,15 +17,16 @@ const stripSlashes = (path: string): string => {
...
         name: _city.name,
         code: stripSlashes(_city.path),
         live: _city.live,
-        eventsEnabled: _city['ige-evts'] === '1',
-        extrasEnabled: true // todo
+        eventsEnabled: _city.events,
+        extrasEnabled: _city.extras
       }))
       .sort(_city => _city.name)
     return sortBy(cities, _city => _city.sortKey)
```
Schedulaar: Sorting should be different now - we get a new key `name_without_prefix` from the backend. That should be used for sorting now.

klinzo: yep, right.

src/modules/layout/containers/__tests__/__snapshots__/LocationLayout.spec.js.snap
```diff
@@ -11,8 +11,8 @@ exports[`LocationLayout should show CategoriesToolbar if current route is catego
   }
   header={
     <Connect(Translate(LocationHeader))
-      isEventsEnabled={false}
-      isExtrasEnabled={false}
+      isEventsEnabled={undefined}
```
Schedulaar: I think these should not be undefined

klinzo: kind of, but we don't need them for LocationLayout... I find it way more clear to read the tests if we don't specify all atributes of models even if we don't need them. Although we have undefined in our snaps then

klinzo: kind of, but we don't need them for LocationLayout... I find it way more clear to read the tests if we don't specify all atributes of models even if we don't need them. Although we have undefined in our snaps then

PR #140, opened 2018-04-14
WEBAPP-77: Add label for inputs

klinzo:

### DIFF
--- a/src/locales.json
+++ b/src/locales.json
@@ -42,7 +42,7 @@
     },
     "en": {
       "pageTitle": "Search",
-      "searchCategory": "Search a category"
+      "searchCategory": "Search for a category"
     },
     "fa": {
       "pageTitle": "جستجو"
@@ -124,7 +124,7 @@
     "en": {
       "where": "Where are you?",
       "pageTitle": "Where are you?",
-      "searchCity": "Search your city"
+      "searchCity": "Search for your city"
     },
     "fa": {
       "where": "تو کجایی؟",
--- a/src/locales.json
+++ b/src/locales.json
@@ -48,20 +48,6 @@
       "pageTitle": "جستجو"
     }
   },
-  "common": {
-    "ar": {
-      "search": "البحث"
-    },
-    "de": {
-      "search": "Suchen"
-    },
-    "en": {
-      "search": "Search"
-    },
-    "fa": {
-      "search": "جستجو"
-    }
-  },
   "categories": {
     "ar": {
     },
--- a/src/locales.json
+++ b/src/locales.json
@@ -37,10 +37,12 @@
       "pageTitle": "البحث"
     },
     "de": {
-      "pageTitle": "Suche"
+      "pageTitle": "Suche",
+      "searchCategory": "Suche nach einer Kategorie"
     },
     "en": {
-      "pageTitle": "Search"
+      "pageTitle": "Search",
+      "searchCategory": "Search a category"
     },
     "fa": {
       "pageTitle": "جستجو"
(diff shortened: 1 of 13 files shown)


src/locales.json
```diff
@@ -37,29 +37,17 @@
...
     },
     "de": {
-      "pageTitle": "Suche"
+      "pageTitle": "Suche",
+      "searchCategory": "Suche nach einer Kategorie"
     },
     "en": {
-      "pageTitle": "Search"
+      "pageTitle": "Search",
+      "searchCategory": "Search a category"
```
NKlug: Search _for_ a category?

src/locales.json
```diff
@@ -130,11 +118,13 @@
...
       "where": "Wo bist du?",
-      "pageTitle": "Wo bist du?"
+      "pageTitle": "Wo bist du?",
+      "searchCity": "Suche nach deinem Ort"
     },
     "en": {
       "where": "Where are you?",
-      "pageTitle": "Where are you?"
+      "pageTitle": "Where are you?",
+      "searchCity": "Search your city"
```
NKlug: I think it should be 'Search _for_ your city'.

src/locales.json
klinzo: yep you're right

src/modules/common/components/SearchInput.js
```diff
@@ -1,33 +1,34 @@
...
   render () {
+    const {onClickInput, filterText, placeholderText} = this.props
     return (
       <Spacer space={this.props.spaceSearch}>
         <Wrapper>
           <SearchIcon />
-          <TextInput placeholder={this.props.t('search')}
-                     defaultValue={this.props.filterText}
+          <TextInput placeholder={placeholderText}
+                     label={placeholderText}
```
maxammann: Is label really used? https://github.com/Integreat/integreat-webapp/blob/f67040565993824bbad672d52e49f9cc5b67358a/src/modules/common/components/SearchInput.styles.js#L10

Also a `<input>` does not have a label attribute: https://www.w3schools.com/tags/tag_input.asp

But there is an aria-label: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute

klinzo: what do you mean by "really used"? Changed to aria-label.

src/modules/common/components/__tests__/__snapshots__/ScrollingSearchBox.spec.js.snap
```diff
@@ -34,7 +35,8 @@ exports[`ScrollingSearchBox should set correct reference 1`] = `
         />
         <input
           class="SearchInputstyles__TextInput-kJgKBC lfoCjL"
-          placeholder="common:search"
+          label="Placeholder"
```
Schedulaar: wrong snapshot, I think you just need to update them.

src/modules/common/components/__tests__/__snapshots__/SearchInput.spec.js.snap
```diff
@@ -1,11 +1,18 @@
...
-  onFilterTextChange={[Function]}
-  t={[Function]}
-/>
+<SearchInputstyles__Spacer>
+  <SearchInputstyles__Wrapper>
+    <SearchInputstyles__SearchIcon />
+    <SearchInputstyles__TextInput
+      autoFocus={true}
+      defaultValue="Test"
+      label="Placeholder"
```
Schedulaar: Here as well...

src/modules/common/components/__tests__/__snapshots__/ScrollingSearchBox.spec.js.snap
klinzo: I was sure I did update them o.O

PR #142, opened 2018-04-18
WEBAPP-259: Make 3rd level categories look more subordinate

klinzo:

Thx @Schedulaar :+1: 
Indent 3rd level categories.


Schedulaar:
Hmm... looks odd to me, when there are no thumbnails for the 3rd level available (which is our official recommendation)
http://localhost:3000/deutschlandtemplate2018/en/education-work-und-university-studies

klinzo:
@Schedulaar 
image

better like this?

klinzo:
I don't think it is necessary asking #design-marketing for this, since the changes of the design are quite small and it's only affecting the design on desktop.

maxammann:
I like it that there are no icons. Icons are also distracting and very difficult to get right.

Like how would you distiguish between an icon for 'Mobilität' vs. 'Öffentliche Verkehrsmittel'.

src/routes/categories/components/CategoryListItem.js
```diff
@@ -8,7 +8,7 @@ import {
   CategoryCaption,
   CategoryThumbnail, Row,
   StyledLink, SubCategoryCaption, SubCategoryThumbnail,
-  SubCategoryThumbnailDiv
+  SubCategoryThumbnailDiv, SubCategory
```
maxammann: Not part of this PR, but: What is the difference between `SubCategoryThumbnailDiv` and `SubCategoryThumbnail`?

klinzo:
So you would just don't show any icons for 3rd level categories even if there are some?

maxammann:
@klinzo I think that's what looks odd to @Schedulaar. imo its fine if there are none.

We should ask in the Slack if we tell people to add icons for sub categories.

Schedulaar:
@maxammann No, I think prior to this task, it looked fine when there were no icons. Since 'no icons' is our official recommendation, we shouldn't care (too much) if it doesn't look good if there are icons.
With this task the indentation of the yellow lines don't align any more which looks (for me) somewhat strange, and we don't follow the design of the native app as well...
Plus it looks odd, when cities mix icons / no icons: http://localhost:3000/augsburg/de/lebensorganisation

klinzo:
imo it also didn't look very good if there were no icons before this task (because the yellow line started before the text which looks strange for me), plus imo there was way to less difference between second and third level categories, as it was criticised by some cities and clara.

maxammann:
As we're already disussing here we should move this to Slack I think :)

PR #143, opened 2018-04-18
WEBAPP-76: Update manifest.json

klinzo:


klinzo:
@schedulaar could you maybe provide a source why we need a apple-touch-icon? I think with the icons this is no longer required

klinzo:
@Schedulaar done

PR #145, opened 2018-04-20
WEBAPP-215: Timestamps for pages

klinzo:

Add timestamps to categories without children (pages), when the category was edited the last time.


src/routes/categories/components/CategoryTimeStamp.styles.js
```diff
@@ -0,0 +1,8 @@
+import styled from 'styled-components'
+
+export const TimeStamp = styled.div`
+  font-family: ${props => props.theme.fonts.contentFontFamily};
+  font-size: ${props => props.theme.fonts.contentFontSize};
+  opacity: 0.5;
+  padding-top: 10px;
+`
```
Schedulaar: make it a `p`-tag, remove the padding-top and use textSecondaryColor instead of opacity

klinzo: I'll still leave the padding-top, imo the space in between should be bigger.

src/modules/endpoint/models/CategoryModel.js
```diff
@@ -69,6 +73,14 @@ class CategoryModel {
...
     return this._availableLanguages
   }
+
+  getLastUpdate (locale: string): Moment {
+    const lastUpdate = this._lastUpdate
+    lastUpdate.locale(locale)
+
+    // only return day, month and year
+    return lastUpdate.format('LL')
+  }
```
Schedulaar: Um actually... :D I think this toString conversion should be made within the TimeStamp component. So we give the timeStamp a moment-object as prop

src/routes/categories/containers/__tests__/CategoriesPage.spec.js
```diff
@@ -86,7 +90,7 @@ describe('CategoriesPage', () => {
     expect(wrapper).toMatchSnapshot()
   })
 
-  it('should match snapshot render a CategoryList if the category is neither the root nor has children', () => {
+  it('should match snapshot render a CategoryList if the category is neither the root but has children', () => {
```
Schedulaar: This description still is no sentence :D

src/locales.json
```diff
@@ -50,14 +50,18 @@
   },
   "categories": {
     "ar": {
```
maxammann: Should we add a placeholder if we see that some translations are missing? e.g. `createPdf`

src/routes/categories/components/__tests__/__snapshots__/CategoryTimeStamp.spec.js.snap
```diff
@@ -0,0 +1,8 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`CategoryTimeStamp should match snapshot 1`] = `
+<CategoryTimeStampstyles__TimeStamp>
+  lastUpdate
+  18. November 2017
```
maxammann: Isn't this platform specific? I think its possible that this test fails on other platforms with an other locale.

klinzo: What do you mean by platform specific? The locale is the current language, so in the test it should always be the same.

src/locales.json
klinzo: I think this is not necessary since the default is used than which is imo better (Create PDF instead of a possible placeholder createPdf)

src/routes/categories/components/__tests__/__snapshots__/CategoryTimeStamp.spec.js.snap
maxammann: But the current language which moment uses is different on different platforms. To fix this I suggest we set the locale of moment in the setupTest.js.

I think the test would fail now if your system language would be arabic.

src/locales.json
maxammann: Maybe we can copy the english text. Yes it falls back but it makes it very difficult to fix missing translations if we don't mark them in some way.

An external list would work too but maybe is difficult to keep up to date.

We can also convert the language files to .js files and include comments.

src/routes/categories/components/__tests__/__snapshots__/CategoryTimeStamp.spec.js.snap
klinzo: No, it doesn't. The language/locale to which the timestamp is formatted is the language prop of CategoryTimeStamp, which is set in the test to 'de'. There is no way this could be different on different platforms.

src/locales.json
Schedulaar: I actually made a script once, that outputs all missing translations to the console... I'll add it to the project :D

PR #148, opened 2018-04-25
WEBAPP-262: Use api v3 categories endpoint

klinzo:


src/modules/common/containers/LanguageSelector.js
```diff
@@ -33,19 +32,19 @@ export class LanguageSelector extends React.Component {
    */
   getLanguageChangeAction (languageCode) {
```
Schedulaar: This doesn't return an action anymore, right?

src/modules/endpoint/endpoints/__tests__/categories.spec.js
```diff
@@ -8,104 +8,78 @@ jest.unmock('../categories')
...
     id: 3649,
-    permalink: {
-      url_page: 'willkommen'
-    },
+    path: '/augsburg/de/willkommen',
     title: 'Willkommen',
     status: 'publish',
     content: '',
-    parent: 0,
+    parent: {},
```
Schedulaar: why do we need an object here? 🤔

src/modules/endpoint/models/CategoryModel.js
```diff
@@ -4,25 +4,23 @@ import type { Moment } from 'moment'
...
-    this._url = params.url
+    this._path = params.path
     this._title = params.title
     this._content = params.content
-    this._parentId = params.parentId
-    this._parentUrl = params.parentUrl
+    this._parentPath = params.parentPath
     this._thumbnail = params.thumbnail
     this._order = params.order
     this._path = params.path
```
Schedulaar: that's a duplicate

src/modules/endpoint/models/CategoryModel.js
```diff
@@ -4,25 +4,23 @@ import type { Moment } from 'moment'
...
-  _parentUrl: string
+  _parentPath: string
   _thumbnail: string
   _order: number
   _availableLanguages: Map<string, string>
   _lastUpdate: Moment
 
-  constructor (params: {| id: number, path: string, url: string, title: string, content: string, parentId: number, thumbnail: string,
-    parentUrl: string, order: number, availableLanguages: Map<string, string>, lastUpdate: Moment |}) {
+  constructor (params: {| id: number, path: string, path: string, title: string, content: string, thumbnail: string,
```
Schedulaar: `path` is duplicated

src/routes/categories/components/__tests__/CategoryListItem.spec.js
```diff
@@ -40,7 +40,7 @@ describe('CategoryListItem', () => {
   })
 
   it('should replace empty thumbnail', () => {
-    const noThumbCategory = {id: 42, url: '', title: 'GotNoThumb :O', parentId: 2}
+    const noThumbCategory = {id: 42, path: '', title: 'GotNoThumb :O', parentId: 2}
```
Schedulaar: We don't need a parentId anymore

src/routes/categories/containers/__tests__/CategoriesPage.spec.js
```diff
@@ -12,21 +12,21 @@ describe('CategoriesPage', () => {
...
       thumbnail: 'no_thumbnail',
-      parentUrl: ''
+      parentPath: ''
     }), new CategoryModel({
       id: 3650,
-      url: '/augsburg/de/anlaufstellen',
+      path: '/augsburg/de/anlaufstellen',
       title: 'Anlaufstellen zu sonstigen Themen',
       content: '',
       parentId: 0,
```
Schedulaar: Maybe also adjust the test-set here as well, so we don't have parentId

src/routes/categories/containers/__tests__/__snapshots__/CategoriesToolbar.spec.js.snap
```diff
@@ -10,7 +10,7 @@ exports[`CategoriesToolbar should render Toolbar, if category can be found 1`] =
 </Toolbar>
 `;
 
-exports[`CategoriesToolbar should render root-url for pdf endpoint 1`] = `
+exports[`CategoriesToolbar should render root-path for pdf endpoint 1`] = `
```
Schedulaar: Here its a URL though :D

src/modules/common/containers/LanguageSelector.js
klinzo: yes it still does (with exception of categories pathname), but I will adjust this (and change the name) to always return a  path with WEBAPP-231.

src/modules/endpoint/endpoints/__tests__/categories.spec.js
klinzo: This is the json returned by the api v3, which returns the id, the full cms url and the path for each parent, so parent is an object.

src/modules/endpoint/models/CategoryModel.js
klinzo: ups :D

src/routes/categories/containers/__tests__/__snapshots__/CategoriesToolbar.spec.js.snap
klinzo: Changed

src/modules/endpoint/endpoints/__tests__/categories.spec.js
Schedulaar: Ah I see... But maybe you could still fill them up to real parent objects like we get them from cms (for categories with no parent we get`{id: 0,  path: null, url: null }`)

src/modules/endpoint/models/CategoryModel.js
klinzo: Removed

src/modules/endpoint/endpoints/__tests__/categories.spec.js
klinzo: okay, sure :)

PR #149, opened 2018-04-26
WEBAPP-276: Find missing locales - script

Schedulaar:

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,6 +144,7 @@
     "test": "jest --config jest.config.json",
     "test:coverage": "jest --config jest.config.json --coverage",
     "test:watch": "jest --config jest.config.json --watchAll",
-    "test:update": "jest --config jest.config.json -u"
+    "test:update": "jest --config jest.config.json -u",
+    "findMissingLocales": "node tools/findMissingLocales"
   }
 }
--- a/tools/findMissingLocales.js
+++ b/tools/findMissingLocales.js
@@ -1,0 +1,35 @@
+const fs = require('fs')
+const task = require('./task')
+
+const {reduce, map, union, isEmpty} = require('lodash')
+
+function findMissingLocales () {
+  const locales = JSON.parse(fs.readFileSync('./src/locales.json', 'utf8'))
+  const keys = reduce(locales, (moduleAcc, languages, moduleKey) => {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -144,6 +144,7 @@
     "test": "jest --config jest.config.json",
     "test:coverage": "jest --config jest.config.json --coverage",
     "test:watch": "jest --config jest.config.json --watchAll",
-    "test:update": "jest --config jest.config.json -u"
+    "test:update": "jest --config jest.config.json -u",
+    "findMissingLocales": "node tools/findMissingLocales"
```
maxammann: `locales:missing`? :)

tools/findMissingLocales.js
```diff
@@ -0,0 +1,35 @@
+const fs = require('fs')
+const task = require('./task')
+
+const {reduce, map, union, isEmpty} = require('lodash')
+
+function findMissingLocales () {
```
maxammann: I think `findMissingKeys()` would be a better name.

tools/findMissingLocales.js
```diff
@@ -0,0 +1,35 @@
...
+    }
+    return moduleAcc
+  }, {})
+
+  return missingKeys
+}
+
+module.exports = task('findMissingLocales', () => {
+  return Promise.resolve(findMissingLocales())
+    .then(locales => console.log(locales))
```
maxammann: As `locales` are keys here, can we also return in which language the keys are missing?

tools/findMissingLocales.js
```diff
@@ -0,0 +1,35 @@
+const fs = require('fs')
```
maxammann: I think locales.js would be a better name.

tools/findMissingLocales.js
Schedulaar: The `locales` here have the following structure:
```js
{
  module1: {
    lang1: ['key1', 'key2'],
    lang2: ['key2', 'key3']
  },
  ...
}
```

tools/findMissingLocales.js
Schedulaar: Hmm.. I don't think so... that doesn't tell you what the file actually does...

PR #150, opened 2018-04-27
WEBAPP-231: Add hreflang links

klinzo:


src/modules/common/containers/LanguageSelector.js
```diff
@@ -50,30 +52,32 @@ export class LanguageSelector extends React.Component {
...
     return languages && languages
       .map(language =>
         new SelectorItemModel({
-          code: language.code, name: language.name, href: this.getLanguageChangeAction(language.code)
+          code: language.code,
+          name: language.name,
+          href: LanguageSelector.getLanguageChangePath({categories, events, location, languageCode: language.code})
         })
       )
       .filter(selectorItem => selectorItem.href)
```
Schedulaar: I think we shouldn't filter these. I would expect to always see all languages, that the city offers, in the languageSelector. Once a city hasn't translated one category, then it wouldn't show up in the selector...
I'd prefer to go to a fallback url, like the categoriesHome of that language then (or at least disable it in the selector - like we do in the header when there are no events available)...
Maybe the latter suggestion would be more appropriate (and easier to implement :) )

src/modules/common/containers/LanguageSelector.js
```diff
@@ -1,37 +1,39 @@
...
-  }
+export class LanguageSelector extends React.Component<Props> {
   /**
    * Maps the given languageCode to an action to go to the current route in the language specified by languageCode
-   * @param languageCode
-   * @return {*}
    */
-  getLanguageChangeAction (languageCode) {
-    const {location, categories, events} = this.props
+  static getLanguageChangePath (params: {| location: Location, categories: CategoriesMapModel,
```
Schedulaar: I'm not happy, that this method is in the LanguageSelector. This should be moved somewhere else...

src/modules/common/containers/LanguageSelector.js
klinzo: Alright, I don't like the first suggestion at all, but the second one sounds reasonable and makes sense for me, will do it like that :+1:

src/modules/common/containers/LanguageSelector.js
klinzo: @Schedulaar do you have a suggestion where to put it? Create a separate file called getLanguageChangePath?

src/modules/common/containers/LanguageSelector.js
klinzo: On the other hand, it does not make sense to add links to sites in the header that don't exist, so filtering makes total sense for the hreflang links. The LanguageSelector is a different thing though.

src/modules/common/containers/LanguageSelector.js
Schedulaar: Hmm... I'd put it somewhere near the definition of our routes, so probably in the /app module 🤔

src/modules/common/containers/LanguageSelector.js
klinzo: @Schedulaar done

src/modules/common/components/Selector.js
```diff
@@ -1,37 +1,52 @@
...
-          <Element key={item.code}
-                onClick={this.props.closeDropDownCallback}
-                to={item.href}>
-            {item.name}
-          </Element>
+          <span data-tip={t(tooltip)} key={item.code}>
+            <InactiveElement>
+              {item.name}
+            </InactiveElement>
+            <Tooltip />
```
Schedulaar: We shouldn't spawn multiple ReactTooltip "instances" (this can lead to unexpected z-index behavior - i've had this problem 😅 ). Instead we should call `ReactTooltip.rebuild()` in `componentDidRender()` and - if needed - in `componentDidUpdate()`.

src/modules/common/components/Selector.js
```diff
@@ -1,37 +1,52 @@
...
         )
       } else {
         return (
-          <Element key={item.code}
-                onClick={this.props.closeDropDownCallback}
-                to={item.href}>
-            {item.name}
-          </Element>
+          <span data-tip={t(tooltip)} key={item.code}>
+            <InactiveElement>
```
Schedulaar: Why not `<InactiveElement data-tip={t(tooltip)} key={item.code}>` instead of these two lines ?

src/modules/common/components/Selector.js
```diff
@@ -1,37 +1,52 @@
...
+import { translate } from 'react-i18next'
+import type { I18nTranslate } from '../../../flowTypes'
+import Tooltip from './Tooltip'
+
+type Props = {
+  verticalLayout: boolean,
+  closeDropDownCallback?: () => {},
+  items: Array<SelectorItemModel>,
+  activeItemCode?: string,
+  tooltip: string,
```
Schedulaar: Rather name it `disabledItemsTooltip` or similar

src/modules/common/components/Selector.js
```diff
@@ -46,4 +61,4 @@ class Selector extends React.Component {
   }
 }
 
-export default Selector
+export default translate('common')(Selector)
```
Schedulaar: I think we shouldn't translate this component. `Selector` should be universally usable and not only with our I18n-library... Instead we should translate the tooltip one component above 🤔 I think it always makes sense to translate the translation-keys in that place where we spawn them :D

src/modules/layout/components/HeaderLanguageSelectorItem.js
```diff
@@ -23,7 +23,8 @@ class HeaderLanguageSelectorItem extends React.Component<Props> {
       ? <HeaderDropDown iconSrc={languageIcon}>
         <Selector verticalLayout={false}
                   items={selectorItems}
-                  activeItemCode={activeItemCode} />
+                  activeItemCode={activeItemCode}
+                  tooltip={'noTranslation'} />
```
Schedulaar: I'd suggest to translate it here already...

src/modules/layout/components/HeaderLanguageSelectorItem.styles.js
```diff
@@ -1,9 +1,9 @@
 import styled from 'styled-components'
-import ReactTooltip from 'react-tooltip'
+import UnstyledTooltip from '../../common/components/Tooltip'
 
-export const Tooltip = styled(ReactTooltip)`
-  display: inline-block
+export const Tooltip = styled(UnstyledTooltip)`
+  display: inline-block;
```
Schedulaar: why do we need this inline-block display ? :D

src/modules/common/components/Tooltip.js
```diff
@@ -0,0 +1,12 @@
...
+import React from 'react'
+import ReactTooltip from 'react-tooltip'
+
+class Tooltip extends React.Component<{}> {
+  render () {
+    return <ReactTooltip effect='solid' delayShow={0} />
+  }
+}
+
+export default Tooltip
```
Schedulaar: I don't think that we need this file since we should only have one `ReactTooltip` in our renderTree

src/modules/common/components/Selector.js
Schedulaar: (especially since the Selector is in `lib.js`)

src/modules/common/components/Selector.styles.js
```diff
@@ -14,12 +14,6 @@ export const Element = styled(Link)`
   border-radius: 30px;
   transition: background-color 0.2s, border-radius 0.2s;
   user-select: none;
-
```
Schedulaar: You can remove the export and the `color` above, if you decide to specify the color for each item itself and export them separately

src/modules/common/components/Selector.styles.js
```diff
@@ -30,7 +24,19 @@ export const Element = styled(Link)`
...
+export const ActiveElement = Element.extend`
+  :hover {
+    font-weight: 700;
+    border-radius: 0;
+  }
+  color: ${props => props.theme.colors.textColor};
+`
+
+export const InactiveElement = Element.withComponent('span').extend`
+  color: ${props => props.theme.colors.textSecondaryColor};
```
Schedulaar: Maybe we can even go with `textDecorativeColor` but I'm not sure... For me `textSecondaryColor` is still too dark and not easy to distinguish with `textColor`... Maybe we need a `textDisabledColor` :D

src/modules/common/components/Selector.js
klinzo: Ahh okay, only componentDidRender did not work, componentDidUpdate solved it :+1: thanks

src/modules/common/components/Selector.js
klinzo: See comment above.

src/modules/common/components/Selector.styles.js
klinzo: I agree, could you create a new task?

src/modules/common/components/Selector.js
```diff
@@ -1,37 +1,50 @@
...
-    /** The code of the item which is currently active **/
-    activeItemCode: PropTypes.string
+class Selector extends React.Component<Props> {
+  componentDidMount () {
+    /* https://www.npmjs.com/package/react-tooltip#1-using-tooltip-within-the-modal-eg-react-modal- */
+    ReactTooltip.rebuild()
+  }
+
+  componentDidUpdate () {
+    /* https://www.npmjs.com/package/react-tooltip#1-using-tooltip-within-the-modal-eg-react-modal- */
```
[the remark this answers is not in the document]
klinzo: yeah, the tooltip is in the Layout.js file, https://github.com/Integreat/integreat-webapp/blob/e3a07a60598a716c389548db3a3c07d78e175b20/src/modules/layout/components/Layout.js#L42

PR #151, opened 2018-04-29
WEBAPP-277: Remove opacity

klinzo:


src/modules/layout/components/HeaderNavigationItem.js
```diff
@@ -1,46 +1,44 @@
...
-    </React.Fragment>
+    if (active) {
+      return <ActiveNavigationItem key={text} to={href} selected={selected}>
+        {text}
+      </ActiveNavigationItem>
+    } else {
+      return <InactiveNavigationItem key={text} data-tip={tooltip}>
+        {text}
+      </InactiveNavigationItem>
+    }
```
Schedulaar: tests are missing for this logic right?

klinzo: Yes, I'll add a test for this. Was already missing before though. :D

PR #154, opened 2018-05-11
WIP: WEBAPP-289: Add vscode configs

maxammann:

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,14 @@
+{
+    "recommendations": [
+        "dbaeumer.vscode-eslint",
+        "mgmcdermott.vscode-language-babel",
+        "flowtype.flow-for-vscode",
+        "eamodio.gitlens",
+        "yzhang.markdown-all-in-one",
+        "IBM.output-colorizer",
+        "gamunu.vscode-yarn",
+        "christian-kohler.npm-intellisense",
+        "eg2.vscode-npm-script",
+        "msjsdiag.debugger-for-chrome"
+    ]
+}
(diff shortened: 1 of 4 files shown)


.vscode/extensions.json
```diff
@@ -0,0 +1,14 @@
+{
```
Schedulaar: So will you switch to VSCode in the future? I think it'd be cool, if we all have the same IDE^^

maxammann: I'll switch most likely. It just feels much better on the first try :)

Requirement for me right now: Dont't break anything with intellij. Formatting could be a problem.

But I found a plugin for vscode which formats the files according to the eslint rules, which should solve all the problems.

.vscode/launch.json
```diff
@@ -0,0 +1,24 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
Schedulaar: comments in json ? :D Oh micrsoft :shaking_my_head:

src/modules/layout/containers/LocationHeader.js
```diff
@@ -10,16 +10,16 @@ import landingIcon from '../assets/location-icon.svg'
...
+import {CATEGORIES_ROUTE, goToCategories} from '../../app/routes/categories'
+import {EVENTS_ROUTE, goToEvents} from '../../app/routes/events'
+import {goToSearch} from '../../app/routes/search'
+import {goToLanding} from '../../app/routes/landing'
+import {connect} from 'react-redux'
+
+import type {LocationState} from 'redux-first-router/dist/flow-types'
 import EventModel from '../../endpoint/models/EventModel'
-import type { I18nTranslate, State } from '../../../flowTypes'
+import type {I18nTranslate, State} from '../../../flowTypes'
```
Schedulaar: Why are all the spaces removed here?

PR #156, opened 2018-05-27
WEBAPP-259: Remove 3rd level category icons

klinzo:


src/routes/categories/components/CategoryListItem.js
```diff
@@ -22,23 +21,22 @@ type Props = {
  * Displays a single CategoryListItem
  */
 class CategoryListItem extends React.Component<Props> {
-  getChildren () {
-    return this.props.children.map(child =>
-      <div key={child.id}>
-        <StyledLink to={child.path}>
-          {
-            child.thumbnail
```
maxammann: :+1: for less trinary

PR #159, opened 2018-05-27
WEBAPP-266: Use api v3 in disclaimer endpoint

klinzo:


src/modules/endpoint/endpoints/disclaimer.js
```diff
@@ -6,6 +6,7 @@ import { isEmpty } from 'lodash/lang'
 import EndpointBuilder from '../EndpointBuilder'
 import ParamMissingError from '../errors/ParamMissingError'
 import type { EndpointParams } from '../../../flowTypes'
+import moment from 'moment/moment'
```
Schedulaar: Do we really import `moment/moment`? Does it differ from `moment`?

PR #161, opened 2018-05-30
WEBAPP-261: Add meta description

maxammann:


src/modules/endpoint/models/CategoryModel.js
```diff
@@ -62,6 +63,14 @@ class CategoryModel {
   get lastUpdate (): Moment {
     return this._lastUpdate
   }
+
+  isRoot (): boolean {
```
klinzo: :+1:

src/routes/landing/containers/LandingPage.js
```diff
@@ -20,9 +20,10 @@ export class LandingPage extends React.Component<Props> {
   render () {
     const {t, language, cities} = this.props
     return <React.Fragment>
-      <Helmet>
+      <ReactHelmet>
         <title>{t('pageTitle')}</title>
-      </Helmet>
+        <meta name='description' content={t('metaDescription')} />
```
klinzo: why do you use a different meta description here?

klinzo: if you did this because you want to customize the meta description some time for cities, you could still define a default meta description in Helmet and override it by placing a new meta description as child. Although, if you want to have a different meta description for the landing page, you should probably also use the same description (or at least one that differs from the description for the cities) for the main disclaimer.

src/modules/common/containers/Helmet.js
```diff
@@ -16,7 +16,8 @@ type Props = {
   categories: CategoriesMapModel,
   events: Array<EventModel>,
   languages: Array<LanguageModel>,
-  location: Location
+  location: Location,
+  metaDescription: string
```
Schedulaar: make metaDescription optional

PR #165, opened 2018-06-20
WEBAPP-202: Remove the scrolling to the search input on LandingPage

Thomas125:


src/modules/common/components/ScrollingSearchBox.js
```diff
@@ -52,8 +42,6 @@ export class ScrollingSearchBox extends React.PureComponent {
 
   onSearchInputTextChange = value => this.onFilterTextChange(value)
```
Schedulaar: duplicate right here :D
You could even remove both of these functions and just pass the function from the props.

maxammann:
I think this can be merged, right?

PR #166, opened 2018-06-27
WEBAPP-174: Adjust linting

maxammann:


klinzo:
Relates to issue 274 right? https://integreat.atlassian.net/browse/WEBAPP-274

maxammann:
@klinzo oops yep, will rewrite the branch history

src/flowTypes.js
```diff
@@ -9,34 +9,34 @@ import EventModel from './modules/endpoint/models/EventModel'
...
-  +uiDirection: UiDirection
+  +uiDirection: UiDirectionType
 }
 
-export type EndpointParams = {city?: string, language?: string, url?: string}
-export type MapParamsToUrl = (params: EndpointParams) => string
-export type MapResponse = (json: any, params: EndpointParams) => PayloadData
+export type EndpointParamsType = {city?: string, language?: string, url?: string}
+export type MapParamsToUrlType = (params: EndpointParamsType) => string
+export type MapResponseType = (json: any, params: EndpointParamsType) => PayloadDataType
```
Schedulaar: When should these have suffix `Type`? Only these explicitly defined as `type`? Normal classes (like all our react components) are also valid flow `type`s afaik, and we obviously don't append `Type` to those...
At least it's a bit odd to me, that we need to append this suffix...

src/modules/app/containers/App.js
```diff
@@ -10,24 +12,30 @@ import PlatformProvider from '../../platform/containers/PlatformProvider'
 import routesMap from '../routesMap'
 import Switcher from './Switcher'
 import theme from '../constants/theme'
+import { Store } from 'redux'
+
+export interface PropsType {
```
Schedulaar: Do we need empty PropTypes now?
Should we always export them?

src/modules/i18n/containers/I18nProvider.js
```diff
@@ -86,22 +113,38 @@ export class I18nProvider extends React.Component {
...
+          style={{
+            direction: RTL_LANGUAGES.includes(this.state.language)
+              ? 'rtl'
+              : 'ltr'
+          }}
+        >
+          <ReactHelmet>
+            {lateef && (
+              <link href='/fonts/lateef/lateef.css' rel='stylesheet' />
+            )}
```
Schedulaar: Do we force the parantheses `(` and `)` around jsx components? I'd suggest not to do that.

src/routes/events/containers/EventsPage.js
```diff
@@ -51,15 +51,15 @@ export class EventsPage extends React.Component<Props> {
...
-  city: state.location.payload.city,
-  eventId: state.location.payload.eventId,
-  events: state.events.data,
-  cities: state.cities.data
+const mapStateTypeToProps = (stateType: StateType) => ({
+  language: stateType.location.payload.language,
+  city: stateType.location.payload.city,
+  eventId: stateType.location.payload.eventId,
+  events: stateType.events.data,
+  cities: stateType.cities.data
```
Schedulaar: `stateType` looks odd, as its not a type. Should be `state`.

src/routes/landing/components/FilterableCitySelector.js
```diff
@@ -8,34 +9,46 @@ import CityModel from 'modules/endpoint/models/CityModel'
...
+
+type StateType = {
+  filterText: string
+}
 
-  constructor (props) {
+export class FilterableCitySelector extends React.Component<
+  PropsType,
+  StateType
+> {
```
Schedulaar: maybe we need to adjust our maximum line width, so we don't need too much of these line breaks.

src/routes/search/containers/SearchPage.js
```diff
@@ -10,22 +9,22 @@ import CategoriesMapModel from 'modules/endpoint/models/CategoriesMapModel'
...
+type PropsType = {
   categories: CategoriesMapModel,
   cities: Array<CityModel>,
   city: string,
-  t: I18nTranslate
+  t: I18nTranslateType
 }
 
-type LocalState = {
+type LocalStateTypeType = {
```
Schedulaar: LocalStateTypeType? :)

maxammann: Oops, will fix, thanks :)

src/routes/landing/components/FilterableCitySelector.js
maxammann: Yes, that looks odd. If we change a eslint rule I won't do it in this task.

But I'll open an issue for this.

src/routes/events/containers/EventsPage.js
maxammann: Yep, thank :+1:

src/modules/i18n/containers/I18nProvider.js
maxammann: I think we do not force them.

src/modules/app/containers/App.js
maxammann: I think we should not export them. Not sure about the empty ones.

src/flowTypes.js
maxammann: That is a linting rule in neuburg-backend. Open for discussion I suppose.

src/modules/i18n/containers/I18nProvider.js
klinzo: There are quite some in the changes above though.

src/modules/layout/components/HeaderDropDown.js
```diff
@@ -1,23 +1,32 @@
...
  * Header. Once the user clicks outside, the node is hidden again. Additionally, the inner node gets a
  * closeDropDownCallback through its props to close the dropDown and hide itself.
  */
-class HeaderDropDown extends React.Component {
-  static propTypes = {
-    iconSrc: PropTypes.string.isRequired,
-    children: PropTypes.element.isRequired
-  }
+class HeaderDropDown extends React.Component<PropsType, StateType> {
+  toggleDropDown: Function
```
klinzo: What type of function?

maxammann: All of these are marked as warnings in flow. Can be fixed later. There are a lot of similar problems

PR #167, opened 2018-07-11
WEBAPP-299: Update deployment docu

maxammann:


tools/deploy/ssh-deploy.sh
```diff
@@ -14,4 +15,4 @@ then
     rsync -vr --delete $(mktemp -d)/ ${HOST}:${REMOTE_DIR}
 fi
 
-rsync -vIacz --progress --stats --exclude=*.ejs www/* ${HOST}:${REMOTE_DIR}
```
Schedulaar: why don't we exclude the .ejs file?

maxammann: We could extend the deploy script, but I think this is more a non-critical problem of our build process (www contains data which should not be deployed).

I want to keep all deploy scritps the same across all repos to simplify stuff.

PR #168, opened 2018-07-18
WEBAPP-307: Upgrade packages

maxammann:
* Fixed flow types once and for all
* Fix dependency problems
* Types can be added by using: yarn flow:install-types

### DIFF
--- a/src/modules/app/createReduxStore.js
+++ b/src/modules/app/createReduxStore.js
@@ -1,4 +1,4 @@
-import { applyMiddleware, combineReducers, compose, createStore, Store } from 'redux'
+import { applyMiddleware, combineReducers, compose, createStore } from 'redux'
 import thunkMiddleware from 'redux-thunk'
 import { connectRoutes } from 'redux-first-router'
 import { createLogger } from 'redux-logger'
@@ -9,7 +9,7 @@ import { createResponsiveStateReducer, responsiveStoreEnhancer } from 'redux-res
 import defaultRoutesMap from './routesMap'
 import onBeforeChange from './onBeforeChange'
 
-const createReduxStore = (createHistory, initialState = {}, routesMap = defaultRoutesMap): Store => {
+const createReduxStore = (createHistory, initialState = {}, routesMap = defaultRoutesMap) => {
   const history = createHistory()
 
   const {reducer, middleware, enhancer} = connectRoutes(history, routesMap, {onBeforeChange: onBeforeChange})


Schedulaar:
Do we really need these flow files in our repo as these are auto generated anyway?

maxammann:
@Schedulaar the docu says to check it in

maxammann:
@Schedulaar I think the reason is that all team members use the same types. As there is no lock file or some other concept.

package.json
```diff
@@ -6,150 +6,153 @@
...
+    "@integreat-app/eslint-config-jest": "^0.2.1",
+    "@integreat-app/eslint-config-react": "^0.2.1",
+    "assets-webpack-plugin": "^3.8.4",
+    "autoprefixer": "^9.0.0",
     "babel-core": "^7.0.0-bridge.0",
-    "babel-eslint": "~8.2.3",
-    "babel-jest": "^22.4.3",
-    "babel-loader": "~8.0.0-beta.2",
+    "babel-eslint": "^8.2.6",
+    "babel-jest": "22.4.4",
```
Schedulaar: Why are you missing the caret ^ here?

package.json
```diff
@@ -6,150 +6,153 @@
...
-    "react-hot-loader": "~4.0.1",
-    "react-scroll": "~1.7.8",
-    "redux-mock-store": "~1.5.1",
-    "regenerator-runtime": "^0.11.1",
-    "rimraf": "~2.6.2",
-    "s3": "~4.4.0",
-    "style-loader": "~0.20.3",
-    "stylefmt": "~6.0.0",
-    "stylelint": "~9.2.0",
+    "jest": "22.4.4",
```
Schedulaar: same here

package.json
```diff
@@ -6,150 +6,153 @@
...
+    "raf": "^3.4.0",
+    "react-hot-loader": "^4.3.3",
+    "react-scroll": "^1.7.10",
+    "redux-mock-store": "^1.5.3",
+    "regenerator-runtime": "^0.12.0",
+    "rimraf": "^2.6.2",
+    "s3": "^4.4.0",
+    "style-loader": "^0.21.0",
+    "stylefmt": "^6.0.2",
+    "stylelint": "8.4.0",
```
Schedulaar: same here

package.json
```diff
@@ -6,150 +6,153 @@
...
-    "react-dom": "~16.3.1",
-    "react-fontawesome": "~1.6.1",
+    "font-awesome": "^4.7.0",
+    "history": "^4.7.2",
+    "i18next": "^11.3.6",
+    "i18next-browser-languagedetector": "^2.2.2",
+    "lodash": "^4.17.10",
+    "moment": "^2.22.2",
+    "moment-timezone": "^0.5.21",
+    "normalize.css": "^8.0.0",
```
Schedulaar: Don't we need normalize-url anymore ? (:tada: :pray: )

src/modules/app/containers/App.js
```diff
@@ -12,15 +12,15 @@ import PlatformProvider from '../../platform/containers/PlatformProvider'
...
 import Switcher from './Switcher'
 import theme from '../constants/theme'
-import { Store } from 'redux'
 
 type PropsType = {
 }
 
 class App extends React.Component<PropsType> {
-  store: Store;
+  store: any;
```
Schedulaar: why did you need to remove the type here?

src/modules/common/components/__tests__/__snapshots__/Headroom.spec.js.snap
```diff
@@ -14,7 +14,7 @@ exports[`Headroom should render correct if state is pinned, transition 1`] = `
     translateY={0}
   >
     <div
-      className="Headroomstyles__HeaderWrapper-iDCIAq fsTJoL"
+      className="Headroomstyles__HeaderWrapper-jYoiGy kNRfRu"
```
Schedulaar: Why are these classnames messed up ?

src/modules/i18n/containers/I18nProvider.js
```diff
@@ -89,13 +90,13 @@ export class I18nProvider extends React.Component<PropsType, StateType> {
     this.i18n.changeLanguage(targetLanguage)
   }
 
-  componentWillMount () {
+  componentDidMount () {
```
Schedulaar: Hmm... I think we shouldn't do this change to the new react version in this task. Maybe we should create a new task for that - These changes could potentially have unwanted side effects...

src/modules/common/components/__tests__/__snapshots__/Headroom.spec.js.snap
maxammann: My guess is that they changed the hashing algorithm :P

src/modules/i18n/containers/I18nProvider.js
maxammann: I tested the changes and there should be no problems. Can change it back to the UNSAFE_ methods though.
I only had to change the lifecycle methods in the I18nProvider. So if there are any problems we know its because of this change.

src/modules/app/containers/App.js
maxammann: Out of scope of this task
Store has 3 generics. There is now a warning because of the any.

package.json
maxammann: Nope. Found normalize-path because newer versions of normalize-url actually work only with urls :+1:

package.json
maxammann: Because we need version 22 because of an incompatibility with the IDE

package.json
maxammann: Because we need version 22 because of an incompatibility with the webpack plugin

package.json
maxammann: Because we need version 22 because of an incompatibility with the IDE

src/modules/i18n/containers/I18nProvider.js
maxammann: Can change it back but I don't see an issue with doing it on DidMount.
Actually we are not allowed to do it in WillMount. WillMount does not allow side-effects.

package.json
Schedulaar: Can we add a comment somewhere so we don't forget to upgrade it when the issue is resolved? (With link to the GitHub issue)

Schedulaar: Also the caret only corresponds to minor package upgrades like 22.4.x. So this wouldn't upgrade jest to v23 or similar :thinking:

Schedulaar: Do you want to do it? I don't know what exactly went wrong 🙈

maxammann: done

src/modules/app/routes/disclaimer.js
```diff
@@ -3,11 +3,11 @@
 import disclaimerEndpoint from '../../endpoint/endpoints/disclaimer'
 import { createAction } from 'redux-actions'
 
-import type { Action, Dispatch, GetState } from 'redux-first-router/dist/flow-types'
+import type { Dispatch, GetState } from 'redux-first-router'
 
 export const DISCLAIMER_ROUTE = 'DISCLAIMER'
 
-export const goToDisclaimer = (city: string, language: string): Action => createAction(DISCLAIMER_ROUTE)({city, language})
+export const goToDisclaimer = (city: string, language: string) => createAction(DISCLAIMER_ROUTE)({city, language})
```
klinzo: Why did you remove the Action return type?

src/modules/endpoint/models/CategoryModel.js
```diff
@@ -13,10 +13,10 @@ class CategoryModel {
   _thumbnail: string
   _order: number
   _availableLanguages: Map<string, string>
-  _lastUpdate: Moment
+  _lastUpdate: ?Moment
```
klinzo: Why is this optional?

src/routes/categories/containers/CategoriesPage.js
```diff
@@ -58,7 +58,7 @@ export class CategoriesPage extends React.Component<PropsType> {
       return <React.Fragment>
         <Page title={category.title}
               content={category.content} />
-        <CategoryTimeStamp lastUpdate={category.lastUpdate} language={language} />
+        {category.lastUpdate && <CategoryTimeStamp lastUpdate={category.lastUpdate} language={language} /> }
```
klinzo: Space after but not before.

Schedulaar:
Has this been merged?

src/modules/app/routes/disclaimer.js
maxammann: Not sure what the type here is. Not needed by the linting rules.

src/modules/endpoint/models/CategoryModel.js
maxammann: it actually was already Moment or null. Just fixed flow issues.

maxammann:
Yes, why?

Schedulaar:
Because the pr has closed manually 🙈

maxammann:
Yeah github did strange stuff with the rebases :P

Schedulaar:
Maybe we should continue to merge instead of rebasing :P

PR #171, opened 2018-07-22
WEBAPP-311: Add simple rating functionality

klinzo:

I've added two buttons two the categories toolbar (happy smiley and sad smiley), which are posting a feedback if clicked.
Also I moved some files in the layout module to a separate header/footer/toolbar package.


klinzo:
Most of the changes are just because I moved some files in the layout module to a separate header/footer/toolbar package.

src/routes/categories/containers/CategoriesToolbar.js
```diff
@@ -3,12 +3,13 @@
...
 
-import Toolbar from '../../../modules/layout/components/Toolbar'
+import Toolbar from '../../../modules/layout/components/toolbar/Toolbar'
 import CategoriesMapModel from '../../../modules/endpoint/models/CategoriesMapModel'
-import ToolbarItem from '../../../modules/layout/components/ToolbarItem'
+import ToolbarAnchor from '../../../modules/layout/components/toolbar/ToolbarAnchor'
 import { apiUrl } from '../../../modules/endpoint/constants'
 import CategoryModel from '../../../modules/endpoint/models/CategoryModel'
 import type { TFunction } from 'react-i18next'
+import RatingToolbarItem from '../../../modules/layout/components/toolbar/ToolbarRatingItem'
```
Schedulaar: Wrong naming: RatingToolbarItem vs ToolbarRatingItem

Schedulaar:
Wrong branch prefix: WEBAPP-311 instead of 311 :)

src/routes/categories/containers/CategoriesToolbar.js
klinzo: Done

src/modules/app/containers/Switcher.js
```diff
@@ -31,8 +31,8 @@ import CityModel from '../../endpoint/models/CityModel'
 import LoadingSpinner from '../../common/components/LoadingSpinner'
 import Layout from '../../layout/components/Layout'
 import LocationLayout, { LocationLayoutRoutes } from '../../layout/containers/LocationLayout'
-import GeneralHeader from '../../layout/components/GeneralHeader'
-import GeneralFooter from '../../layout/components/GeneralFooter'
+import GeneralHeader from '../../layout/components/header/GeneralHeader'
```
maxammann: I think we should keep our flat structure. Its more scalable not to have a deeply nested sturcture. Also having up to 50 files in a folder should be no problem. Looks a bit much, almost all projects work with a structure like this.

src/modules/endpoint/FeedbackEndpointBuilder.js
```diff
@@ -0,0 +1,64 @@
+// @flow
```
maxammann: I think we do not need a Builder for for one specific endpoint which is used only once.

src/modules/endpoint/FeedbackEndpointBuilder.js
```diff
@@ -0,0 +1,64 @@
...
+
+import type { MapParamsToUrlType } from './MapParamsToUrlType'
+import type { MapParamsToBodyType } from './MapParamsToBodyType'
+import FeedbackEndpoint from './FeedbackEndpoint'
+
+/**
+ * Helper class to build a {@link FeedbackEndpoint}
+ */
+class EndpointBuilder<Params> {
```
maxammann: I would go with `<P>` and documentation.

src/modules/endpoint/endpoints/feedback/categoriesFeedback.js
```diff
@@ -0,0 +1,42 @@
...
+  .withParamsToUrlMapper((params): string => {
+    if (!params.city) {
+      throw new ParamMissingError(CATEGORIES_FEEDBACK_ENDPOINT_NAME, 'city')
+    }
+    if (!params.language) {
+      throw new ParamMissingError(CATEGORIES_FEEDBACK_ENDPOINT_NAME, 'language')
+    }
+    return `${apiUrl}/${params.city}/${params.language}/wp-json/extensions/v3/feedback`
+  })
+  .withParamsToBodyMapper((params): FormData => {
```
maxammann: I think we should introduce a proper type here which has a method to convert to a FormData. This would make the code easier to debug. Not sure what the internals of a FormData are.

src/modules/app/containers/Switcher.js
maxammann: You can make a footer and a header module. Which depend on the layout module.

src/modules/endpoint/endpoints/feedback/categoriesFeedback.js
maxammann: I think `withParamsToBodyMapper` should be called `withParamsToFormDataMapper`

src/modules/endpoint/FeedbackEndpointBuilder.js
maxammann: The main reason is that you do not mix generics with actual classes.

src/modules/endpoint/endpoints/feedback/categoriesFeedback.js
```diff
@@ -0,0 +1,42 @@
...
+import { apiUrl } from '../../constants'
+import ParamMissingError from '../../errors/ParamMissingError'
+import FeedbackEndpointBuilder from '../../FeedbackEndpointBuilder'
+
+const CATEGORIES_FEEDBACK_ENDPOINT_NAME = 'categoriesFeedback'
+const POSITIVE_RATING_STRING = 'up'
+const NEGATIVE_RATING_STRING = 'down'
+
+export default new FeedbackEndpointBuilder(CATEGORIES_FEEDBACK_ENDPOINT_NAME)
+  .withParamsToUrlMapper((params): string => {
```
maxammann: params should have types imo. If you do not add types flow interpretes that params has the type:

`{ city: string, language: string }`

In the second mapper blow more types like `isPositiveRating` get added. This makes it not obvious which type params has.

src/modules/endpoint/endpoints/feedback/categoriesFeedback.js
```diff
@@ -0,0 +1,42 @@
...
+      throw new ParamMissingError(CATEGORIES_FEEDBACK_ENDPOINT_NAME, 'language')
+    }
+    return `${apiUrl}/${params.city}/${params.language}/wp-json/extensions/v3/feedback`
+  })
+  .withParamsToBodyMapper((params): FormData => {
+    if (params.id === undefined) {
+      throw new ParamMissingError(CATEGORIES_FEEDBACK_ENDPOINT_NAME, 'page id')
+    }
+
+    if (params.isPositiveRating === undefined && !params.comment) {
```
maxammann: Using undefined in a type is a bit unsecure. `isPositiveRating` should be a boolean so it only can be true or false.
In this code `isPositiveRating` has the type `boolean|void`

Same goes for id etc.

PR #176, opened 2018-07-26
WEBAPP-208: Add feedback to categories page

klinzo:

Tests will be added in a different task.
Design has to be improved, should also be done in a different task.


src/locales.json
```diff
@@ -270,5 +270,43 @@
...
+      "news": "News",
+      "disclaimer": "Imprint and contact",
+      "searchFor": "Search for",
+      "contentOfPage": "Content of the site",
+      "send": "Send",
+      "feedback": "Feedback",
+      "positiveComment": "What can we improve?",
+      "negativeComment": "What was helpful?",
+      "feedbackType": "What is the feedback for?",
+      "negativeRating": "This site was not useful",
```
maxammann: Should be called 'This site is not useful'

src/modules/app/createReduxStore.js
```diff
@@ -8,11 +8,14 @@ import endpointReducers from '../endpoint/reducers'
...
 import defaultRoutesMap from './routesMap'
 import onBeforeChange from './onBeforeChange'
+import queryString from 'query-string'
 
 const createReduxStore = (createHistory, initialState = {}, routesMap = defaultRoutesMap) => {
   const history = createHistory()
 
-  const {reducer, middleware, enhancer} = connectRoutes(history, routesMap, {onBeforeChange: onBeforeChange})
+  const {reducer, middleware, enhancer} = connectRoutes(history, routesMap,
+    {onBeforeChange: onBeforeChange, querySerializer: queryString}
```
maxammann: So query-string is like a dependency for redux-first-router?

src/routes/feedback/components/Feedback.js
```diff
@@ -0,0 +1,216 @@
...
+    this.setState({selectedFeedbackOption: selectedDropdown})
+  }
+
+  componentDidUpdate (prevProps: PropsType) {
+    const {isOpen} = this.props
+    const prevIsOpen = prevProps.isOpen
+
+    // If the Feedback is opened, we have to reset and initialize the state
+    if (prevIsOpen !== isOpen && isOpen) {
+      /* eslint-disable react/no-did-update-set-state */
```
maxammann: We have to find a solution if we want to change linting rules. I think its easier to find disabled linting rules if we collect them in the project eslint config, instead of having lines like this spread accross the project.

Also you can enable lines again after disabling them. So you only disable it for a selected portion of the code.

src/routes/feedback/components/Feedback.js
```diff
@@ -0,0 +1,216 @@
...
+  t: TFunction
+}
+
+type StateType = {
+  feedbackOptions: Array<FeedbackDropdownType>,
+  selectedFeedbackOption: FeedbackDropdownType,
+  comment: string
+}
+
+class Feedback extends React.Component<PropsType, StateType> {
```
maxammann: I think the name FeedbackBox would be better here, like in the styles.

src/routes/feedback/components/Feedback.js
```diff
@@ -0,0 +1,216 @@
...
+
+    options.push(this.getFeedbackOption(t('technicalTopic'), CATEGORIES_FEEDBACK_TYPE))
+
+    return options
+  }
+
+  getFeedbackOption = (label: string, feedbackType: string | null): FeedbackDropdownType =>
+    ({value: label, feedbackType, label})
+
+  getCurrentPageFeedbackLabel = (): ?string => {
```
maxammann: You can refactor this component by extracting the components of the form into separate components.
Like a FeedbackDropdown, FeedbackInput.

That would make this component less complex.

src/routes/feedback/components/__tests__/__snapshots__/FeedbackLink.spec.js.snap
```diff
@@ -0,0 +1,23 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`FeedbackLink should render a negative FeedbackLink 1`] = `
+<withProps(Connect(Link))
+  to="undefined?feedback=down"
```
maxammann: I think pathname shouldn't be undefined

src/routes/feedback/components/__tests__/__snapshots__/FeedbackLink.spec.js.snap
```diff
@@ -0,0 +1,23 @@
...
+  <FontAwesome
+    data-tip="negativeRating"
+    name="frown-o"
+  />
+</withProps(Connect(Link))>
+`;
+
+exports[`FeedbackLink should render a positive FeedbackLink 1`] = `
+<withProps(Connect(Link))
+  to="undefined?feedback=down"
```
maxammann: Same here.

src/modules/layout/containers/__tests__/__snapshots__/LocationLayout.spec.js.snap
```diff
@@ -35,9 +35,23 @@ exports[`LocationLayout should show CategoriesToolbar if current route is catego
...
+            "_eventsEnabled": undefined,
+            "_extrasEnabled": undefined,
+            "_live": undefined,
+            "_name": "Mambo No. 5",
+            "_sortingName": undefined,
+          },
+        ]
+      }
       city="city1"
+      feedbackType={undefined}
```
maxammann: Quite some fields are undefined in this snap. You can try to add flow to fix this "automatically" :)

maxammann: But probably we are ok with this because we do not test for these values...

src/modules/endpoint/FeedbackEndpoint.js
```diff
@@ -0,0 +1,82 @@
+// @flow
```
maxammann: Did you remove the FeedbackEndpointBuilder? In the last PR I did not get that there are more than one endpoints. Can not find the comment but there was no discussion.

src/modules/app/createReduxStore.js
klinzo: nope, this is just so that you can write <Link to={'path?withQuery'}/>  instead of <Link to={type={} payload={} query={}/>

src/routes/feedback/components/Feedback.js
klinzo: This is just a temporary fix, I want to delete these rules entirely from our eslint config, but I still don't have permissions to push to the eslint project.

src/routes/feedback/components/Feedback.js
klinzo: Thought about it, also did it, but imo it does not accomplish it (at least with FeedbackDropdown, because the dropdowns still have to live in the state of the Feedback(Box). Maybe it would work for FeedbackInput.

src/routes/feedback/components/__tests__/__snapshots__/FeedbackLink.spec.js.snap
klinzo: Testing will be done in a separate PR.

klinzo: This is just a leftover from WEBAPP-311.

src/modules/endpoint/FeedbackEndpoint.js
klinzo: Yes I removed it. There is just this FeedbackEndpoint.

src/modules/app/createReduxStore.js
maxammann: Do we need this? I like the second way more because you do not need to build strings.

klinzo: Okay, changed it in #181. query-string is needed anyway.

src/routes/feedback/components/FeedbackBox.js
```diff
@@ -0,0 +1,205 @@
...
+  componentDidUpdate (prevProps: PropsType) {
+    const {isOpen} = this.props
+    const prevIsOpen = prevProps.isOpen
+
+    // If the FeedbackBox is opened, we have to reset and initialize the state
+    if (prevIsOpen !== isOpen && isOpen) {
+      /* eslint-disable react/no-did-update-set-state */
+      const feedbackOptions = this.getFeedbackOptions()
+      const selectedFeedbackOption = feedbackOptions[0]
+      this.setState({feedbackOptions: feedbackOptions, selectedFeedbackOption: feedbackOptions[0], comment: ''})
```
maxammann: Why do you need to copy the props into the state here?

maxammann:
Maybe we should use a hash param instead of a query string. What do you think? 
https://stackoverflow.com/questions/15238391/hash-params-vs-url-params-when-to-use-which

Schedulaar:
Looking very good :+1:
I think there should be some feedback once the user has given some feedback...^^
Meaning: Once the user sends the feedback, we should maybe show some "Thanks for your feedback" window, so the user knows, the feedback has been sent successfully.

maxammann:
I agree on the feedback. Maybe we can implement this by changing the smiley icon to some other icon?

src/routes/feedback/components/FeedbackBox.js
klinzo: This resets the feedback options to choose from as well as the entered comment. Otherwise it would stay the same no matter where you open the feedback box from. Which means if you open the feedback box once in /augsburg/de and the next time in /augsburg/de/willkommen, you will not have the option "Inhalt der Seite Willkommen" and the same comment prefilled which you typed for /augsburg/de. The reason for this is that the FeedbackBox always lives in the background.

klinzo:
I'll create a task for adding a feedback for the feedback, it was planned but I forgot to create a task.

src/routes/categories/containers/CategoriesToolbar.js
```diff
@@ -9,31 +9,55 @@ import ToolbarItem from '../../../modules/layout/components/ToolbarItem'
...
+    const feedbackType = location.query && location.query.feedback
+    const category = this.props.categories.findCategoryByPath(location.pathname)
     if (!category) {
       return null
     }
     return <Toolbar>
-      <ToolbarItem name='file-pdf-o' text={this.props.t('createPdf')} href={this.getPdfUrl(category)} />
+      <ToolbarItem name='file-pdf-o' text={t('createPdf')} href={this.getPdfUrl(category)} />
+      <FeedbackToolbarItem isPositiveRatingLink location={location} />
+      <FeedbackToolbarItem isPositiveRatingLink={false} location={location} />
```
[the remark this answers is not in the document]
klinzo: The width is the same, only the height differs, but I think this okay because the height of a document is bigger than its width, whereas the width and height of a smiley should be the same. Or did I misunderstood you? But I could also force them to be of the same size of course.

src/routes/categories/containers/CategoriesToolbar.js
```diff
@@ -9,31 +9,55 @@ import ToolbarItem from '../../../modules/layout/components/ToolbarItem'
...
+      <FeedbackModal
```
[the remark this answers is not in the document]
klinzo: Yep sure, good idea. This also fixed the issue with the lost focus, which is quite logic :D. If we want to display feedback in non LocationLayout routes we may have to rethink our approach.

klinzo:
@maxammann I think hash params would also work, I think query params are more appropriate though, since we use it not like a bookmark or something like that but to show a completely different view. So with hash params I would expect just scrolling to a specific part of the page. But maybe I understood something wrong...

src/routes/categories/containers/CategoriesToolbar.js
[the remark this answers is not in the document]
Schedulaar: If you look closely, the "margin" of the icons vary (from `pdf` to `:)` there's less margin than from `:)` to `:(` )

src/modules/layout/containers/LocationLayout.js
```diff
@@ -64,20 +97,19 @@ export class LocationLayout extends React.Component<PropsType, StateType> {
...
-                                                                        categories={categories} />}>
-        {children}
+                   toolbar={showCategoriesToolbar && cities && <CategoriesToolbar categories={categories}
+                                                                                  location={location} />}>
+
+        <React.Fragment>
+          {children}
+          {this.renderFeedbackModal()}
+        </React.Fragment>
       </Layout>
```
Schedulaar: That way, the header is displayed above the feedbackmodal. To fix that do the following:
here: 
```jsx
<React.Fragment>
      <Layout asideStickyTop={this.state.asideStickyTop}
                   header={<LocationHeader isEventsEnabled={cityModel.eventsEnabled}
                                           isExtrasEnabled={cityModel.extrasEnabled}
                                           onStickyTopChanged={this.onStickyTopChanged} />}
                   footer={<LocationFooter city={city} language={language} />}
                   toolbar={showCategoriesToolbar && cities && <CategoriesToolbar categories={categories}
                                                                                  location={location} />}>

          {children}
      </Layout>
      {this.renderFeedbackModal()}
    </React.Fragment>
```

Plus some add. changes (see below)

src/routes/feedback/components/FeedbackModal.js
```diff
@@ -0,0 +1,75 @@
...
+import React from 'react'
+import FeedbackBox from './FeedbackBox'
+import CityModel from '../../../modules/endpoint/models/CityModel'
+import styled from 'styled-components'
+import CleanLink from '../../../modules/common/components/CleanLink'
+import type { LocationState } from 'redux-first-router'
+import { goToFeedback } from '../../../modules/app/routes/feedback'
+
+const Overlay = styled(CleanLink)`
+  position: fixed;
```
Schedulaar: This should be `position: absolute;`

src/routes/feedback/components/FeedbackModal.js
```diff
@@ -0,0 +1,75 @@
...
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  background-color: ${props => props.theme.colors.textSecondaryColor};
+  z-index: 3;
```
Schedulaar: you can remove the z-index here

src/routes/feedback/components/FeedbackModal.js
```diff
@@ -0,0 +1,75 @@
...
+  opacity: 0.9;
+`
+
+const ModalContainer = styled.div`
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  display: ${props => props.isOpen ? 'flex' : 'none'};
```
Schedulaar: ... and add a z-index here

src/modules/layout/containers/LocationLayout.js
Schedulaar: Attention: If you do this, you need to inject the font-families for the feedbackModal again...
Maybe it would be nice, to extend the Layout component to allow for Modals...

klinzo: Like with toolbar, header and stuff?

klinzo: Right, I thought I checked it. :/

src/routes/feedback/components/FeedbackModal.js
klinzo: Right.

klinzo:
We want a different url so that going back does not leave the current page but  just closes the feedback (important for mobile devices). All right, so you would prefer hash params @Schedulaar? If we put it just in the state it would work though (without the going back thing).

src/routes/categories/containers/CategoriesToolbar.js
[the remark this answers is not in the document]
klinzo: Ahh right, I just looked at the size. Fixed it.

maxammann:
@Schedulaar I think search engines never add query parameters if not needed. Else there would be no difference to normal pahts. But I'm just guessing.

Schedulaar:
@maxammann I don't really get what you want to say :/ but I think Search Engines look at queried paths just like at all other paths..

maxammann:
@Schedulaar woops edited it. I think they behave differently. But we would beed to deep dive into this topic to get this right. Maybe the easiest is too look at 2-3 examples and do it like they do.

maxammann:
Github uses a hashtag if you click on "Add review" in a PR. This opens the popup with the review options directly.

src/routes/feedback/components/FeedbackModal.js
```diff
@@ -0,0 +1,75 @@
...
+  left: 0;
+  z-index: 2;
+  display: ${props => props.isOpen ? 'flex' : 'none'};
+  align-items: center;
+  justify-content: center;
+`
+
+const FeedbackContainer = styled.div`
+  position: relative;
+  z-index: 3;
```
Schedulaar: This is also unnecessary, if you apply the suggestion below...
(Reduce z-indices because they're strange^^)

src/routes/feedback/components/FeedbackModal.js
```diff
@@ -0,0 +1,75 @@
...
+class FeedbackModal extends React.Component<PropsType> {
+  render () {
+    const {location, isOpen} = this.props
+    return (
+      <div>
+        <ModalContainer isOpen={isOpen}>
+          <FeedbackContainer>
+            <FeedbackBox {...this.props} />
+          </FeedbackContainer>
+          <Overlay to={goToFeedback(location)} />
```
Schedulaar: Just move `<Overlay />` to the top, like:
```jsx
      <div>
        <ModalContainer isOpen={isOpen}>
          <Overlay to={goToFeedback(location)} />
          <FeedbackContainer>
            <FeedbackBox {...this.props} />
          </FeedbackContainer>
        </ModalContainer>
      </div>
```

Schedulaar:
Also somehow the modal has a darker background if you click on "Nicht das Richtige gefunden?" than when you click on `:)` on a category page

src/routes/feedback/components/FeedbackHeader.js
```diff
@@ -0,0 +1,58 @@
...
+
+export class FeedbackHeader extends React.Component<PropsType> {
+  render () {
+    const {t, location, selectedFeedbackOption, feedbackOptions, onFeedbackOptionChanged} = this.props
+
+    return (
+      <React.Fragment>
+        <Header>
+          <Title>{t('feedback')}</Title>
+          <CloseButton to={goToFeedback(location)}>x</CloseButton>
```
Schedulaar: Also we may want to consider using `<CloseButton to={goToFeedback(location)} replace={true}>`. That way, after clicking on the CloseButton and then clicking back, doesn't open the feedback-menu again, but rather navigate to the page before that... I think that's what users would expect (at least me^^)

src/routes/feedback/components/FeedbackBox.js
```diff
@@ -0,0 +1,205 @@
...
+            selectedFeedbackOption={selectedFeedbackOption}
+            feedbackOptions={feedbackOptions}
+            onFeedbackOptionChanged={this.onFeedbackOptionChanged} />
+        )}
+        <FeedbackComment
+          comment={comment}
+          commentMessageOverride={commentMessageOverride}
+          isPositiveRatingSelected={isPositiveRatingSelected}
+          onCommentChanged={this.onCommentChanged} />
+        <SubmitButton to={goToFeedback(location)} onClick={this.onSubmit}>
```
Schedulaar: Same here as with the CloseButton (see below)

src/routes/feedback/components/FeedbackModal.js
Schedulaar: And apply the same as with CloseButton (see above/below)

src/routes/feedback/components/FeedbackHeader.js
klinzo: That way, clicking on the back button does nothing, because the stack is like this:
Page1 < Page2 < FeedbackForPage2. So if you click on the close button/overlay/submit button with replace=true, you remove FeedbackForPage2 and add Page2 to the history, so it looks like this: Page1 < Page2 < Page2, so a click on back does nothing. I would leave it like this for now.

klinzo:
@Schedulaar you're right, the reason was that there were to modals on the SearchPage, on from the SearchPage itself and one  from the locationLayout.

src/routes/feedback/components/FeedbackHeader.js
Schedulaar: Oh yeah, you're right. Convinced :)

PR #177, opened 2018-07-27
WEBAPP-210: Add feedback to search results

klinzo:

### DIFF
--- a/src/routes/feedback/components/Feedback.js
+++ b/src/routes/feedback/components/Feedback.js
@@ -54,15 +54,15 @@ const Title = styled.div`
   font-size: ${props => props.theme.fonts.subTitleFontSize};
 `
 
-export const Description = styled.div`
+const Description = styled.div`
   padding: 10px 0 5px;
 `
 
-export const CommentField = styled.textarea`
+const CommentField = styled.textarea`
   resize: none;
 `
 
-export const SubmitButton = styled(CleanLink)`
+const SubmitButton = styled(CleanLink)`
   margin: 15px 0;
   padding: 5px;
   background-color: ${props => props.theme.colors.themeColor};
@@ -71,7 +71,7 @@ export const SubmitButton = styled(CleanLink)`
   border-radius: 0.25em;
 `
 
-export type FeedbackDropdownType = {
+type FeedbackDropdownType = {
   value: string,
   feedbackType: string | null,
   label: string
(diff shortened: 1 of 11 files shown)


src/routes/feedback/components/Feedback.js
```diff
@@ -71,7 +71,7 @@ const SubmitButton = styled(CleanLink)`
   border-radius: 0.25em;
 `
 
-export type FeedbackDropdownType = {
+type FeedbackDropdownType = {
   value: string,
   feedbackType: string | null,
```
maxammann: You can use `feedbackType: ?string`

src/routes/search/containers/SearchPage.js
```diff
@@ -54,6 +83,50 @@ export class SearchPage extends React.Component<PropsType, LocalStateType> {
 
   onFilterTextChange = (filterText: string) => this.setState({filterText: filterText})
 
+  renderFeedback = (categories: Array<CategoryListItemType>) => {
```
maxammann: Probalby can be extracted to a separate component.

src/routes/search/containers/SearchPage.js
```diff
@@ -54,6 +83,50 @@ export class SearchPage extends React.Component<PropsType, LocalStateType> {
...
+          )}
+          <FeedbackModal
+            query={filterText}
+            city={city}
+            cities={cities}
+            language={language}
+            route={route}
+            pathname={pathname}
+            isPositiveRatingSelected={feedbackType === POSITIVE_RATING}
+            isOpen={feedbackType !== undefined}
```
maxammann: Why do you need to check for undefined here?

I think we shouldn't make a difference between null and undefined in our code. So just go with !!feedbackType

src/routes/search/containers/SearchPage.js
klinzo: Done

maxammann:
Also I think the used font for the questions should be Open Sans

maxammann:
Also the dropdown behaves a bit weird:

If you click on one element and do not move the cursor, then the cursor still looks like "the pointing finger".

maxammann:
Design: I  think the fading animation is a bit too much. It should be quicker or not used.

klinzo:
This just happens in the dev console, doesn't it? I've never seen the feedback box looking like that outside.

maxammann:
@klinzo I think it happens if the browser window looses focus. Do we know why this happens?

klinzo:
@maxammann Right. No, I don't know why it happens. I'll create a task for this and the cursor thing. I'll also create a task for the font, since openSans looks a bit weird imo. Regarding the fading animation, there isn't one for me and I also didn't implement one.

maxammann:
2018-07-31-10 48 41-screenshot

There is one.

klinzo:
@maxammann This is the transition of the toolbar which was there before.

PR #180, opened 2018-07-28
WEBAPP-325: Fix url encoding

maxammann:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,7 +7,8 @@ pipeline {
         sh 'yarn'
         sh 'yarn run flow'
         sh 'yarn run lint'
-        sh 'yarn run test --runInBand'
+        sh 'yarn run test'
+        sh 'yarn run test --coverage'
       }
     }
     stage('Build') {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "integreat-webapp",
-  "version": "2018.07.1",
+  "version": "2018.07.2",
   "private": true,
   "engines": {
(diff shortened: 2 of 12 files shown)


src/modules/endpoint/models/CategoriesMapModel.js
```diff
@@ -30,7 +30,7 @@ class CategoriesMapModel {
    * @return {CategoryModel | undefined} The category
    */
   findCategoryByPath (path: string): ?CategoryModel {
-    return this._categories.get(normalizePath(path))
+    return this._categories.get(decodeURIComponent(normalizePath(path)))
```
Schedulaar: should we just use these globally defined functions like `decodeURIComponent` or should these be wrapped in a module (for separation / testability purposes)?

Is there an eslint rule for things like that?

maxammann: What do you mean wrapped in a module? Define a composition of  those two methods?

I don't see advantages of a separate function module.

Schedulaar: Hmmm don't know^^ I am just a little bit surprised, that we just use globally defined functions, but I think that's alright^^

maxammann: Global functions are perfectly fine as long as they do not have side effects :)

Schedulaar: Like `fetch` :P

Schedulaar:
This wants to be merged still :D

PR #186, opened 2018-08-01
WEBAPP-330 Fix loading of events because of type error

maxammann:


Schedulaar:
I'd suggest, not to open PRs from feature -> master. Even if there are hotfixes, we should try and stick to the protocol, meaning we should have everything put into develop and always be able to release the version in the dev-branch.

maxammann:
@Schedulaar that is no problem. According to git-flow you should branch from the master to create a hotfix: https://jeffkreeftmeijer.com/git-flow/

I branched from develop, but it was no problem, because where were no new changes in develop. But yeah we should remember to branch from master.

src/modules/endpoint/endpoints/__tests__/events.spec.js
```diff
@@ -39,31 +60,6 @@ describe('events', () => {
...
 
-  it('should throw if the city to map the url are missing', () => {
-    expect(() => events.mapParamsToUrl({})).toThrowErrorMatchingSnapshot()
-  })
-
-  it('should throw if the language to map the url are missing', () => {
-    expect(() => events.mapParamsToUrl({city: 'city'})).toThrowErrorMatchingSnapshot()
-  })
-
-  const toEventModel = json => {
```
maxammann: @klinzo found this function while fixing thix test. I think this test is quite old, but we should not duplicate intentional funtionality for tests.

PR #188, opened 2018-08-06
WEBAPP-334: add post parameter support for extras

Schedulaar:


src/modules/common/components/__tests__/__snapshots__/Tiles.spec.js.snap
```diff
@@ -15,6 +15,7 @@ exports[`Tiles should match snapshot 1`] = `
           "_id": 35,
           "_isExternalUrl": false,
           "_path": "/augsburg/de/willkommen/willkommen-in-augsburg",
+          "_postData": undefined,
```
Schedulaar: Fix this test

src/modules/common/components/__tests__/__snapshots__/Tiles.spec.js.snap
```diff
@@ -27,6 +28,7 @@ exports[`Tiles should match snapshot 1`] = `
           "_id": 36,
           "_isExternalUrl": false,
           "_path": "/augsburg/de/willkommen/erste-schritte",
```
Schedulaar: Fix this test

src/modules/common/components/Tile.js
```diff
@@ -33,13 +33,22 @@ class Tile extends React.Component<PropsType> {
...
         {this.getTileContent()}
       </Link>
+    } else if (!tile.postData) {
+      return <a href={tile.path} target='_blank'>
+        {this.getTileContent()}
+      </a>
+    } else {
+      const inputs = []
+      tile.postData.forEach((value, key) => inputs.unshift(<input type='hidden' value={value} key={key} name={key} />))
+      return <form method='POST' action={tile.path} target='_blank'>
```
klinzo: Is there already an example where this is used or do we have to wait for the cms? Cause klicking on the Lehrstellenradar for Augsburg just opens the same page as before without PLZ entered.

src/modules/common/components/Tile.css
```diff
@@ -28,6 +32,7 @@
 .thumbnail img {
   position: absolute;
   top: 0;
+  left: 0;
```
Schedulaar: Oh shit I just realized that I shouldn't do that... Or at least add a ' right: 0;' as well!

klinzo: It would be nice to have a linting rule preventing stuff like this.

klinzo: Yeah sure, but a linting rule demanding 'right' if you've used 'left' and the other way round should be okay. Or at least displaying a warning.f

Schedulaar: Yeah, but I'd guess we'd have to implement such a linting rule ourselves :D
(Not sure though, if it really doesn't exist :see_no_evil:)

klinzo: Yep, with a quick search I couldn't find anything suitable. Should we create a task for this?

Schedulaar: If you want you can create one, but I'm not really optimistic that that would be an easy task 😅

Schedulaar: If you want you can create one, but I'm not really optimistic that that would be an easy task 😅

src/modules/endpoint/endpoints/extras.js
```diff
@@ -8,6 +8,12 @@ import ParamMissingError from '../errors/ParamMissingError'
 
 const EXTRAS_ENDPOINT_NAME = 'extras'
 
+const createPostMap = (jsonPost: any): Map<string, string> => {
```
NKlug: `any` should not be used if we really need such a broad type here `mixed` is the better option. Doesn't linting show a warning for "unexpected use of weak type" here?

Schedulaar: Yes, I don't like it either, but we currently use `any` instead of `mixed` in all endpoints, so I wanted to stay consistent here... You can create a task for adjusting it, if you'd like :)

PR #189, opened 2018-08-07
WEBAPP-333: Show user feedback for feedback

klinzo:


klinzo:
Yep I know, gets both fixed with the next PR.

klinzo:
@Schedulaar also fixed it here now :D

PR #190, opened 2018-08-08
WEBAPP-292: Add feedback to events, extras and disclaimer

klinzo:


src/modules/layout/containers/LocationLayout.js
```diff
@@ -64,24 +72,65 @@ export class LocationLayout extends React.Component<PropsType, StateType> {
...
+      }
+    }
+
+    if (location.type === WOHNEN_ROUTE && extras) {
+      const extra = extras.find(extra => extra.alias === WOHNEN_EXTRA)
+      if (extra) {
+        alias = extra.alias
+        title = extra.title
+      }
+    }
```
Schedulaar: A more compact version of lines 83-97:
```js
if ([SPRUNGBRETT_ROUTE, WOHNEN_ROUTE].includes(location.type) && extras) {
  const extra = extras.find(extra => [SPRUNGBRETT_EXTRA, WOHNEN_EXTRA].includes(extra.alias))
  if (extra) {
    alias = extra.alias
    title = extra.title
  }
}
```

Schedulaar: Also: In each of these `if`-clauses:
What should happen, if `category` / `event` / `extra` is not found?
I think we should render some kind of error in that case

src/modules/layout/containers/__tests__/__snapshots__/LocationLayout.spec.js.snap
```diff
@@ -23,15 +23,15 @@ exports[`LocationLayout should show CategoriesToolbar if current route is catego
...
+            "_extrasEnabled": false,
+            "_live": true,
             "_name": "Mambo No. 5",
-            "_sortingName": undefined,
+            "_sortingName": "Mambo",
           },
         ]
       }
-      feedbackType={undefined}
+      feedbackStatus={undefined}
```
Schedulaar: feedbackStatus should never be undefined. Sth's wrong with the test.

src/modules/layout/containers/__tests__/__snapshots__/LocationLayout.spec.js.snap
```diff
@@ -119,15 +105,15 @@ exports[`LocationLayout should show LocationHeader and LocationFooter if city is
...
```
Schedulaar: Same here

src/modules/layout/containers/__tests__/__snapshots__/LocationLayout.spec.js.snap
klinzo: Yes it should, as this is a query param which isn't always there. We display the modal only if it is there.

src/modules/layout/containers/LocationLayout.js
Schedulaar: then we should throw an error, so that we at least have some point to refer to, if the code doesn't behave as expected.

src/modules/layout/containers/__tests__/__snapshots__/LocationLayout.spec.js.snap
Schedulaar: But in the PropsType of FeedbackModal, you defined, that `feedbackStatus` is always given a string-value (which can't be undefined)

Schedulaar: So the connect() thingy is breaking flow to recognize the PropsType properly

src/modules/layout/containers/LocationLayout.js
klinzo: I don't agree. The only thing this does here is that the FeedbackModal is only rendered if there is a category/event/extra. If not, no modal is rendered, which is perfectly okay, because there should be an error displayed by a Page or the Switcher anyway.

src/modules/layout/containers/__tests__/__snapshots__/LocationLayout.spec.js.snap
klinzo: Yep you're right. I think LocationLayout shouldn't be connected to the state anyway. But I think the problem here is that we don't have flow types for query params in state...

src/modules/layout/containers/LocationLayout.js
klinzo: Where should we handle this error?

klinzo: @Schedulaar ?

Schedulaar:
Is it possible, when you're giving feedback on the extras page to select any extra in the dropDown, to give detailed feedback on a specific extra? That would be cool I think...
Also the extras-Feedback show up as "Verfügbare Kategorien" in the cms... I think we should change that...

klinzo:
You mean if you're on /augsburg/de/extras, there should be a dropdown for every available extra? RIght now, you can only give for internal extras like Sprungbrett and Wohnen. This is of course possible, but I would do this in a different task because I think I have to change some things.

klinzo:
Not quite sure if it makes sense to differ here, at least I agreed with @dkehne that we shouldn't differ for the users (always just show `Content of Augsburg` instead of a different dropdown for extras and categories).

src/modules/layout/containers/LocationLayout.js
klinzo: Okay, np.

klinzo:
Okay, I think it makes sense. I will change it.

dkehne:
Any more comments or discussion needed here?

Schedulaar:
Also (maybe out of scope): If the user gives feedback on technical issues, we only receive it as "Verfügbare Kategorien". Here we should maybe have sth like "Technische Funktionen auf WebApp bei 'Aktuelle Route / Seite / Extra / Whatever'" and not only post it in the city's comments section, but to some central place, where all dev's have access to.
Maybe we also need to discuss this cms...

src/routes/feedback/components/FeedbackModal.js
```diff
@@ -53,7 +53,7 @@ type PropsType = {
   id?: number,
   alias?: string,
   query?: string,
-  feedbackStatus: string,
+  feedbackStatus: ?string,
```
Schedulaar: This only allows for nulls, still not for undefined...

klinzo: No, see https://flow.org/en/docs/types/maybe. Maybe types accept the provided type as well as null or undefined. So ?number would mean number, null, or undefined.

klinzo:
I think we agreed to send technical feedback/ feedback for the whole app to the IntegreatInstance. I will speak with timo again and will fix this in https://github.com/Integreat/integreat-webapp/pull/196.

src/modules/layout/containers/LocationLayout.js
Schedulaar: not yet applied :)

dkehne:
The seperatation between content and technical app is really hard to see for the users. That's why i mentioned the idea of "content", which is related to the content, how we understand it internally, but also content like extras, which is also content for the outstanding user. In also see that this point of having an access for the developers to feedback is important: We can maybe summarize all comments at one central place e.g. network instance?

PR #191, opened 2018-08-09
WEBAPP-245: Add flow to categories module

klinzo:


src/routes/categories/components/CategoryListItem.js
```diff
@@ -4,11 +4,58 @@ import * as React from 'react'
 
 import CategoryModel from 'modules/endpoint/models/CategoryModel'
 import iconPlaceholder from '../assets/IconPlaceholder.svg'
-import {
-  CategoryCaption,
-  CategoryThumbnail, Row,
-  StyledLink, SubCategoryCaption, SubCategory
-} from './CategoryListItem.styles'
```
Schedulaar: Why did you move these out of the .styles folder?

src/routes/categories/components/CategoryTimeStamp.js
```diff
@@ -3,8 +3,15 @@
 import React from 'react'
 import type { TFunction } from 'react-i18next'
 import { translate } from 'react-i18next'
-import { TimeStamp } from './CategoryTimeStamp.styles'
```
Schedulaar: Same here

src/routes/categories/components/CategoryListItem.js
klinzo: I think we agreed on doing this, see https://issues.integreat-app.de/browse/WEBAPP-314

Schedulaar: Hmm never seen this task before :D

But it should also be made within this task then  🙃

klinzo: Hmm I don't see a problem to do this here, makes the other task way smaller.

Schedulaar: Not that you need to revert it, but we shouldn't randomly start doing parts of a some different task :)
Also I'm not yet convinced, that we should even do it ^^

PR #192, opened 2018-08-10
WEBAPP-255 add flow to common

Schedulaar:

### DIFF
--- a/src/modules/common/components/Caption.styles.js
+++ b/src/modules/common/components/Caption.styles.js
@@ -1,6 +1,6 @@
-import styled from 'styled-components'
+// @flow
 
-// fixme: H1 is connected to redux state -> not a plain old component
+import styled from 'styled-components'
 
 export const H1 = styled.h1`
   margin: 25px 0;
--- a/src/modules/common/components/Headroom.js
+++ b/src/modules/common/components/Headroom.js
@@ -1,5 +1,6 @@
 // @flow
 
+import type { Element, Node } from 'react'
 import * as React from 'react'
 import { HeaderWrapper } from './Headroom.styles'
 
@@ -15,19 +16,19 @@ type DirectionType = 'up' | 'down'
 
 type PropsType = {
   /** The child node to be displayed as a header */
-  children: Array<React.Node>,
+  children: Node,
   /** The maximum amount of px the header should move up when scrolling */
   scrollHeight: number,
   /** The minimum scrollTop position where the transform should start */
   pinStart: number,
   /** Gets rendered with a corresponding stickyTop prop as an ancestor */
-  stickyAncestor: React.Element<*>,
+  stickyAncestor?: Element<*>,
   /** Used for rendering stickyTop position of stickyAncestor */
-  height: number,
+  height?: number,
   /** Fired, when Headroom changes its state. Passes stickyTop of the ancestor. */
-  onStickyTopChanged: ?(number) => void,
+  onStickyTopChanged?: (number) => void,
   /** True, if sticky position should be disabled (e.g. for edge 16 support) */
-  positionStickyDisabled: ?boolean
+  positionStickyDisabled?: boolean
 }
 
 type StateType = {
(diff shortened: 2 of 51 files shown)


src/modules/common/components/ScrollingSearchBox.js
```diff
@@ -1,26 +1,35 @@
...
-    filterText: PropTypes.string.isRequired,
-    onFilterTextChange: PropTypes.func.isRequired,
-    spaceSearch: PropTypes.bool,
-    children: PropTypes.element.isRequired,
-    placeholderText: PropTypes.string.isRequired
-  }
+type PropsType = {
+  filterText: string,
+  onFilterTextChange: (string) => void,
+  spaceSearch: boolean,
```
NKlug: This should be `?boolean` because it wasn't/isn't required, shouln't it?

Schedulaar: It's not required because it defaults to false.

src/routes/landing/components/FilterableCitySelector.js
```diff
@@ -42,7 +39,8 @@ export class FilterableCitySelector extends React.Component<
         <ScrollingSearchBox
           filterText={filterText}
           onFilterTextChange={this.onFilterTextChange}
-          placeholderText={t('searchCity')}>
+          placeholderText={t('searchCity')}
+          spaceSearch={false}>
```
klinzo: Also defaults to false.

src/modules/common/components/ScrollingSearchBox.js
```diff
@@ -1,26 +1,35 @@
...
+  children: Element<*>,
```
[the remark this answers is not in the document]
Schedulaar: This is intentionally React.Element because it defaults see React.cloneElement which only works with React elements

src/modules/common/components/Headroom.js
```diff
@@ -153,9 +154,11 @@ class Headroom extends React.PureComponent<PropsType, StateType> {
       scrollHeight,
       positionStickyDisabled
     } = this.props
-    const { mode, transition } = this.state
+    const {mode, transition} = this.state
     const transform = mode === UNPINNED ? -scrollHeight : 0
-    const stickyTop = Headroom.calcStickyTop(mode, height, scrollHeight)
+    if (stickyAncestor && !height) {
+      console.warn('Headroom: If you supply stickyAncestor, you must define height!')
```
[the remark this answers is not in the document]
Schedulaar: Because it doesn't break the display, but only glitches a bit... If I throw an error, the complete website would stop working. (But I also thought about throwing an error)...

src/modules/common/components/Page.js
```diff
@@ -1,20 +1,20 @@
...
+  content: string
+}
 
+class Page extends React.Component<PropsType> {
   render () {
     return (
       <div>
         <Caption title={this.props.title} />
-        <RemoteContent dangerouslySetInnerHTML={{__html: this.props.content}} />
+        <RemoteContent dangerouslySetInnerHTML={{__html: this.props.content}} centered={false} />
```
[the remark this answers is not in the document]
Schedulaar: Hmm.. yeah true, but also not really harmful :see_no_evil:

src/modules/common/components/LoadingSpinner.styles.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+
+import styled from 'styled-components'
+import Spinner from 'react-spinkit'
+
+export const CenteredSpinner = styled(Spinner)`
```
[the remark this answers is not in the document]
Schedulaar: Haha yeah true. Touché :D

src/modules/common/components/Caption.styles.js
```diff
@@ -1,6 +1,6 @@
-import styled from 'styled-components'
+// @flow
 
-// fixme: H1 is connected to redux state -> not a plain old component
```
maxammann: It is not connected to redux but to an other context. So this is still not a plain component.

Schedulaar: Then we probably need to adjust our definition of component/container, otherwise everythings a container (almost)

maxammann: Yep, will write one after thinking about them in the react native app. It actually is pretty clear an makes sense. A lot of containers will be pretty simple and just contain one or two lines.

PR #193, opened 2018-08-10
WEBAPP-308: Simplify switcher

klinzo:


src/modules/app/containers/Switcher.js
```diff
@@ -64,67 +64,55 @@ type PropsType = {
...
+  static renderFailureLoadingComponents = (payloads: Array<Payload<any>>): Node => {
+    for (let i = 0; i < payloads.length; i++) {
+      const payload = payloads[i]
+      if (payload.error) {
+        return <FailureSwitcher error={payload.error} />
+      } else if (payload.isFetching || !payload.data) {
+        return <LoadingSpinner />
+      }
     }
+    return null
```
Schedulaar: That implementation is kind of weird I think... I suggest the following:
```jsx
const payloadWithError = payloads.find(payload => payload.error)
if (payloadWithError) {
  return <FailureSwitcher error={payloadWithError.error} />
} else if (payloads.find(payload => payload.isFetching || !payload.data)) {
  return <LoadingSpinner />
}
return null; // I'm surprised, that the type Node allows nulls...
```
(That way, we also treat displaying failures with a higher priority than displaying loadingSpinners)

klinzo: Way better than mine, thank you!

klinzo: But I even think we should display loadingSpinners with a higher priority, otherwise we sometimes display an failure for  a short time and than display something else which looks kind of weird. For example if you're on `/augsburg/xy` the categories endpoint fails instantly, whereas the cities and languages are still being loaded. This way,  an failure is displayed until these endpoints are loaded and afterwards a LanguageSwitcher is displayed.

Schedulaar: Hmm.. I'm not really sure, how we currently fetch stuff^^
I'd expected, that once we change our routes, all corresponding payloads are reset (if necessary), until the request is finished. Thus there shouldn't exist any endpoints that represent failures and at the same time, after waiting for other payloads are loaded, start loading themselves :thinking: (But I'm not sure tbh^^ - At least that how I thought it *should* be handled)

klinzo: No, I think you misunderstood me. We start fetching cities, languages and categories at the same time, but categories fails instantly cause `xy` is an invalid language, whereas languages and cities can be fetched.

Schedulaar: Hmm... I see... there are some weird dependencies there... Yeah, then we should probably do it the other way around...

src/modules/app/containers/Switcher.js
```diff
@@ -64,67 +64,53 @@ type PropsType = {
...
-   * @param payload The payload to check for errors or fetching process
+   * Renders a failure component if a payload contains an error or a LoadingSpinner if the data is still being fetched
+   * @param payloads The payloads to check for errors or fetching process
    * @return {*}
    */
-  static renderFailureLoadingComponents = (payload: Payload<any>): Node => {
-    if (payload.error) {
-      return <FailureSwitcher error={payload.error} />
-    } else if (payload.isFetching || !payload.data) {
+  static renderFailureLoadingComponents = (payloads: Array<Payload<any>>): ?Node => {
```
NKlug: You should rather use `mixed` instead of `any`. Best would, of course, be to use neither of them.

src/modules/app/containers/Switcher.js
```diff
@@ -136,52 +122,44 @@ export class Switcher extends React.Component<PropsType> {
     throw new Error(`For the route ${currentRoute} was no content selected in the Switcher which should be displayed.`)
```
NKlug: Just a grammar thing: I think it should be "No content was selected for route ${currentRoute} to be displayed" or something similar :)

src/modules/app/containers/Switcher.js
Schedulaar: This is a perfect example to use `Payload<*>` here.
http://sitr.us/2015/05/31/advanced-features-in-flow.html#existential-types

src/modules/app/containers/Switcher.js
```diff
@@ -68,14 +68,12 @@ export class Switcher extends React.Component<PropsType> {
...
-      if (payload.error) {
-        return <FailureSwitcher error={payload.error} />
-      } else if (payload.isFetching || !payload.data) {
-        return <LoadingSpinner />
-      }
+  static renderFailureLoadingComponents = (payloads: Array<Payload<any>>): ?Node => {
+    const errorPayload = payloads.find(payload => payload.error)
+    if (payloads.find(payload => (payload.isFetching || !payload.data) && !payload.error)) {
+      return <LoadingSpinner />
+    } else if (errorPayload && errorPayload.error) {
```
Schedulaar: Why do we need to check `errorPayload.error` here? Because of flow?

klinzo: Yep, flow does not recognize it.

PR #195, opened 2018-08-10
WEBAPP-247: Add flow to events, extras, main disclaimer, i18n redirect route

klinzo:


src/routes/i18nRedirect/containers/__tests__/I18nRedirectPage.spec.js
```diff
@@ -28,7 +30,7 @@ describe('I18nRedirectPage', () => {
   describe('get redirect action', () => {
     it('should return goToLanding action if there is no param or the param is landing', () => {
       const instanceWithoutParam = shallow(
-        <I18nRedirectPage cities={cities} redirect={() => {}} />,
+        <I18nRedirectPage cities={cities} redirect={() => {}} param={undefined} />,
```
Schedulaar: I think we should never pass `undefined` explicitly as a prop :thinking:

PR #197, opened 2018-08-16
WEBAPP-250: Add flow to landing and search route

klinzo:


src/routes/landing/components/Heading.js
```diff
@@ -1,22 +1,34 @@
...
-  static propTypes = {
-    t: PropTypes.func.isRequired
-  }
+const Logo = styled.img`
+  display: block;
+  height: 70px;
+  margin: 0 auto;
+`
+
+const StyledCaption = styled(Caption)`
```
Schedulaar: We should just remove this and stick with the unstyled Caption. (Since this didn't have any effect before and I don't think we need this slight adjustment of margin here...)

klinzo: I think it does not really matter, so yes, I'll remove it.

PR #198, opened 2018-08-16
WEBAPP-254: Add flow to app

klinzo:


src/modules/app/createReduxStore.js
```diff
@@ -9,8 +11,19 @@ import { createResponsiveStateReducer, responsiveStoreEnhancer } from 'redux-res
...
+
+declare var __DEV__: boolean
+
+type ActionType = startFetchAction | finishFetchAction<*>
 
-const createReduxStore = (createHistory, initialState = {}, routesMap = defaultRoutesMap) => {
+// todo: Change type to correct State type,
+// https://blog.callstack.io/type-checking-react-and-redux-thunk-with-flow-part-2-206ce5f6e705
+const createReduxStore = (createHistory: () => History, initialState: {} = {},
+  routesMap: RoutesMap = defaultRoutesMap): Store<ActionType, any> => {
```
NKlug: Is `any` really necessary here? At least in the neuburg-backend eslint lints `any`.

klinzo: I have already discussed this with @Schedulaar, who is currently adding flow to the endpoint package. To define a correct type here, his PR needs to be finished. Since we have quite some occurances of any, we have to remove them at some other time, so I think this shouldn't be a problem for now. But you're right, sooner or later this has to be changed.

PR #200, opened 2018-08-17
WEBAPP-337: Add wohnen support for testumgebungwohnraum

Schedulaar:


src/routes/wohnen/containers/WohnenExtraPage.js
```diff
@@ -53,19 +53,19 @@ export class WohnenExtraPage extends React.Component<PropsType> {
...
+        </>
       )
     }
 
     return (
-      <React.Fragment>
+      <>
         <Helmet title={`${extra.title} - ${cityName}`} />
         <Caption title={extra.title} />
         <OfferList city={city} language={language} hashFunction={this.hash} offers={offers} />
```
NKlug: This whole hashing thing could be made easier or even completely avoided if we just return the `hashedToken` from the backend. What are your thoughts about this?

Schedulaar: Well we should rather generate an id for that, since we kind of use the hashedToken as our main authentication mechanism (like a hashed password).
Using the hashedToken would be an abuse I think...

maxammann: Yeah, the hash is just a workaround for now

PR #202, opened 2018-08-21
WEBAPP-256: Add flow to endpoint module

Schedulaar:
This PR only introduces only easy-to add types.
For better types for all the actions, reducers, etc. I think we would need to do either heavy refactorings or too many hardcoded types...
Not sure how to proceed here...


src/modules/endpoint/models/__tests__/CategoriesMapModel.spec.js
```diff
@@ -21,26 +23,49 @@ describe('CategoriesMapModel', () => {
...
   const categories = [
-    new CategoryModel({id: 0, path: '/augsburg/de', title: 'augsburg', parentPath: ''}),
+    new CategoryModel({
+      id: 0,
+      path: '/augsburg/de',
+      title: 'augsburg',
+      parentPath: '',
+      availableLanguages: new Map(),
+      content: 'exampleContent0',
+      lastUpdate: moment(),
```
NKlug: Not sure if this is important but using `moment()` here makes the tests a bit non-deterministic. Maybe a fixed date and time would be better.

src/modules/endpoint/models/__tests__/CategoryModel.spec.js
```diff
@@ -9,7 +12,8 @@ describe('CategoryModel', () => {
     parentPath: '/augsburg/en',
     order: 75,
     availableLanguages: new Map([['de', '/augsburg/de/']]),
-    thumbnail: 'https://cms.integreat-ap…/03/Hotline-150x150.png'
+    thumbnail: 'https://cms.integreat-ap…/03/Hotline-150x150.png',
+    lastUpdate: moment()
```
NKlug: Same in this file and in some others.

src/routes/feedback/components/FeedbackBoxContainer.js
```diff
@@ -0,0 +1,202 @@
...
+import FeedbackDropdownItem from '../FeedbackDropdownItem'
+import { WOHNEN_ROUTE } from '../../../modules/app/routes/wohnen'
+import { SPRUNGBRETT_ROUTE } from '../../../modules/app/routes/sprungbrett'
+import { EXTRAS_ROUTE } from '../../../modules/app/routes/extras'
+import ExtraModel from '../../../modules/endpoint/models/ExtraModel'
+import FeedbackBox from './FeedbackBox'
+import { translate } from 'react-i18next'
+import type { FeedbackDataType } from '../../../modules/endpoint/FeedbackEndpoint'
+
+type PropsType = {
```
NKlug: Afaik we want to use `{|…|}` for the props, right?

klinzo: I already did this in #201.

PR #205, opened 2018-08-22
HOTFIX: WEBAPP-338: Fix Display of events

Schedulaar:
PR for Hotfix to merge into master-branch

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "integreat-webapp",
-  "version": "2018.08.1",
+  "version": "2018.08.2",
   "private": true,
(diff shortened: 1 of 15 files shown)


klinzo:
Not using getEventsPath would also be a solution.

Schedulaar:
I don't like, that the router parses the parameter into a number (at least that's what we get from `eventId: state.location.payload.eventId`)... We cannot change this behavior right?
Also the documentation of the router is miserable -.-

Schedulaar:
True, there were some crazier issues with confusing names going on... I fixed them.

PR #212, opened 2018-08-25
WEBAPP-342: Fix events in IE

Schedulaar:
Do you have any idea, why this worked before? Why did the test not fail?


src/modules/endpoint/endpoints/events.js
```diff
@@ -77,7 +77,7 @@ export default new EndpointBuilder(EVENTS_ENDPOINT_NAME)
     .map((event: JsonEventType) => {
       const allDay = event.event.all_day !== '0'
       const availableLanguages = new Map()
-      Object.keys(availableLanguages)
+      Object.keys(event.available_languages)
```
maxammann: WTF? :joy:

PR #214, opened 2018-08-26
WEBAPP-312: Migrate css files to styled components

Schedulaar:


Schedulaar:
Sorry about the darkmode - that escalated quickly... :sweat_smile:

src/modules/layout/components/Footer.js
```diff
@@ -1,30 +1,53 @@
 // @flow
 
+import type { Node } from 'react'
```
maxammann: I prefere to use React.Node. As this does not cause confustion with the ES Node class.

maxammann:
Also this should have been splitted into 2 tasks

Schedulaar:
Yep, should have been two tasks as mentioned already :sweat_smile:

Yoas, but it only adds veery small complexity overhead. It actually helped discover problems with the switcher that was somehow broken again... (That's why the pr is so large)

Plus for development it's sometimes nice, to switch theme for a second, because then you have a different perspective on things :sweat_smile:

Also I thought about different font sizes which should be relatively easy to integrate now. High contrast mode sounds nice as well...

PR #215, opened 2018-08-27
WEBAPP-309: upgrade packages

Schedulaar:


package.json
```diff
@@ -25,137 +25,149 @@
...
-    "@babel/cli": "^7.0.0-beta.54",
-    "@babel/core": "^7.0.0-beta.54",
-    "@babel/plugin-transform-react-jsx": "^7.0.0-beta.54",
-    "@babel/plugin-transform-runtime": "^7.0.0-beta.54",
-    "@babel/preset-env": "^7.0.0-beta.54",
-    "@babel/preset-flow": "^7.0.0-beta.54",
-    "@babel/preset-react": "^7.0.0-beta.54",
-    "@babel/preset-stage-1": "^7.0.0-beta.54",
-    "@babel/runtime": "^7.0.0-beta.54",
+    "@babel/cli": "^7.0.0",
```
maxammann: Finally babel 7...

package.json
```diff
@@ -25,137 +25,149 @@
...
+    "@babel/core": "^7.0.0",
+    "@babel/plugin-proposal-class-properties": "^7.0.0",
```
maxammann: Why are all these proposals needed? A change in babel 7?

src/modules/layout/components/HeaderDropDown.js
```diff
@@ -20,16 +20,13 @@ type StateType = {
...
-  closeDropDown: Function
-  handleClickOutside: Function
-
   constructor (props: PropsType) {
     super(props)
     this.state = { dropDownActive: false }
-    this.toggleDropDown = this.toggleDropDown.bind(this)
-    this.closeDropDown = this.closeDropDown.bind(this)
-    this.handleClickOutside = this.handleClickOutside.bind(this)
+    const self: any = this // https://github.com/facebook/flow/issues/5874
```
maxammann: You can also define the functions as fields with the type of a lambda

package.json
Schedulaar: Yes, they no longer provide the stage-1 preset (or any stage-x) instead we should explicitly define the proposals we want to use. We could even define our own babel-preset

src/modules/layout/components/HeaderDropDown.js
Schedulaar: Yah, but I had some reason, not do it like this... I think flow just didn't want to accept any of my suggestions :D

PR #216, opened 2018-08-28
WEBAPP-344: Add preconnect tag

Schedulaar:


Schedulaar:
It loads some of the data we always request before/while main.js is loading (and therefore the connection to the CMS will be established already when loading main.js)

Schedulaar:
Can't provide percentages thkugh, but I could tell the difference from looking at it (so probably pretty high)

PR #217, opened 2018-08-28
WEBAPP-345: Use fort awesome libs

Schedulaar:


Schedulaar:
Yeah upgrade to fontawesome 5 plus we now only include the icons we explicitly import and not the whole iconpack in our bundle (- 450kb uncompressed)

PR #218, opened 2018-08-28
WEBAPP-346: Enforce flow annotation on all files

Schedulaar:


Schedulaar:
We need to exclude eslint for /tools/*.js, because we don't use it there. Should we create a separate eslint config for node scripts?

Schedulaar:
Nvm, I've added another .eslintrc in src folder, so that we have the only the basic .eslintrc for the tools folder.

Schedulaar:
We should highly consider creating an own flow preset right away :D

src/.eslintrc.json
```diff
@@ -0,0 +1,28 @@
...
+    "@integreat-app/eslint-config",
+    "@integreat-app/eslint-config-react",
+    "@integreat-app/eslint-config-jest",
+    "plugin:flowtype/recommended"
+  ],
+  "plugins": [
+    "flowtype"
+  ],
+  "rules": {
+    "flowtype/require-valid-file-annotation": [
```
maxammann: These rules should be in @integreat-app/eslint-config

Schedulaar: Sure, these rules, shouldn't be right here.
Take a look at https://issues.integreat-app.de/browse/LINT-4
Do you agree with that?

PR #220, opened 2018-08-31
WEBAPP-351: dont hijack some cms paths

Schedulaar:


src/routes/categories/containers/CategoriesPage.js
```diff
@@ -53,7 +54,8 @@ export class CategoriesPage extends React.Component<PropsType> {
 
   redirectToPath = (path: string) => {
     const action = pathToAction(path, this.props.routesMap)
-    this.props.redirect(action)
+    setKind(action, 'push')
```
Schedulaar: That router does suck though...

maxammann: New change router task incoming? :P

src/modules/common/components/RemoteContent.js
```diff
@@ -8,15 +8,17 @@ type PropsType = {
   dangerouslySetInnerHTML: {
     __html: string
   },
-  hosts: Array<string>,
+  hijackRegExp: RegExp,
   onInternLinkClick: string => void,
   centered: boolean
 }
 
+const HIJACK = /https?:\/\/(cms\.integreat-app\.de|web\.integreat-app\.de|integreat\.app)(?!\/[^/]*\/(wp-content|wp-admin|wp-json)\/.*).*/
```
maxammann: I'm a bit against using RegExps if this gets more complicated. For now it looks fine though.

Schedulaar: Hmm... rather a function, right? I will leave it for now.

PR #224, opened 2018-09-10
WEBAPP-343: Add rel='norefferer' to external links

klinzo:

For forms rel='noreffer'/rel='noopener' is not available. Our extras using forms(only Lehrstellenradar) can not be opened in the background btw.

### DIFF
--- a/src/modules/common/components/Tile.js
+++ b/src/modules/common/components/Tile.js
@@ -86,7 +86,7 @@ class Tile extends React.Component<PropsType> {
     } else {
       const inputs = []
       tile.postData.forEach((value, key) => inputs.unshift(<input type='hidden' value={value} key={key} name={key} />))
-      return <form method='POST' action={tile.path} target='_blank' rel='noreferrer'>
+      return <form method='POST' action={tile.path} target='_blank'>
         {inputs}
         <button type='submit'>{this.getTileContent()}</button>
       </form>
--- a/src/modules/common/components/Tile.js
+++ b/src/modules/common/components/Tile.js
@@ -82,11 +82,11 @@ class Tile extends React.Component<PropsType> {
     if (!tile.isExternalUrl) {
       return <CleanLink to={tile.path}>{this.getTileContent()}</CleanLink>
     } else if (!tile.postData) {
-      return <CleanAnchor href={tile.path} target='_blank'>{this.getTileContent()}</CleanAnchor>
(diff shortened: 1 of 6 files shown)


Schedulaar:
What do you mean by "cannot be opened in the background"? 🤔 
Also I think,  we should just add the attributes target='_blank' and rel='no-referrer' to the CleanLink. there are no other use-cases for a cleanlink afaik...

maxammann:
@Schedulaar I think the cleanest would be to add a flag to CleanLink to open something in a new tab.

Schedulaar:
@maxammann Yep, but that should default to true, because we (almost) always want to open external links in a new tab...

klinzo:
Okay, should the link to our Datenschutzerklärung also use rel='noreferrer' or just rel='noopener'? And by can't be opened in the background I meant that you can't open it in a new tab because it's a form and not a link.

klinzo:
And do we want to open the Datenschutzerklärung also in a new Tab? Right now we don't.

Schedulaar:
Id suggest to use noreferrer. Also I think it should open in a new tab.
For the Lehrstellenradar extra: for me it opens in a new tab in every browser. Which do you use?

maxammann:
Datenschutzerklärung should open in a new tab, yep.

klinzo:
Done

src/modules/common/components/CleanAnchor.js
```diff
@@ -1,8 +1,33 @@
...
+class CleanAnchor extends React.PureComponent<PropsType> {
+  static defaultProps = {
+    newTab: true
+  }
+
+  render () {
+    const {newTab, href, children} = this.props
+
+    if (newTab) {
+      return <StyledAnchor href={href} target='_blank' rel='noreferrer'>{children}</StyledAnchor>
```
Schedulaar: The CleanAnchor should pass all its other props to the StyledAnchor. Otherwise we cannot restyle it with sth like:
```js
styled(CleanAnchor)`
  some: css;
`
```
At least, that's what I believe^^

klinzo: Yeah, if we want to style it, we have to pass the className to children. Currently, we are not styling a CleanAnchor anywhere though. And right now, we only pass className if we really want to style it and not just as precaution. Because of this, I wouldn't do this here.

PR #226, opened 2018-09-10
WEBAPP-356: Show right error if an extra is not found

klinzo:

### DIFF
--- a/src/routes/categories/containers/CategoriesPage.js
+++ b/src/routes/categories/containers/CategoriesPage.js
@@ -34,7 +34,8 @@ type PropsType = {
   uiDirection: UiDirectionType,
   t: TFunction,
   dispatch: ReceivedAction => void,
-  routesMap: {}
+  routesMap: {},
+  categoryPath: string
 }
 
 /**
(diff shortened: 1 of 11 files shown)


src/routes/categories/containers/CategoriesPage.js
```diff
@@ -117,7 +118,8 @@ export class CategoriesPage extends React.Component<PropsType> {
         {this.getContent(categoryModel)}
       </div>
     } else {
-      const error = new ContentNotFoundError({type: 'category', id: this.props.path, city: city, language})
+      const errorType = categoryPath.indexOf('extras') === 0 ? 'extra' : 'category'
```
Schedulaar: No, the route should just not end up in the categoriesPage in the first place. This is too hacky, sorry :D

maxammann: Also why does the Categories page get rendered when you try to load a extra? I think we should map all routes which contain a /extras/ to the extras route.

klinzo: @maxammann you changed the ExtrasRoute when adding the WohnraumboerseExtra to the WebApp. I'll change it back :+1:

PR #230, opened 2018-09-13
WEBAPP-358: Add points of interest endpoint

klinzo:

Maybe we have to change some attributes or the url in the future. We also have to think about where to display the POIs.


src/modules/app/routes/pointsOfInterest.js
```diff
@@ -0,0 +1,23 @@
...
+import type { Dispatch, GetState } from 'redux-first-router'
+import pointsOfInterestEndpoint from '../../endpoint/endpoints/pointsOfInterest'
+
+export const POINTS_OF_INTEREST_ROUTE = 'POINTS_OF_INTEREST'
+
+export const goToPointsOfInterest = (city: string, language: string, pointOfInterestId: ?string) =>
+  createAction(POINTS_OF_INTEREST_ROUTE)({city, language, pointOfInterestId})
+
+export const getPointsOfInterestPath = (city: string, language: string, pointOfInterestId: ?string): string =>
+  `/${city}/${language}/poi${pointOfInterestId ? `/${pointOfInterestId}` : ''}`
```
Schedulaar: Hmm... It might be easier to use /locations instead of /poi -- also we use plural for extras and events, so we should use plural here as well -- : The cms will probably create (internal) links pointing to sth. with /locations as path segment, so e.g. 
https://cms.integreat-app.de/augsburg/de/locations/cafe-tuer-an-tuer (is this correct, @timoludwig / @svenseeberg ?) 
It would the easiest for us, if we just adopted this I guess (when it comes to creating sitemaps, internal navigation in webapp, etc. pp.)

timoludwig: correct, the permalinks of pois will always have `locations` in it, no matter what we call the api endpoint.

klinzo: We have so many locations already :/ LocationLayout, location of the router, cities have been locations at some time. Of course, I don't really have a problem with locations.

klinzo: Same question again :stuck_out_tongue:

Schedulaar: @maxammann opinion required

maxammann: Yes, lets go with locations if we can not change the permalinks.
But I think we should use the name poi in the client. So the we never really have to care that its named location in the backend.

maxammann: So in this example I would use the name poi.js for the route.

src/modules/app/StateType.js
```diff
@@ -10,6 +10,7 @@ import DisclaimerModel from '../endpoint/models/DisclaimerModel'
 import SprungbrettJobModel from '../endpoint/models/SprungbrettJobModel'
 import WohnenOfferModel from '../endpoint/models/WohnenOfferModel'
 import type { LocationState } from 'redux-first-router'
+import PointOfInterestModel from '../endpoint/models/PointOfInterestModel'
```
maxammann: Should we go with POI instead of PointOfInterrest? Way shorter and has a more specific meaning because its a well known abbreviation.

src/modules/endpoint/endpoints/jsonTypes/JsonCommonTypes.js
```diff
@@ -0,0 +1,24 @@
+// @flow
```
maxammann: We need to come up with a way of structuring our types. So in this case we really need types because we apply them to untyped data.
My proposal would be to create a `types.js` file in each module/route and just throw all of them in there. Thats how a lot of npm packages are doing it.
Or we create a `types` folder and structure the types there accordingly. For example: `src/modules/endpoint/types/json.js`

maxammann: We will have the same problem with the type of our state. Not sure where to put it right now in the react-native prototype.

In the webapp we do not type the state and actions.

src/modules/app/StateType.js
Schedulaar: It doesn't have a "more specific meaning" - it's literally the same meaning afaik - but I like POI better because it's shorter and its meaning is clear

maxammann: The same meaning but you see it more often as POI instead of Point Of Interrest.

src/modules/app/routes/pointsOfInterest.js
klinzo: So our route should be /augsburg/de/pois or /augsburg/de/locations? And the name of the files and routes should be poi.js?

src/modules/endpoint/endpoints/jsonTypes/JsonCommonTypes.js
klinzo: You mean our redux state?

src/modules/app/routes/pointsOfInterest.js
Schedulaar: Yes the route with locations and the files with pois

klinzo: @timoludwig could you maybe supply the permalink id as a separate attribute?  i.e. 'cafe-tuer-an-tuer' in '/augsburg/de/locations/cafe-tuer-an-tuer/'?

klinzo: @maxammann @Schedulaar sollten wir dann LocationLayout und so auch mal zu CitiesLayout umbenennen? Oder lassen wir das so wies ist?

klinzo: @timoludwig wir brauchen die permlink id dann auch in den available_languages

Schedulaar: @klinzo Afaik ist https://cms.integreat-app.de/augsburg/en/wp-json/extensions/v3/locations fast genauso aufgebaut wie der pages endpoint - insbesondere der available_languages bereich ist identisch... Sollte also also eigentlich schon gehen oder? :thinking:

timoludwig: yes, no problem. i will add the ´permalinkId´ property to all post types (pages, events, locations, disclaimer).

klinzo: @Schedulaar Ja schon, der Unterschied is nur, dass wir für Pages den kompletten path (`/augsburg/de/willkommen/erste-schritte`) verwenden, weil wir ja mehrere Ebenen haben. Das macht bei pois und events aber keinen Sinn, da reicht die pathId, zb `cafe-tür-an-tür`. Die steckt zwar im Pfad auch drin, is aber in der WebApp schöner, wenn wir die extra geliefert bekommen anstatt die aus dem path raussuchen  zu müssen. Also ja geht, is aber hässlich :D

klinzo: Hatte nen Denkfehler, dass wir die permlink Id brauchen, um den path zu bekommen, unter dem der poi oder das event aufrufbar ist... Aber den bekommen wir ja schon vom cms :man_facepalming: :unamused:

maxammann: @klinzo About the layout names. Both names are fine imo. But I'd leave it as it is until we gain benefit from a change.

src/modules/endpoint/models/PoiModel.js
```diff
@@ -0,0 +1,92 @@
+// @flow
```
maxammann: Can we maybe create a sub class here? This shares a lot of types with a Page.

klinzo: I think this just adds more complexity, because a page/category has not that many attributes anyway (you would always have to access them by catgory.subclass.shared_type...which would lead to lots of changes). Maybe it makes sense to introduce a custom location type shared by pois and events which contains stuff like town, adress, longitude and latitude and provides some methods for formatting and displaying the address. I would do this in another task though.

PR #231, opened 2018-09-18
WEBAPP-67: Show js disabled ie deprecated warning

Schedulaar:


Schedulaar:
Wow, that is funny^^ That's because NoScript (the addon) replaces `<noscript>` tags with a span tag (and therefore the styles don't apply any longer)^^. Will adjust it.

PR #232, opened 2018-09-28
WEBAPP-314: Move .styles to components file

Schedulaar:


src/modules/common/components/Breadcrumbs.js
```diff
@@ -1,8 +1,47 @@
 // @flow
 
 import * as React from 'react'
-import { Breadcrumb, OrderedList, Wrapper } from './Breadcrumbs.styles'
 import type { UiDirectionType } from '../../app/StateType'
+import styled from 'styled-components'
+
+const opposite = direction => direction === 'ltr' ? 'rtl' : 'ltr'
```
klinzo: Flow types missing

PR #233, opened 2018-09-30
WEBAPP-359: Fix opening of feedback modal

klinzo:


src/modules/feedback/components/FeedbackLink.js
```diff
@@ -18,6 +18,8 @@ type PropsType = {
   className?: string
 }
 
+const FeedbackToolbarItem = StyledToolbarItem.withComponent('div')
```
maxammann: StyledToolbarItem should have its own file if its used externally. Also `withComponent` is kind of deprecated: https://www.styled-components.com/docs/api#withcomponent

I also think we should not miss-use styles like this. For example I have absolutly no clue what kind of style FeedbackToolbarItem will have.

klinzo: We're using v3.4.5 right now, and withComponent() is only a deprecation candidate in v4 (and the polymorphic prop 'as' is only available in v4) and v4 is still beta.
For me this is not a misuse of styles, because it makes clear that FeedbackToolbarItem has quite the same style as a normal ToolbarItem, which is what we want.

I could move it to a separate file though (although we then have .styles files again).

klinzo: The style is the same for every toolbar item... The difference is that some are external links (e.g. PDF) and some are just divs doing something onClick (e.g. Feedback up and down). So it makes no sense to have a styled flag imo. Extracted it to a separate file though.

PR #234, opened 2018-10-03
WEBAPP-361: Add flow types to shared lib

Schedulaar:

Flow types are working now when importing latest version of shared package as npm module.
But still Intellij cannot resolve modules from @integreat-app/shared package... Ideas @maxammann ?


maxammann:
Should we change to relative paths for integreat-webapp in this task?

package.json
```diff
@@ -10,7 +10,7 @@
     "start": "node tools/run.js",
     "build": "node tools/build.js",
     "build:debug": "node tools/build.js --debug",
-    "build:lib": "node tools/lib.build.js",
+    "build:lib": "node tools/lib.build.js && flow-copy-source -i '**/@(__mocks__|__tests__)/**' ./src ./dist && mv ./dist/lib.js.flow ./dist/index.js.flow",
```
maxammann: Why do we have to rename lib.js.flow to index.js.flow? Can we change the name in the webpack script?

Schedulaar: Webpack builds the library into index.js which is necessary to distribute the library via npm package (you could theoretically adjust the default location of you library-root-file in the package.json, but then IDEs have a hard time resolving it -- at least when I did that the last time Intellij had no clue what was going on ^^)

Schedulaar:
> Should we change to relative paths for integreat-webapp in this task?

Yeah, I can do that.

PR #237, opened 2018-10-08
WEBAPP-363: Merge event, category and poi models

klinzo:

Sorry for the large PR, but the changes are mostly updated tests because of changes in the models. I'm not quite sure if this is a step forward, but I tend to say yes. The endpoints and models are simpler than before. So just say what you think.
As a next step, I would generalize some of our components (EventDetail, Page and the PoiDetail (does not exist yet) and EventList, CategoryList and PoiList(does not exist yet) are mostly the same (with more or less info displayed)).

Because jira is currently not accessible this has no corresponding task.


Schedulaar:
Haha yeah, in PRs like this, I wish GitHub had a function to provide a pre-defined filter to only show "important" changes...
So my thoughts are the following:

So if you want, we can do sth. like a class-hierarchy, but I don't really like your suggested structure. Your suggestion is:
```
         PageModel
      ↑              ↑
  PoiModel     CategoryModel
      ↑
EventModel
```
And the DisclaimerModel didn't fit in there. Also, imo, a PoiModel is not a superclass of EventModel (in a sementical way -- probably looking at the attributes it kindof is). But in the future, we probably also should have a reference in an EventModel to the corresponding PoiModel (the location of the event); Allowing this PoiModel to be an EventModel doesn't make sense.

Another point, I don't like, is that you generalized all the endpoint-types. We should not do that. These types represent the actual data-representation we expect to get from the cms, and should not have unnecessary optional arguments and arguments we expect should not be optional.

A suggested hierarchy would be for me (although I'm also not 100% convinced of it):
```
          BasePageModel
      ↑                     ↑
DisclaimerModel     ContentPageModel
                            ↑            ↑                ↑  
                      EventModel   CategoriesModel    PoiModel
```

klinzo:
Yeah I definitely see your point, thereto:

1. EventModel can definitely also inherit from PageModel, no problem here.
2. DisclaimerModel does fit in here, it has exactly the same attributes as categories, but I figured it would be a too big overhead to include this in this structure, since we only use three of these (title, content and modified_gmt), so it does not make sense imo to create a BasePageModel and a ContentPageModel just to fit in the Disclaimer. I would rather just use a CategoryModel for the disclaimer.
3. So the endpoints should just stay the way they were?

maxammann:
I think we should discuss this in a call. A thing that I miss in this issue/PR is a real benefit of changing our models.
We should start to groom tasks again, because there are much more important architectural stuff to change. For example to make the routing easier. This also would break the current react-native use of the endpoint package.

klinzo:
There is no real benefit apart from sticking to DRY, since our most of the attributes of our models were the same. What is the problem with react-native?

klinzo:
@maxammann My main reason for doing this is that we already have like 5 different Lists, ListIElements and also some detail views which are all roughly the same, but with different designs and approaches. I think this is not what it should be like and since I'm implementing the POI stuff, which will also need a List, ListElement and so on, this is, at least in my opinion, quite important. This PR just makes it easier for me to generalize those components, since the models are more similar than before.

Schedulaar:
@klinzo You could instead just generalize the lists; so for example you would probably use the same kind-of-list as we use for the events. Then you could just generalize the eventList (for example call it thumbnailList) and an according model if necessary (see also TileModel -- we use the Tiles both for categories and extras)

BTW I just saw, that the excerpt of the news shouldn't be contained in a RemoteContent, since it's a plain string we get from cms. (Also it looks odd, that there are 3 different font-styles for each news-entry)

maxammann:
Yes you should generalize the input of components. Then map the models to to the input. In java you sometimes create a View of objects so they have the interface you need. There are also interfaces in flow: https://flow.org/en/docs/types/interfaces/

We should also think about using a framework for components: https://hackernoon.com/23-best-react-ui-component-libraries-and-frameworks-250a81b2ac42

Main reason is accessibility. Also could force us to follow a more strict guideline.

klinzo:
Yeah sure, I'll generalize the lists, was also my plan. I think it would be good to know what the problem though and why this pr is bad, especially the react native topic. @maxammann

maxammann:
I think we should remove DisclaimerPage and use a PageModel which fits the disclaimer API. So we have a structure which makes sense. 
The disclaimer api looks like this: https://cms.integreat-app.de/nuernberg/de/wp-json/extensions/v3/disclaimer

This would simplify the structure even more and would fit the models in the cms a bit more.

maxammann:
Apart from this I'm oke with merging :+1:

src/modules/app/StateType.js
```diff
@@ -6,7 +6,7 @@ import ExtraModel from '../endpoint/models/ExtraModel'
 import EventModel from '../endpoint/models/EventModel'
 import CityModel from '../endpoint/models/CityModel'
 import LanguageModel from '../endpoint/models/LanguageModel'
-import DisclaimerModel from '../endpoint/models/DisclaimerModel'
+import BasePageModel from '../endpoint/models/BasePageModel'
```
maxammann: I would just name it PageModel. No need for a Base imo. 
@Schedulaar what do you think?

klinzo: Would be okay for me.

Schedulaar: yup

src/routes/events/components/EventDetail.js
```diff
@@ -43,21 +43,18 @@ class EventDetail extends React.Component<PropsType> {
...
         <Thumbnail src={event.thumbnail || this.getEventPlaceholder()} />
         <Caption title={event.title} />
         <div>
-          <Identifier>{this.props.t('date')}: </Identifier>
-          <TimeSpan allDay={event.allDay}
-                    startDate={event.startDate}
-                    endDate={event.endDate}
-                    locale={this.props.language} />
+          <Identifier>{t('date')}: </Identifier>
+          <TimeSpan date={event.date} locale={this.props.language} />
```
Schedulaar: move the `this.props.language` to the object destructor above

src/modules/app/StateType.js
klinzo: @maxammann done.

PR #238, opened 2018-10-17
WEBAPP-375: Generalize event detail and page

klinzo:


klinzo:
This PR makes the `Page` redundant (at least for the WebApp), but since it is exported and probably also used (in the Wohnraumboerse frontend), I can't remove it. How should I deal with it?

src/modules/endpoint/models/DateModel.js
```diff
@@ -24,6 +24,37 @@ class DateModel {
   get allDay (): boolean {
     return this._allDay
   }
+
+  /**
+   * Returns a formatted string containing all relevant start and end date and time information
+   * @param {String} locale The locale to format the span in
+   * @return {String} The formatted span string
+   */
+  toFormattedString (locale: string): string {
```
klinzo: @maxammann I would leave this in the model if it's okay. I like it better like this and think it makes things a little bit easier.

src/modules/common/components/Page.js
```diff
@@ -1,24 +1,48 @@
...
+  object-fit: contain;
+`
 
 type PropsType = {|
   title: string,
+  thumbnail?: string,
   content: string,
-  onInternLinkClick: string => void,
-  hijackRegExp?: RegExp
+  lastUpdate: Moment,
```
Schedulaar: Please make lastUpdate optional - we might not want to show a lastUpdate on every Page (in the wohnraumbörse, we use the Page component and we probably don't want to show a last-update there)

klinzo: Done

src/modules/endpoint/models/DateModel.js
maxammann: Imo this is code to render a two timestamp and belongs next to a render method, but you can leave it here.

klinzo: @Schedulaar what do you think?

Schedulaar: I don't really mind 😅

PR #239, opened 2018-10-17
WEBAPP-381: Highlight Extras in wohnen and sprungbrett route

klinzo:


maxammann:
Not required but we could add tests for this check.

klinzo:
@maxammann done

PR #240, opened 2018-10-17
WEBAPP-268: Add aria labels to buttons

klinzo:


src/modules/layout/components/__tests__/__snapshots__/ToolbarItem.spec.js.snap
```diff
@@ -2,12 +2,13 @@
 
 exports[`ToolbarItem should render 1`] = `
 <StyledToolbarItem
-  data-tip="Click here!"
+  aria-lablel="Click here!"
```
Schedulaar: Typo

klinzo: Fixed, thanks :)

PR #242, opened 2018-10-24
WEBAPP-387: Fix internal linking

maxammann:
This url part was encoded twice. To fix this we first decode it when getting it from the DOM. Secondly we decode after the parsing with URL because it encodes internally.


maxammann:
@Schedulaar oh yeah, will do :)

PR #244, opened 2018-10-28
WEBAPP-214: Add poi functionality

klinzo:


src/routes/pois/containers/PoiPage.js
```diff
@@ -0,0 +1,92 @@
...
+        </>
+      } else {
+        const error = new ContentNotFoundError({type: 'poi', id: poiId, city, language})
+        return <FailureSwitcher error={error} />
+      }
+    }
+    return <>
+      <Helmet title={`${t('pageTitle')} - ${CityModel.findCityName(cities, city)}`} />
+      <Caption title={t('pois')} />
+      <List noItemsMessage={t('noPois')} items={pois} renderItem={this.renderPoiListItem} />
```
Schedulaar: Maybe we want to apply some sorting (alphabetical?) to the pois?

src/routes/pois/components/PoiListItem.js
```diff
@@ -0,0 +1,25 @@
...
+type PropsType = {|
+  poi: PoiModel
+|}
+
+class PoiListItem extends React.Component<PropsType> {
+  render () {
+    const {poi} = this.props
+    return (
+      <ListItem key={poi.path}
+                thumbnail={poi.thumbnail}
```
Schedulaar: How does it react, if some pois have a thumbnail and some don't?
Are the items unaligned then?

src/routes/pois/components/PoiListItem.js
```diff
@@ -0,0 +1,25 @@
...
+                title={poi.title}
+                path={poi.path}>
+        <div>{poi.location.location}</div>
```
Schedulaar: I think we should also add the excerpt here, as we do for the events. TBD in #feature-poi

src/routes/pois/containers/PoiPage.js
klinzo: Sounds reasonable :+1:

src/routes/pois/components/PoiListItem.js
klinzo: I don't know. Probably most pois are just autogenerated event locations which doesn't have an excerpt. We could make it optional of course. @maxammann what do you think?

maxammann: excerpt and description are optional for events. @svenseeberg is this correct?

src/modules/app/containers/__tests__/Switcher.spec.js
```diff
@@ -284,6 +307,14 @@ describe('Switcher', () => {
 
       expect(switcher).toMatchSnapshot()
     })
+
+    it('is the pois route', () => {
```
maxammann: should be the pois route

src/routes/pois/components/PoiListItem.js
klinzo: I'll just leave it for now, we can add it later if an excerpt is provided.

src/modules/common/components/Page.js
```diff
@@ -24,7 +24,7 @@ type PropsType = {|
   language: string,
   onInternalLinkClick: string => void,
   hijackRegExp?: RegExp,
-  children?: Array<React.Node>
+  children?: Array<React.Node> | React.Node
```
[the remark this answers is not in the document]
klinzo: Right :ok_hand:

src/modules/app/containers/__tests__/Switcher.spec.js
klinzo: Nope, this is good, the `describe()` says `should get the right page if data has been fetched and`.

PR #245, opened 2018-10-28
WEBAPP-380: Fix showing of times

klinzo:


src/modules/endpoint/endpoints/events.js
```diff
@@ -20,7 +20,7 @@ export default new EndpointBuilder(EVENTS_ENDPOINT_NAME)
   )
   .withMapper((json: Array<JsonEventType>): Array<EventModel> => json
     .map((event: JsonEventType) => {
```
Schedulaar: Do we need to adjust the JsonEventType?

klinzo: No, it was already right there.

PR #247, opened 2018-11-01
WEBAPP-391: Extract headroom to own repo

Schedulaar:


package.json
```diff
@@ -30,6 +30,7 @@
     "@fortawesome/free-regular-svg-icons": "^5.3.0",
     "@fortawesome/free-solid-svg-icons": "^5.3.0",
     "@fortawesome/react-fontawesome": "^0.1.2",
+    "@integreat-app/sticky-headroom": "^1.0.1",
```
maxammann: should we maybe rename this to just sticky-headroom? I probably would not use something that has a branding in the name.

Schedulaar: Hmm... I would prefer to have a branding...
1) Helps us getting attention in open-source-community
2) I personally do prefer using branded packages, because then I know, who developed the component (and maybe trust the people better). E.g. I like, that babel forces them now, so we know which packages are actually made by the babel team.

maxammann: For babel it makes sense because all packages are in one project.

We can keep it and see :) if we get no downloads we can think of removing it.

We should definitly include react- though: @integreat-app/react-sticky-headroom

Schedulaar: Yeah, we should include `react-`...
Should we also move the repo to the digitalfabrik github orga, since its not directly connected to integreat?  If so, should we move the npm package to a new digitalfabrik npm account?

maxammann: I think this can stay in integreat as this exists because of Integreat :)

PR #248, opened 2018-11-01
WEBAPP-267: Add StrictMode to App

Schedulaar:


src/modules/app/containers/App.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-import * as React from 'react'
+import React from 'react'
```
maxammann: I think we have to stick to `* as React` because of flow. I think something like `React.Node` does not work without.

PR #249, opened 2018-11-01
WEBAPP-279: introduce text disabled color

Schedulaar:

### DIFF
--- a/src/modules/layout/components/HeaderLanguageSelectorItem.js
+++ b/src/modules/layout/components/HeaderLanguageSelectorItem.js
@@ -10,8 +10,8 @@ import languageIcon from '../assets/language-icon.svg'
 import ReactTooltip from 'react-tooltip'
 import styled from 'styled-components'
 
-const InactiveImage = styled.img`
-  color: ${props => props.theme.colors.textSecondaryColor};
+const DisabledImage = styled.img`
+  color: ${props => props.theme.colors.textDisabledColor};
 `
 
 type PropsType = {|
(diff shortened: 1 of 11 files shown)


src/modules/theme/constants/colors.js
```diff
@@ -17,6 +18,7 @@ export const brightColors: ColorsType = {
   textColor: '#000000',
   textSecondaryColor: '#585858',
   textDecorationColor: '#c7c7c7',
+  textDisabledColor: '#dddddd',
```
klinzo: Really hard to read in the header imo, maybe it should be a little bit darker (also the contrast to the textColor is quite good). Maybe #d0d0d0?

src/modules/layout/components/HeaderLanguageSelectorItem.js
```diff
@@ -10,8 +10,8 @@ import languageIcon from '../assets/language-icon.svg'
 import ReactTooltip from 'react-tooltip'
 import styled from 'styled-components'
 
-const InactiveImage = styled.img`
-  color: ${props => props.theme.colors.textSecondaryColor};
+const DisabledImage = styled.img`
+  color: ${props => props.theme.colors.textDisabledColor};
```
klinzo: This doesn't have any effect since its an image (Probably code written by me though :smile: )

src/modules/theme/constants/colors.js
Schedulaar: done

src/modules/layout/components/HeaderLanguageSelectorItem.js
Schedulaar: done

klinzo: You just removed the code right? Could you create a task to grey the icon out?

PR #252, opened 2018-11-04
WEBAPP-349: Upgrade packages

Schedulaar:


Schedulaar:
Flow sucks...

src/modules/app/createReduxStore.js
```diff
@@ -20,12 +20,11 @@ export type ActionType<T> = { type: string, payload: Payload<T> }
...
 // todo: Change type to correct State type,
 // https://blog.callstack.io/type-checking-react-and-redux-thunk-with-flow-part-2-206ce5f6e705
-const createReduxStore = (createHistory: () => History, initialState: {} = {}, routesMap: RoutesMap = defaultRoutesMap): Store<any, any> => {
-  const history = createHistory()
-
-  const {reducer, middleware, enhancer} = connectRoutes(history, routesMap,
-    {onBeforeChange: onBeforeChange, querySerializer: queryString}
-  )
+const createReduxStore = (initialState: {} = {}, routesMap: RoutesMap = defaultRoutesMap): Store<any, any> => {
+  const { reducer, middleware, enhancer } = connectRoutes(routesMap, {
```
Schedulaar: redux-first-router now doesn't take a `history` as param anymore, but has one internally.
But scroll-behavior is not really good with the internal one: When you're already scrolled down and click on a link which leads to a long page, you already start "scrolled down" on the new page.

Schedulaar: Probably just not upgrade the router?

maxammann: Hm actually we implemented the scrolling behaviour. Thats nothing what the previous version did.

You can still use our custom version of history: https://github.com/faceyspacey/redux-first-router/blob/master/docs/migration.md

but maybe redux first router can also do the scrolling to the top.

maxammann: Here you go: https://github.com/faceyspacey/redux-first-router/blob/master/docs/connectRoutes.md#options

See scrollTop.

maxammann:
This is now ready for merge.

maxammann:
@klinzo @NKlug @Thomas125 ping

klinzo:
Still WIP :stuck_out_tongue: I'll have a look

src/modules/app/containers/__tests__/__snapshots__/Switcher.spec.js.snap
```diff
@@ -514,4 +514,4 @@ exports[`Switcher should return a failure if there was an error during fetching
 />
 `;
 
-exports[`Switcher should return a spinner if the data has not been fetched yet 1`] = `<LoadingSpinner />`;
+exports[`Switcher should return a spinner if the data has not been fetched yet 1`] = `<ForwardRef />`
```
klinzo: Where does this `ForwardRef` come  from?

src/modules/app/reducers/index.js
```diff
@@ -31,8 +31,17 @@ const endpoints = [
   poisEndpoint
 ]
 
-export const startFetchReducer = <T> (oldPayload: Payload<T>, action: ActionType<T>): Payload<T> => action.payload
-export const finishFetchReducer = <T> (oldPayload: Payload<T>, action: ActionType<T>): Payload<T> => {
+// fixme: WEBAPP-400 This type should be removed
+export type ReducerType<T> = <T> (oldPayload?: Payload<T>, action: ActionType<T>) => Payload<T>
```
klinzo: I think `oldPayload` should always be supplied, even if it is `null` or `undefined`, so `oldPayload: ?Payload<T>` would be better imo.

src/modules/common/components/FailureSwitcher.js
```diff
@@ -69,12 +70,12 @@ export class FailureSwitcher extends React.Component<PropsType> {
   }
 
   render () {
-    return <Fragment>
+    return <React.Fragment>
```
klinzo: You could replace this with `<>` and `</>`

src/modules/i18n/reducers/index.js
```diff
@@ -1,6 +1,15 @@
...
 
-import { handleAction } from 'redux-actions'
-import { setUiDirectionAction } from '../actions/setUIDirection'
+import { type ActionType, handleAction, type ReduxReducer } from 'redux-actions'
+import setUIDirection from '../actions/setUIDirection'
 
-export default handleAction(setUiDirectionAction, (state, action) => action.payload, 'ltr')
+export type DirectionStateType = 'ltr' | 'rtl'
+
+const uiDirectionReducer: ReduxReducer<DirectionStateType, { type: 'SET_UI_DIRECTION', payload: 'ltr' | 'rtl' }> =
```
klinzo: You could just use `SetUiDirectionActionType` instead of `{ type: 'SET_UI_DIRECTION', payload: 'ltr' | 'rtl' }`

src/modules/theme/actions/toggleDarkMode.js
```diff
@@ -2,6 +2,6 @@
 
 import { createAction } from 'redux-actions'
 
-export const toggleDarkModeAction = createAction<string, empty>('TOGGLE_DARK_MODE')
+export type ToggleDarkModeActionType = { type: 'TOGGLE_DARK_MODE' }
```
klinzo: Isn't there a `payload: boolean` missing?

src/modules/theme/reducers/index.js
```diff
@@ -1,6 +1,12 @@
 // @flow
 
-import { handleAction } from 'redux-actions'
-import { toggleDarkModeAction } from '../actions/toggleDarkMode'
+import { handleAction, type ReduxReducer } from 'redux-actions'
 
-export default handleAction(toggleDarkModeAction, state => !state, false)
+const toggleDarkModeReducer: ReduxReducer<boolean, { type: 'TOGGLE_DARK_MODE', payload: boolean }> =
```
klinzo: You could just use `ToggleDarkMode` (if `payload: boolean` is added)

src/routes/i18nRedirect/containers/I18nRedirectPage.js
```diff
@@ -11,31 +10,27 @@ import { goToCategories } from '../../../modules/app/routes/categories'
...
-  }
-
   getRedirectAction (): Action {
-    const {param, cities} = this.props
+    const {param, cities, i18n} = this.props
     if (!cities) {
       throw new Error('Payload not available')
     }
 
-    const i18n = this.context.i18n
```
klinzo: :+1:

maxammann:
woopy you are right :man_facepalming:

Schedulaar:
Can be merged.

PR #253, opened 2018-11-06
WEBAPP-393: Remove dependencies from endpoint package

maxammann:
The only depencies are now lodash, moment and mormalize-path

I also refacatored the feedback endpoint here.


src/modules/endpoint/endpoints/__tests__/feedback.spec.js
```diff
@@ -22,7 +22,15 @@ describe('FeedbackEndpoint', () => {
     formData.append('comment', 'comment')
     formData.append('query', 'query')
     formData.append('alias', 'alias')
-    expect(FeedbackEndpoint.mapParamsToFormData({
+
+    expect(feedback.mapParamsToBody).not.toBeNull()
+
+    // For flow inspection
+    if (!feedback.mapParamsToBody) {
+      return
```
Schedulaar: throw instead maybe... Hrgh flow should be able to recognize this -.-

src/modules/feedback/components/FeedbackBox.js
```diff
@@ -50,7 +50,7 @@ type PropsType = {|
 |}
 
 /**
- * Renders all necessary inputs for a Feedback and posts the data to the FeedbackEndpoint
+ * Renders all necessary inputs for a Feedback and posts the data to the feedback endpointjhkhji
```
Schedulaar: Ahh now I understandjhkhji

src/modules/feedback/components/FeedbackBoxContainer.js
```diff
@@ -162,13 +162,13 @@ export class FeedbackBoxContainer extends React.Component<PropsType, StateType>
   }
 
   /**
-   * Returns the data that should be posted to the FeedbackEndpoint
+   * Returns the data that should be posted to the feedback endpoint
    * @param selectedFeedbackOption
    * @param comment
    * @return {{feedbackType: string, isPositiveRating: boolean, comment: string, id: number, city: *, language: *,
```
Schedulaar: is return type correct?

src/modules/endpoint/types.js
```diff
@@ -78,3 +78,70 @@ export type JsonEventType = {
...
+export type JsonLanguageType = { code: string, native_name: string }
+
+export type JsonCityType = {
+  name: string,
+  path: string,
+  live: boolean,
+  events: boolean,
+  extras: boolean,
+  name_without_prefix: string
+}
```
Schedulaar: not a fan of this massive types file

src/routes/search/components/SearchFeedback.js
```diff
@@ -49,7 +49,7 @@ export class SearchFeedback extends React.Component<PropsType, StateType> {
   openFeedbackBox = () => {
     const {location, query} = this.props
     const {city, language} = location.payload
-    FeedbackEndpoint.postData({
+    feedback.fetchData({
```
Schedulaar: imo fetchData sounds like we're expecting to get data from the backend... maybe .request() would be a better name now?

src/modules/endpoint/types.js
maxammann: I think this is oke because this file should be generated automatically anyway. Or the cms team should provide this to us.

src/modules/feedback/components/FeedbackBoxContainer.js
maxammann: Removed them because flow provides them

src/routes/search/components/SearchFeedback.js
maxammann: request sounds good :+1:

src/modules/endpoint/EndpointBuilder.js
```diff
@@ -3,23 +3,25 @@
...
  * Helper class to build a {@link Endpoint}
  */
 class EndpointBuilder<P, T> {
-  _name: string
-  _paramsToUrlMapper: MapParamsToUrlType<P>
-  _mapper: MapResponseType<P, T>
-  _responseOverride: ?T
-  _errorOverride: ?Error
+  name: string
+  paramsToBodyMapper: MapParamsToBodyType<P>
```
klinzo: Shouldn't this type be optional?

src/modules/endpoint/endpoints/feedback.js
```diff
@@ -30,13 +30,12 @@ export type FeedbackDataType = {
   query?: string
 }
 
-class FeedbackEndpoint {
-  static mapParamsToUrl (params: FeedbackDataType): string {
+export default new EndpointBuilder<ParamsType, {}>(FEEDBACK_ENDPOINT_NAME)
+  .withParamsToUrlMapper((params): string => {
```
klinzo: Type annotation of params missing

src/modules/endpoint/types.js
klinzo: Imo also okay.

src/modules/feedback/components/FeedbackBoxContainer.js
```diff
@@ -4,8 +4,8 @@ import * as React from 'react'
 import 'react-dropdown/style.css'
 
 import CityModel from '../../../modules/endpoint/models/CityModel'
-import type { FeedbackDataType } from '../../../modules/endpoint/FeedbackEndpoint'
-import FeedbackEndpoint, {
+import type { ParamsType as FeedbackParamsType } from '../../../modules/endpoint/endpoints/feedback'
+import feedback, {
```
klinzo: I think naming this `feedbackEndpoint` makes sense, we also use this kind of terminology in the routes (e.g. `categoriesEndpoint` in the `categories` route), but you could also leave it like this of course. If you should change it, there is also an occurance in `SearchFeedback`

maxammann:
@klinzo oh missed your PR comments :/ Will fix them in an other PR

PR #254, opened 2018-11-07
WEBAPP-302: Readd icon above heading on category list

Thomas125:


src/routes/categories/components/CategoryList.js
```diff
@@ -28,9 +42,10 @@ type PropsType = {|
  */
 class CategoryList extends React.Component<PropsType> {
   render () {
-    const {categories, title, content, query, onInternLinkClick} = this.props
+    const {categories, title, thumbnail, content, query, onInternLinkClick} = this.props
     return (
       <div>
+        {thumbnail && <Centering><CategoryIcon src={thumbnail || iconPlaceholder} /></Centering>}
```
NKlug: As far as I can see an icon placeholder will basically never be shown because the `CategoryIcon` only renders when `thumbnail` exists?

src/routes/categories/components/CategoryList.js
```diff
@@ -7,19 +7,33 @@ import Caption from '../../../modules/common/components/Caption'
...
 import styled from 'styled-components'
+import iconPlaceholder from '../assets/IconPlaceholder.svg'
 
 const List = styled.div`
   & a {
     ${props => props.theme.helpers.removeLinkHighlighting}
   }
 `
 
+const Centering = styled.div`
```
maxammann: This can be added to CategoryIcon instead of Centering

Schedulaar: how?

maxammann: I meant the `text-align: center;` I think it can be moved to CateoryIcon.

Do we really need the text-align here? I mean there is no text in CategoryIcon.

Schedulaar: text-align centers all inline elements inside its container such as img

src/routes/categories/components/CategoryList.js
klinzo: I think we don't want to display a placeholder icon here. If there is no icon, we shouldn't display one.

maxammann: I agree, just hiding it would be sufficient.

PR #260, opened 2018-11-13
WEBAPP-348: Sanitize remote content

Schedulaar:


src/modules/endpoint/endpoints/__tests__/categories.spec.js
```diff
@@ -13,7 +13,7 @@ describe('categories', () => {
       id: 3650,
       path: '/augsburg/de/anlaufstellen',
       title: 'Anlaufstellen zu sonstigen Themen',
-      content: '',
+      content: '<a href="javascript:IWantToBeRemoved();">Ich bleib aber da.</a>',
```
maxammann: :rofl:

maxammann:
We need to sanitize all input is inserted directly into the dom. We probably also need to sanitize input which is used as href anywhere.
We could introduce a type like: SanitizedString. Our components could require you to sanitize it before usage.

But yeah there are probably a lot more places where we need to sanitize.

Schedulaar:
@klinzo Yes, that's true, but this task was about sanitizing html content that we inject into the dom.

PR #261, opened 2018-11-13
WEBAPP-394: Readd indentation

Schedulaar:


maxammann:
Yeah you are right, we discussed that we want to move it about 5px... Also the horizontal line should align with the text.

PR #267, opened 2018-11-18
WEBAPP-329: Adjust feedback box

michael-markl:


src/modules/feedback/components/FeedbackBox.js
```diff
@@ -37,6 +37,12 @@ export const SubmitButton = styled.span`
   cursor: pointer;
 `
 
+const ModifiedDropDown = props => <Dropdown {...props} controlClassName={props.className} />
+
+const StyledDropdown = styled(ModifiedDropDown)`
+  cursor: pointer !important;
```
maxammann: Is there still a pointer when you hover the dropdown? The problem was that the pointer would not disappear.

Schedulaar: There's a pointer now when you hover over the drop-down.
Not sure what we wanted it to behave. It sounds like a bug of react-dropdown what you're saying. Wouldn't fix that but I'll need to have another look.

maxammann: Maybe react-select is also a look worth: https://www.npmjs.com/package/react-select

But it has the same pointer behavior apparently, but is accessible

maxammann:
Can this get merged?

Schedulaar:
@maxammann Since the checks don't run: No... 🙄

klinzo:
I think you could merge this now.

PR #270, opened 2018-11-20
HOTFIX: WEBAPP-425: Fix ie11 pointer events

Schedulaar:

### DIFF
--- a/www/index.ejs
+++ b/www/index.ejs
@@ -73,6 +73,7 @@
 
       animation: show 1.5s forwards;
       pointer-events: none;
+      transition: height 0s 0.2s;
     }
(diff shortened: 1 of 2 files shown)


www/index.ejs
```diff
@@ -73,6 +73,7 @@
 
       animation: show 1.5s forwards;
       pointer-events: none;
+      transition: height 0s 0.2s;
```
maxammann: lolz

Schedulaar: haha yeah, IE11 cannot handle pointer-events: none; (only on IE11, Win7) although caniuse.com says it can :(

PR #272, opened 2018-11-25
WEBAPP-427: Remove unnecessary packages from packages.json

Schedulaar:


tools/webpack.config.js
```diff
@@ -40,11 +39,6 @@ const config = {
   stats: 'minimal',
   // The list of plugins for Webpack compiler
   plugins: [
```
maxammann: That way stylelint errors are not thrown while webpack is watching and building. Do we want this behavior?

Schedulaar: Well it never threw an error in debug mode since emitErrors was turned off anyways.
Also the plug-in is low on maintainance and since we don't particularly need it, I thought I'd just drop it.

Schedulaar: Should be enough to make stylelint a mandatory step in the jenkinsfile I think

PR #275, opened 2018-12-07
WEBAPP-426: Test route configs

klinzo:

### DIFF
--- a/src/modules/app/route-configs/SprungbrettRouteConfig.js
+++ b/src/modules/app/route-configs/SprungbrettRouteConfig.js
@@ -66,6 +66,9 @@ class SprungbrettRouteConfig implements RouteConfig<SprungbrettRouteParamsType, 
     this.getRoutePath({city: location.payload.city, language})
 
   getPageTitle = ({cityName, payloads}) => {
+    if (!cityName) {
+      return null
+    }
     const extras = payloads.extras.data
     const sprungbrettExtra = extras && extras.find(extra => extra.alias === SPRUNGBRETT_EXTRA)
     return sprungbrettExtra ? `${sprungbrettExtra.title} - ${cityName}` : ''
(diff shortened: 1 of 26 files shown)


src/modules/app/route-configs/SprungbrettRouteConfig.js
```diff
@@ -66,6 +66,9 @@ class SprungbrettRouteConfig implements RouteConfig<SprungbrettRouteParamsType,
     this.getRoutePath({city: location.payload.city, language})
 
   getPageTitle = ({cityName, payloads}) => {
+    if (!cityName) {
```
maxammann: Maybe we can change `cityName` to `cityName: string`. Then it would be not possible to pass a null object.

Looks a bit dangerous as a Page could end up having null as title.

klinzo: No, if the page title is null, no page title is rendered so the page title is still the default `Integreat`. I think @Schedulaar agrees since this was one of his requested changes in the last PR.

maxammann: I think we should not have a title Tag as long as it is not meaningful.
@Schedulaar what do you think? Or probably it does not really matter.

klinzo: I don't understand your point, what do you mean by a not meaningful title tag?

Schedulaar: I guess @maxammann wants to say, that we shouldn't have a default title tag (?). I think I agree with him, so we don't confuse search engines,  but I don't know if that has any impact

klinzo: Ahhh okay, yeah. Should I create a task for this? @Schedulaar @maxammann

maxammann: I'm not sure. Create one and we decide later. 

Probably search engines do not care but we should at least google a bit.

PR #278, opened 2018-12-28
WEBAPP-400: Fix type of store

klinzo:

### DIFF
--- a/src/modules/app/StoreActionType.js
+++ b/src/modules/app/StoreActionType.js
@@ -1,0 +1,19 @@
+// @flow
+
+import type { StartFetchActionType } from './actions/startFetchAction'
+import type { FinishFetchActionType } from './actions/finishFetchAction'
+import type { SetUiDirectionActionType } from '../i18n/actions/setUIDirection'
+
+export type SetLanguageActionType = { type: 'SET_LANGUAGE', payload: string }
+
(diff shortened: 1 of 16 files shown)


maxammann:
There are a few Dispatch usages which do not use the correct type: https://github.com/Integreat/integreat-webapp/blob/fa43c2e18f8fbfb134c1ce349a6b5b966859b7fe/src/modules/i18n/containers/I18nProvider.js#L129

src/modules/app/StoreActionType.js
```diff
@@ -0,0 +1,19 @@
...
+import type { SetUiDirectionActionType } from '../i18n/actions/setUIDirection'
+
+export type SetLanguageActionType = { type: 'SET_LANGUAGE', payload: string }
+
+export type SetCurrentCityActionType = { type: 'SET_CURRENT_CITY', payload: string }
+
+export type ToggleDarkModeActionType = { type: 'TOGGLE_DARK_MODE' }
+
+export type StoreActionType =
+  StartFetchActionType<*>
```
maxammann: Do we already have a Payload type which covers all our Payloads?

klinzo: You mean the data in the Payload? No currently not, should we introduce one?

PR #279, opened 2018-12-28
WEBAPP-435: Test RouteContentSwitcher

klinzo:

### DIFF
--- a/src/modules/app/containers/__tests__/__snapshots__/RouteContentSwitcher.spec.js.snap
+++ b/src/modules/app/containers/__tests__/__snapshots__/RouteContentSwitcher.spec.js.snap
@@ -1,0 +1,45 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`RouteContentSwitcher renderFailureLoadingComponents should render a FailureSwitcher if a payload contains an error 1`] = `
+<Translate(FailureSwitcher)
+  error={[Error: error]}
+/>
+`;
+
+exports[`RouteContentSwitcher renderFailureLoadingComponents should render a Spinner if data has not been fetched yet 1`] = `<ForwardRef />`;
+
+exports[`RouteContentSwitcher renderFailureLoadingComponents should render a Spinner if data has not been fetched yet 2`] = `<ForwardRef />`;
+
+exports[`RouteContentSwitcher should render and match snapshot 1`] = `
+<Connect(CategoriesPage)
+  categories={
+    CategoriesMapModel {
+      "_categories": Map {
+        "path01" => CategoryModel {
+          "_availableLanguages": Map {},
+          "_content": "contnentl",
+          "_id": 1,
+          "_lastUpdate": "2017-11-18T09:30:00.000Z",
+          "_order": 4,
+          "_parentPath": "parent/url",
+          "_path": "path01",
+          "_thumbnail": "thumb/nail",
+          "_title": "Title10",
+        },
+      },
+    }
+  }
+  cities={
+    Array [
+      CityModel {
+        "_code": "city1",
+        "_eventsEnabled": true,
+        "_extrasEnabled": false,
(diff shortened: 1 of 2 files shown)


src/modules/app/containers/__tests__/__snapshots__/RouteContentSwitcher.spec.js.snap
```diff
@@ -0,0 +1,45 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`RouteContentSwitcher renderFailureLoadingComponents should render a FailureSwitcher if a payload contains an error 1`] = `
+<Translate(FailureSwitcher)
+  error={[Error: error]}
+/>
+`;
+
+exports[`RouteContentSwitcher renderFailureLoadingComponents should render a Spinner if data has not been fetched yet 1`] = `<ForwardRef />`;
```
maxammann: A Spinner and not a ForwardRef should be rendered

src/modules/app/containers/__tests__/__snapshots__/RouteContentSwitcher.spec.js.snap
```diff
@@ -0,0 +1,45 @@
...
+
+exports[`RouteContentSwitcher renderFailureLoadingComponents should render a Spinner if data has not been fetched yet 2`] = `<ForwardRef />`;
```
maxammann: A Spinner and not a ForwardRef should be rendered

src/modules/app/containers/__tests__/__snapshots__/RouteContentSwitcher.spec.js.snap
```diff
@@ -0,0 +1,45 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`RouteContentSwitcher renderFailureLoadingComponents should render a FailureSwitcher if a payload contains an error 1`] = `
+<Translate(FailureSwitcher)
```
maxammann: Probably should dive() one layer deeper

src/modules/app/containers/__tests__/__snapshots__/RouteContentSwitcher.spec.js.snap
```diff
@@ -0,0 +1,45 @@
...
+  error={[Error: error]}
+/>
+`;
+
+exports[`RouteContentSwitcher renderFailureLoadingComponents should render a Spinner if data has not been fetched yet 1`] = `<ForwardRef />`;
+
+exports[`RouteContentSwitcher renderFailureLoadingComponents should render a Spinner if data has not been fetched yet 2`] = `<ForwardRef />`;
+
+exports[`RouteContentSwitcher should render and match snapshot 1`] = `
+<Connect(CategoriesPage)
```
maxammann: Maybe should render once deeper

src/modules/app/containers/__tests__/__snapshots__/RouteContentSwitcher.spec.js.snap
klinzo: Why? we want to render the FailureSwitcher component with the translate function, but we don't care how and what exactly is rendered. This test should test the RouteContentSwitcher and not the FailureSwitcher, so imo this is enough.

src/modules/app/containers/__tests__/__snapshots__/RouteContentSwitcher.spec.js.snap
klinzo: See comment regarding FailureSwitcher.

src/modules/app/containers/__tests__/__snapshots__/RouteContentSwitcher.spec.js.snap
klinzo: Done

PR #281, opened 2019-01-26
WEBAPP-410: Remove Question "Where are you" on city selection page

a-kud:

image

### DIFF
--- a/src/locales.json
+++ b/src/locales.json
@@ -209,8 +209,10 @@
       "searchCity": "البحث عن مكانك"
     },
     "de": {
-      "where": "Wo bist du?",
-      "searchCity": "Suche nach einem Ort"
+      "where": "",
+      "pageTitle": "Ortsauswahl",
+      "searchCity": "Ort suchen",
+      "metaDescription": "Finde lokale und aktuelle Informationen, Events und werde Teil deiner neuen Stadt. Integreat hilft dir dich mit anderen zu vernetzen und unterstützt mehrere Sprachen."
     },
     "en": {
       "where": "Where are you?",
--- a/src/locales.json
+++ b/src/locales.json
@@ -17,7 +17,7 @@
         "search": "Suche",
         "events": "News",
         "extras": "Extras",
-        "landing": "Wo bist du?",
+        "landing": "Ortsauswahl",
         "disclaimer": "Impressum und Kontakt",
         "mainDisclaimer": "Impressum und Kontakt - Integreat",
         "notFound": "Seite nicht gefunden"
@@ -210,7 +210,6 @@
     },
     "de": {
       "where": "",
-      "pageTitle": "Ortsauswahl",
       "searchCity": "Ort suchen",
       "metaDescription": "Finde lokale und aktuelle Informationen, Events und werde Teil deiner neuen Stadt. Integreat hilft dir dich mit anderen zu vernetzen und unterstützt mehrere Sprachen."
     },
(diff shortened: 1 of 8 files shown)


src/locales.json
```diff
@@ -209,8 +209,9 @@
       "searchCity": "البحث عن مكانك"
     },
     "de": {
-      "where": "Wo bist du?",
-      "searchCity": "Suche nach einem Ort"
+      "where": "",
```
maxammann: You can remove this in the code and remove this key completely.

src/locales.json
```diff
@@ -17,7 +17,7 @@
         "search": "Suche",
         "events": "News",
         "extras": "Extras",
-        "landing": "Wo bist du?",
+        "landing": "Ortsauswahl",
```
maxammann: The issue descrption was a bit unclear. Please also change this in the other languages. For languages like farsi or arabic you can omit the key currently.

maxammann: If you omit a key it will fallback automatically to english

src/routes/landing/components/Heading.js
```diff
@@ -3,9 +3,7 @@
 import React from 'react'
 
 import LocationBig from '../assets/LocationBig.svg'
-import type { TFunction } from 'react-i18next'
 import { withNamespaces } from 'react-i18next'
```
a-kud: Component should stay connected and this should remain here, right?

a-kud:
I'm not sure why some unrelated snapshots got updated.

src/routes/landing/components/Heading.js
maxammann: You can remove it as it is no longer used. Should be easy to add again afterwards.

src/modules/common/components/__tests__/__snapshots__/Page.spec.js.snap
```diff
@@ -19,7 +19,7 @@ exports[`Page should render 1`] = `
   />
   <Translate(LastUpdateInfo)
     language="en"
-    lastUpdate={"2016-01-07T09:36:24.000Z"}
+    lastUpdate={"2016-01-07T08:36:24.000Z"}
```
maxammann: We need to do something about this :) I'll create an other task for this. So you do no need to change anything here.

The reason why these changes were necessary *for you* is the difference in the timezone. We need to switch to UTC here.

src/locales.json
```diff
@@ -42,7 +42,7 @@
         "search": "جستجو",
         "events": "تازه​ها",
         "extras": "اضافات",
-        "landing": "تو کجایی؟",
+        "landing": "انتخاب محل",
```
maxammann: If you are not a native farsi speaker remove this for now :)

src/locales.json
```diff
@@ -54,7 +54,7 @@
         "search": "Rechercher",
         "events": "Actualités",
         "extras": "Extras",
-        "landing": "Tu es où?",
+        "landing": "Sélection de l'emplacement",
```
maxammann: If you are not a native french speaker remove this for now :)

maxammann:
Uuiui the reason why the build is failing is that our build server has a CET as timezone :dagger: 
Just revert the shapshot changes and the build server will build just fine. (The build will fail locally which is fine for now)

I'll do the task about the timezones now to fix this.

maxammann:
Feel free to merge :)

PR #282, opened 2019-02-13
WEBAPP-458: Cannot load events with Cyrillic characters in title

a-kud:

### DIFF
--- a/src/routes/events/containers/EventsPage.js
+++ b/src/routes/events/containers/EventsPage.js
@@ -36,7 +36,7 @@ export class EventsPage extends React.Component<PropsType> {
   render () {
     const {events, path, eventId, city, language, t} = this.props
     if (eventId) {
-      const event = events.find(_event => _event.path === path)
+      const event = events.find(_event => _event.path === decodeURIComponent(path))
 
       if (event) {
         return <>


src/routes/events/containers/EventsPage.js
```diff
@@ -36,7 +36,7 @@ export class EventsPage extends React.Component<PropsType> {
   render () {
     const {events, path, eventId, city, language, t} = this.props
     if (eventId) {
-      const event = events.find(_event => _event.path === path)
+      const event = events.find(_event => _event.path === decodeURIComponent(path))
```
Schedulaar: I just looked up, how we avoid this problem with normal categories.
The logic for that is located in the external 'integreat-api-client': https://github.com/Integreat/integreat-api-client/blob/master/src/models/CategoriesMapModel.js#L33

There we use `decodeURIComponent(normalizeUrl(path))`. Maybe we should do that here as well (to be consistent). Not sure what edge cases this covers though :sweat_smile:

maxammann: Thought about this aswell, but we actually have a task which fixes that: https://issues.integreat-app.de/browse/WEBAPP-452

:P

Schedulaar: Alright :+1:

PR #283, opened 2019-03-13
WEBAPP-441: Upgrade packages

Schedulaar:
I couldn't upgrade react-redux (https://issues.integreat-app.de/browse/WEBAPP-460).
I couldn't manage to fix tests for I18nProvider (https://issues.integreat-app.de/browse/WEBAPP-461).
I ignored some flow-files of redux-first-router, because they are broken with new flow-version. (https://issues.integreat-app.de/browse/WEBAPP-462)


src/polyfills.js
```diff
@@ -5,3 +5,5 @@ import '@babel/polyfill'
 import 'whatwg-fetch'
 
 import 'url-polyfill'
+
+window.regeneratorRuntime = require('regenerator-runtime') // todo: remove again
```
NKlug: Should this be removed before merging?

src/setupTest.js
```diff
@@ -3,6 +3,7 @@
 import 'raf/polyfill'
 import { configure } from 'enzyme'
 import Adapter from 'enzyme-adapter-react-16'
+window.regeneratorRuntime = require('regenerator-runtime') // todo remove
```
NKlug: Maybe also remove?

Schedulaar:
How do we continue here @maxammann ?

src/setupTest.js
Schedulaar: Was already removed.

src/polyfills.js
Schedulaar: Was already removed

src/__mocks__/react-i18next.js
```diff
@@ -4,15 +4,14 @@ import * as React from 'react'
 import i18next from 'i18next'
 
 // this mock makes sure any components using the translate HoC receive the t function as a prop
-export const withNamespaces = <Props> (namespace: string) => (Component: React.ComponentType<Props>) => {
-  const Translated = (props: Props) => <Component t={key => `${namespace}:${key}`} {...props} />
-  Translated.displayName = `Translate(${Component.displayName || Component.name})`
+const withTranslation = <Props> (namespace: string) => (Component: React.ComponentType<Props>) => {
+  const i18n = i18next.createInstance()
```
maxammann: Why do we need an i18next instance in here? I think it is only used in the tests and expected to be anything.

PR #284, opened 2019-06-15
WEBAPP-476: Make app themable

Schedulaar:


package.json
```diff
@@ -7,10 +7,10 @@
     "npm": ">=3.8"
   },
   "scripts": {
-    "start": "node tools/run.js",
-    "build": "node tools/build.js",
-    "build:debug": "node tools/build.js --debug",
-    "build:lib": "node tools/lib.build.js && flow-copy-source -i '**/@(__mocks__|__tests__)/**' ./src ./dist && mv ./dist/lib.js.flow ./dist/index.js.flow",
+    "start": "babel-node tools/run.js",
```
maxammann: babel-node should not be used in production: https://babeljs.io/docs/en/babel-node
 As this is only used in development this should be fine. I'd rather avoid it though if possible in any way.

tools/webpack.config.js
```diff
@@ -7,126 +7,132 @@ const getVersion = require('git-repo-version')
...
+    devtool: isDebug ? 'source-map' : false,
+    // What information should be printed to the console
+    stats: 'minimal',
+    // The list of plugins for Webpack compiler
+    plugins: [
+      new webpack.DefinePlugin({
+        'process.env.NODE_ENV': isDebug ? '"development"' : '"production"',
+        __DEV__: isDebug,
+        __VERSION__: JSON.stringify(getVersion()),
+        __CONFIG__: JSON.stringify(appConfig)
```
maxammann: Everytime you need a config parameter you use `__CONFIG__` right now in the code.
This means the same string (in this case `JSON.stringify(appConfig)`) is used over and over.

I think it would make sense to create a config.js which exports `__CONFIG__` once as default.

At least that is how I think that the DefinePlugin works.

maxammann:
Wait for the merge until we discussed how/where we want to deploy this :)

package.json
Schedulaar: Yeah, I wanted to avoid it as well, but then the flow-typed AppConfig was (for me) an argument for using it...

tools/webpack.config.js
Schedulaar: Done

maxammann:
@svenseeberg @Schedulaar If the requirements for the prototype are met we can merge this.

PR #289, opened 2019-08-23
WEBAPP-486: Fix language modal when title too long

Schedulaar:


Schedulaar:
Will merge this with 1 missing review as @klinzo is on vacation and @NKlug probably has no time...

PR #293, opened 2019-11-22
WEBAPP-468:  Change extras to angebote

maxammann:


maxammann:
Missing here:
* The extras in the url is still there
* Refactoring of code is missing
* There are still locales which use the term extra

PR #294, opened 2019-11-22
WEBAPP-449: Change web.integreat-app.de to integreat.app

a-kud:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -60,7 +60,7 @@ pipeline {
       steps {
         script {
           if (env.BRANCH_NAME == 'master') {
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -60,7 +60,7 @@ pipeline {
       steps {
         script {
           if (env.BRANCH_NAME == 'master') {
-            deploy("/web.integreat-app.de")
+            deploy("/integreat.app")
```
Schedulaar: Not sure, if this will work. @maxammann ?

maxammann: Yep!

PR #302, opened 2020-02-26
WEBAPP-408: Enable Search by Municipalities

a-kud:


src/routes/landing/components/CityEntry.js
```diff
@@ -31,11 +34,31 @@ type PropsType = {|
...
     const { city, language, filterText } = this.props
+    const getMunicipality = (city, filterText): Array<CityModel> => {
+      if (!filterText) {
+        return []
+      }
+      if (city._aliases) {
+        const municipalities = Object.keys(city._aliases).map(municipality => municipality.toLowerCase())
+        const municipalityMatch = municipalities.filter(municipality => municipality.includes(filterText.toLowerCase()))
+        const municipalityMatchCapitalized = municipalityMatch.map(match => match[0].toUpperCase() + match.slice(1))
+        return municipalityMatchCapitalized
```
klinzo: I'd prefer chaining the calls (with comments if unclear whats happening). Also, I think it is not guaranteed that municipalities are always containing only one word, so your approach on recapitalizing is probably not enough. I'd rather don't lowercase it in the first place, so you don't have to revert it later on.
```suggestion
        return Object.keys(city._aliases)
             .filter(municipality => municipality.toLowerCase().includes(filterText.toLowerCase()))
        return municipalityMatchCapitalized
```

src/routes/landing/components/CityEntry.js
```diff
@@ -31,11 +34,31 @@ type PropsType = {|
 class CityEntry extends React.PureComponent<PropsType> {
   render () {
     const { city, language, filterText } = this.props
+    const getMunicipality = (city, filterText): Array<CityModel> => {
+      if (!filterText) {
```
klinzo: You could check here for both `filterText` and `city._aliases`, so that only one if is needed.

src/routes/landing/components/CityEntry.js
```diff
@@ -31,11 +34,31 @@ type PropsType = {|
 class CityEntry extends React.PureComponent<PropsType> {
   render () {
     const { city, language, filterText } = this.props
+    const getMunicipality = (city, filterText): Array<CityModel> => {
```
klinzo: Either move this out of `render` and create a separate method (maybe even a `renderMunicipalities` or something which includes `{
          municipalities.map(municipality => (
            <MunicipalityItem key={municipality} search={filterText}>
              {municipality}
            </MunicipalityItem>
          ))
        }`) or inline it later at that occurance.

src/routes/landing/components/CitySelector.js
```diff
@@ -42,7 +43,7 @@ class CitySelector extends React.PureComponent<PropsType> {
         .filter(_city => {
           const isCityName = _city.name.toLowerCase().includes(filterText)
           const isAlias = _city._aliases && Object
-            .keys(_city._aliases).some(alias => alias.toLowerCase().includes(filterText))
+            .keys(_city._aliases).some(alias => normalize(alias).toLowerCase().includes(filterText))
```
klinzo: `toLowerCase` missing

src/routes/search/containers/__tests__/SearchPage.spec.js
```diff
@@ -88,7 +88,7 @@ describe('SearchPage', () => {
     const searchInputProps = tree.find('SearchInput').props()
 
     // the root category should not be returned
-    expect(searchPage.findCategories()).toHaveLength(categories.toArray().length - 1)
+    expect(searchPage.findCategories()).toHaveLength(categories.toArray().length)
```
klinzo: Why is this not the case anymore? At least the comment should be removed probably

src/routes/landing/components/CityEntry.js
```diff
@@ -32,21 +33,17 @@ type PropsType = {|
 |}
 
 class CityEntry extends React.PureComponent<PropsType> {
+  getMunicipality = (city, filterText): Array<CityModel> => {
+    if (city._aliases && filterText) {
+      return Object.keys(city._aliases)
+        .filter(municipality => normalize(municipality).toLowerCase().includes(filterText.toLowerCase()))
```
klinzo: I guess you have to normalize the filter text to. Otherwise searching for `Königsbrunn` does not work, since the municipality key gets normalized to `Konigsbrunn`

maxammann:
@a-kud do you have time currently to  work on this?

src/routes/landing/components/CitySelector.js
maxammann: Maybe this would be a good time to refactor this into a separate function and reuse it.

jest.config.json
```diff
@@ -10,7 +10,8 @@
     "\\.(css)$": "identity-obj-proxy"
   },
   "moduleFileExtensions": [
-    "js"
+    "js",
+    "json"
```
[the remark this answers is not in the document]
a-kud: some tests fail:
image

src/routes/landing/components/CitySelector.js
a-kud: let me create a separate ticket for it

src/modules/app/fetchData.js
```diff
@@ -45,10 +45,10 @@ async function fetchData<P, T> (
     return payload
   } catch (e) {
     let error
-    if (e instanceof LoadingError || e instanceof ParamMissingError || e instanceof MappingError) {
+    if (e instanceof ResponseError || e instanceof ParamMissingError || e instanceof MappingError) {
```
maxammann: Did we rename this error?

jest.config.json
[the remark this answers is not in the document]
maxammann: Ah alright this is a dependency of normalize-strings

A bit weird, but oke..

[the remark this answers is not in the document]
maxammann: normalize-strings seems not to be properly packaged

src/modules/app/fetchData.js
a-kud: According to this commit it was renamed to ResponseError

maxammann:
I think this is ready for merge

PR #303, opened 2020-03-09
WEBAPP-512: Fix (link) accessibility

klinzo:

Also fixes
- https://issues.integreat-app.de/browse/WEBAPP-397 apart from the fact that the dropDown is not closable with a press of `Enter` from outside the dropdown. I created an issue for this: https://issues.integreat-app.de/browse/WEBAPP-514
- https://issues.integreat-app.de/browse/WEBAPP-396 (only partly, everything else is still selectable). Will create a separate branch and PR for this.

### DIFF
--- a/src/modules/common/components/Tile.js
+++ b/src/modules/common/components/Tile.js
@@ -45,7 +45,7 @@ const TileTitle = styled.div`
 
 const TileContainer = styled(Col)`
   margin-bottom: 20px;
-  
+
   & > a,
   & button {
     display: block;
@@ -57,7 +57,7 @@ const TileContainer = styled(Col)`
     box-shadow: none;
     cursor: pointer;
   }
-  
+
   & > a:hover img,
   & button:hover img {
     transform: scale(1.01);
@@ -71,7 +71,7 @@ class Tile extends React.PureComponent<PropsType> {
   getTileContent (): React.Node {
     return <>
       <ThumbnailSizer>
-        <Thumbnail><img src={this.props.tile.thumbnail} /></Thumbnail>
+        <Thumbnail><img alt='' src={this.props.tile.thumbnail} /></Thumbnail>
       </ThumbnailSizer>
       <TileTitle>{this.props.tile.title}</TileTitle>
     </>
@@ -82,7 +82,7 @@ class Tile extends React.PureComponent<PropsType> {
     if (!tile.isExternalUrl) {
       return <CleanLink to={tile.path}>{this.getTileContent()}</CleanLink>
     } else if (!tile.postData) {
-      return <CleanAnchor href={tile.path}>{this.getTileContent()}</CleanAnchor>
+      return <CleanAnchor title={tile.title} href={tile.path}>{this.getTileContent()}</CleanAnchor>
     } else {
       const inputs = []
       tile.postData.forEach((value, key) => inputs.unshift(<input type='hidden' value={value} key={key} name={key} />))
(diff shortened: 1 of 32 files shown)


src/modules/layout/components/__tests__/HeaderDropDown.spec.js
```diff
@@ -62,10 +62,10 @@ describe('HeaderDropDown', () => {
     })
   })
 
-  it('should toggle when user clicks on img', () => {
+  it('should toggle when user clicks on button', () => {
     const component = wrapperComponent.dive()
     const instance: any = component.instance()
-    const onClick = component.find('img').prop('onClick')
+    const onClick = component.find({ selector: 'button' }).prop('onClick')
```
klinzo: Any better way to do this?

maxammann: I think this is fine

src/modules/layout/components/LocationFooter.js
```diff
@@ -21,8 +21,10 @@ export class LocationFooter extends React.PureComponent<PropsType> {
     const { t, city, language, onClick } = this.props
 
     return <Footer onClick={onClick}>
-      <CleanLink to={new DisclaimerRouteConfig().getRoutePath({ city, language })}>{t('imprintAndContact')}</CleanLink>
-      <CleanAnchor href='https://integreat-app.de/datenschutz/'>{t('privacy')}</CleanAnchor>
+      <CleanLink aria-label={t('imprintAndContact')} to={new DisclaimerRouteConfig().getRoutePath({ city, language })}>
```
Schedulaar: Also here, an `aria-label` is not necessary.

src/modules/layout/components/LocationFooter.js
```diff
@@ -21,8 +21,10 @@ export class LocationFooter extends React.PureComponent<PropsType> {
...
+        {t('imprintAndContact')}
+      </CleanLink>
+      <CleanAnchor title={t('privacy')} href='https://integreat-app.de/datenschutz/'>{t('privacy')}</CleanAnchor>
```
Schedulaar: A `title` is not appropriate here. Just leave it as is.

src/modules/layout/components/ToolbarItem.js
```diff
@@ -20,7 +20,7 @@ class ToolbarItem extends React.PureComponent<PropsType> {
   render () {
     const { href, text, icon } = this.props
     return (
-      <StyledToolbarItem href={href} aria-label={text}>
+      <StyledToolbarItem href={href} title={text}>
```
Schedulaar: `aria-label` was better than `title`. Title doesn't make sense here. (We already have a custom tooltip for that)

src/modules/common/components/Tile.js
```diff
@@ -82,7 +82,7 @@ class Tile extends React.PureComponent<PropsType> {
     if (!tile.isExternalUrl) {
       return <CleanLink to={tile.path}>{this.getTileContent()}</CleanLink>
     } else if (!tile.postData) {
-      return <CleanAnchor href={tile.path}>{this.getTileContent()}</CleanAnchor>
+      return <CleanAnchor title={tile.title} href={tile.path}>{this.getTileContent()}</CleanAnchor>
```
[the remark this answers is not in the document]
klinzo: Actually, I don't know anymore why I did what I did here. It did make sense to me tough. But yeah, I guess it's totally redundant.

PR #307, opened 2020-03-25
WEBAPP-517: Breadcrumbs display code

Schedulaar:


src/routes/categories/containers/__tests__/__snapshots__/CategoriesToolbar.spec.js.snap
```diff
@@ -29,7 +29,7 @@ exports[`CategoriesToolbar should render root-url for pdf endpoint 1`] = `
   openFeedbackModal={[Function]}
 >
   <ToolbarItem
-    href="https://cms.integreat-app.de/augsburg/de/wp-json/ig-mpdf/v1/pdf?url=%2Faugsburg%2Fde"
```
maxammann: Alright, for root categories there should be no parameter right? I just checked it and it seems correct.

Schedulaar: Correct.

PR #308, opened 2020-03-26
WEBAPP-504: Add category to feedback

klinzo:


src/modules/common/components/TextButton.js
```diff
@@ -0,0 +1,28 @@
...
+  background-color: ${props => props.theme.colors.themeColor};
+  border: none;
+  text-align: center;
+  border-radius: 0.25em;
+  cursor: pointer;
+`
+
+type ButtonPropsType = {
+  onClick: () => void,
+  text: string
```
maxammann: I think it would make sense to use the children prop here. Then you can place any node at as child.

`children: React.Node`

klinzo: My idea was to just build a TextButton. The styling would probably not go well with other components (e.g. yellow background, specific margins, centering). Therefore I'd leave it as is.

PR #310, opened 2020-03-26
WEBAPP-522 and WEBAPP-523: Inconsistent search and improved alias search

Schedulaar:

### DIFF
--- a/src/modules/common/utils/normalize.js
+++ b/src/modules/common/utils/normalize.js
@@ -1,7 +1,0 @@
-// @flow
-
-import normalizeStrings from 'normalize-strings'
-
-const normalize = (str: string) => normalizeStrings(str).toLowerCase()
-
-export default normalize
--- a/src/modules/common/utils/normalize.js
+++ b/src/modules/common/utils/normalize.js
@@ -1,0 +1,7 @@
+// @flow
+
+import normalizeStrings from 'normalize-strings'
+
+const normalize = (str: string) => normalizeStrings(str).toLowerCase()
+
+export default normalize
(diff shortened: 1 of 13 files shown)


src/routes/landing/components/CityEntry.js
```diff
@@ -23,40 +23,49 @@ const CityListItem = styled(Link)`
...
   }
 `
-const MunicipalityItem = styled(Highlighter)`
+
+const AliasItem = styled(Highlighter)`
+  display: inline-block;
   font-size: 12px;
+
+  &:not(:last-child):after {
+    content: ',\\00a0';
```
klinzo: Pls add a comment what this is used for. Instead, you could also use the same approach as in the native app, i.e. add the comma and whitespace directly to the text below. I'd prefer the latter, because imo its clearer.

src/routes/landing/components/CityEntry.js
```diff
@@ -23,40 +23,49 @@ const CityListItem = styled(Link)`
...
   city: CityModel,
   filterText: string
 |}
 
 class CityEntry extends React.PureComponent<PropsType> {
-  getMunicipality = (city: CityModel, filterText: string): Array<CityModel> => {
-    if (city._aliases && filterText) {
-      return Object.keys(city._aliases)
-        .filter(municipality => normalize(municipality).toLowerCase().includes(normalize(filterText).toLowerCase()))
+  getMatchedAliases = (city: CityModel, normalizedFilter: string): Array<CityModel> => {
```
klinzo: Why `matched`?

src/modules/common/utils/normalize.js
```diff
@@ -0,0 +1,7 @@
+// @flow
```
maxammann: Suggestion for name: `normalize-search-term/string`

PR #312, opened 2020-04-01
WEBAPP-525: bug fix - compile webapp on windows

Taggotty:

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,6 +1,6 @@
 const config = {
   presets: [
-    '@babel/preset-env',
+    ['@babel/preset-env', { "modules": "commonjs" }],
     '@babel/preset-react',
     '@babel/preset-flow'
   ],
--- a/yarn.lock
+++ b/yarn.lock
@@ -3,9 +3,9 @@
 
 
 "@babel/cli@^7.8.3":
-  version "7.8.3"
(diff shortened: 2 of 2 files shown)


.babelrc.js
```diff
@@ -1,6 +1,6 @@
 const config = {
   presets: [
-    '@babel/preset-env',
```
maxammann, line 3: What is the impact of this? is this acually a change or is the default different for linux vs windows?

klinzo: The solution was from here I think: https://github.com/webpack/webpack/issues/4039

Taggotty:
As I already mentioned last week, I had issues running the webapp the next day again, but due to another error. The website compiled without any issues, but I got `Uncaught TypeError: (0 , _typeof3.default) is not a function` somewhere in an webpacks dependency, commpletely unrelated to the project. So my guess was that an dependency was not correct. After some try and error it occured that upgrading _\@babel/babel-runtime_ and _\@babel/preset-env_ to 7.9.0 (current stable) resolved the errors and kind of fixed it. However I could not find any clear hint on why this error occured but [in this blogpost](https://swizec.com/blog/dont-worry-its-a-compiler-bug/swizec/6764) and [in this bug ticket ](https://github.com/babel/babel/issues/3711) from 2015. However it is mentioned that the error still occured.  Now I'd like to ask on how strictly do we manage dependency management?  Is updating those two dependenies alright or is it better to look for another solution? An if it is no issue to update those two dependencies it might als be considerable to update the complete babel scope to the current version?

PR #314, opened 2020-04-06
WEBAPP-467 added about to footer

Taggotty:

### DIFF
--- a/src/modules/layout/components/LocationFooter.js
+++ b/src/modules/layout/components/LocationFooter.js
@@ -24,8 +24,8 @@ export class LocationFooter extends React.PureComponent<PropsType> {
       <CleanLink to={new DisclaimerRouteConfig().getRoutePath({ city, language })}>
         {t('imprintAndContact')}
       </CleanLink>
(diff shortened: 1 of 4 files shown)


src/modules/layout/components/LocationFooter.js
```diff
@@ -24,9 +24,10 @@ export class LocationFooter extends React.PureComponent<PropsType> {
       <CleanLink to={new DisclaimerRouteConfig().getRoutePath({ city, language })}>
         {t('imprintAndContact')}
       </CleanLink>
+      <CleanAnchor href='https://integreat-app.de/about/'>{t('settings:about')}</CleanAnchor>
```
Taggotty, line 27: Look [here](https://react.i18next.com/latest/withtranslation-hoc#withtranslation-params) if you have questions to this syntax

klinzo, line 27: There is also an english version of the about page, https://integreat-app.de/en/about/ . If the content language is english, it should be displayed.

PR #315, opened 2020-04-12
Webapp 529 local news

amusameh:


src/modules/app/route-configs/NewsRouteConfig.js
```diff
@@ -0,0 +1,80 @@
...
+  LocalNewsModel,
+  Payload
+} from '@integreat-app/integreat-api-client'
+import fetchData from '../fetchData'
+import { cmsApiBaseUrl } from '../constants/urls'
+
+type NewsRouteParamsType = {| city: string, language: string |}
+type RequiredPayloadsType = {| news: Payload<Array<LocalNewsModel>> |}
+
+export const NEWS_ROUTE = 'NEWS'
```
mawadkittos, line 20: I suggest to move routes constant to separate file

src/routes/news/containers/NewsPage.js
```diff
@@ -0,0 +1,101 @@
...
+    return (
+      <>
+        <Tabs>
+        {/* TODO: update the locale file realted issues */}
+          <div label={t('LOCAL NEWS')} type={LOCAL_NEWS}>
+            <NewsList
+              noItemsMessage={t('currentlyNoEvents')}
+              isNewsList
+              items={news}
+              renderItem={this.renderNewsElement(language, LOCAL_NEWS)}
```
mawadkittos, line 72: I think its better if we use render prop here rather than this way, what do you think?
https://reactjs.org/docs/render-props.html

src/routes/news/containers/NewsPage.js
```diff
@@ -0,0 +1,101 @@
...
+  };
+
+  renderNewsElement = (language: string, type: string) => (newsItem: EventModel) => (
+    <NewsElement newsItem={newsItem} language={language} key={newsItem.path} type={type} path={this.props.path} />
+  )
+
+  render () {
+    const { news, path, newsId, city, language, t } = this.props
+
+    if (newsId) {
```
mawadkittos, line 46: I suggest if we move details component to separate route will be better, maybe different page so we keep render component neat an clean and make newsPage Component to show list only

What do you think?

src/routes/news/containers/NewsPage.js
```diff
@@ -0,0 +1,101 @@
...
+      const newsItem = news.find(_newsItem => {
+        return _newsItem.title === decodeURIComponent(newsId)
+      })
+
+      if (newsItem) {
+        const { title, message, timestap } = newsItem
+        return (
+          <>
+            <PushNewsDetails title={title} message={message} timestap={timestap} />
```
mawadkittos, line 55: We can change timestap here to timestamp

src/routes/news/components/PushNewsDetails.js
```diff
@@ -0,0 +1,56 @@
...
+    return (
+      <div
+        style={{
+          maxWidth: 660
+        }}
+        >
+        <StyledBanner>
+          Local News
+        </StyledBanner>
+
```
mawadkittos, line 48: No need for this line

src/routes/news/components/PushNewsDetails.js
```diff
@@ -0,0 +1,56 @@
...
+  render () {
+    const { title, message, timestap } = this.props
+    return (
+      <div
+        style={{
+          maxWidth: 660
+        }}
+        >
+        <StyledBanner>
+          Local News
```
mawadkittos, line 46: do we need to add this label to translation?

src/modules/app/route-configs/NewsRouteConfig.js
```diff
@@ -0,0 +1,80 @@
...
+    const { city, language } = state.location.payload
+
+    await Promise.all([
+      fetchData(createLocalNewsEndpoint(cmsApiBaseUrl), dispatch, state.news, { city, language }),
+      fetchData(createCitiesEndpoint(cmsApiBaseUrl), dispatch, state.cities),
+      fetchData(createEventsEndpoint(cmsApiBaseUrl), dispatch, state.events, { city, language }),
+      fetchData(createLanguagesEndpoint(cmsApiBaseUrl), dispatch, state.languages, { city, language })
+    ])
+  }
+}
```
jafkiitos, line 39: Let's add a new line to separate declarations

src/routes/news/components/PushNewsDetails.js
jafkiitos: Yes, we do.

src/routes/news/components/Tab.js
```diff
@@ -0,0 +1,80 @@
...
+import tuNewsLogoActive from '../assets/tu-news-active.png'
+import tuNewsLogoInActive from '../assets/tu-news-inactive.png'
+
+const TU_NEWS = 'tu'
+const decideColor = (activeTab, label, theme) => {
+  let color
+  if (activeTab.toLowerCase() !== 'news') {
+    color = theme.colors.themeColor
+  } else {
+    color = '#0279a6'
```
jafkiitos, line 14: Make color code a constant?

src/routes/news/components/Tab.js
```diff
@@ -0,0 +1,80 @@
...
+  render () {
+    const {
+      onClick,
+      props: { activeTab, label, type }
+    } = this
+
+    if (type === TU_NEWS) {
+      return (
+        <TuStyledTab onClick={onClick} activeTab={activeTab} label={label} />
+
```
jafkiitos, line 68: please remove extra line

src/modules/app/route-configs/NewsRouteConfig.js
maxammann: We have a weak convention to put constants in files `src/modules/app/contants.js` or  `src/modules/app/contants/index.js`

Not sure whether it makes sene there though :)

src/routes/news/containers/NewsPage.js
maxammann: Sounds good. This is required for the React-Native app aswell in order to allow navigating back.

src/routes/news/components/PushNewsDetails.js
maxammann: You can create a new i18n namespace called 'news'

src/routes/news/components/Tab.js
maxammann: Maybe it makes sense to put this into our ThemeProvider and the Theme definition.

src/routes/news/containers/NewsPage.js
amusameh: I think we should leave it like this to be consistent with rest of existing routes.

src/modules/app/route-configs/NewsRouteConfig.js
klinzo: This is fine as it is, since it is consistent with existing route configs. We may have to refactor it at some point though.

src/routes/news/containers/NewsPage.js
klinzo: Moving the details to separate routes sounds good. The reason why we don't do this for most of the other routes is that we don't use single endpoints there but instead just load everything with one request. An example for separation of list and detail components are the `extras` route and its detail routes `sprungbrett` and `wohnen`.

src/modules/layout/containers/LocationHeader.js
```diff
@@ -57,12 +60,13 @@ export class LocationHeader extends React.Component<PropsType> {
...
 
   getNavigationItems (): Array<Element<typeof HeaderNavigationItem>> {
-    const { t, isEventsEnabled, isExtrasEnabled, location, events } = this.props
+    const { t, isEventsEnabled, isNewsEnabled, isExtrasEnabled, location, events } = this.props
     const { city, language } = location.payload
     const currentRoute = location.type
 
     const isEventsActive = events ? events.length > 0 : false
     const isCategoriesEnabled = isExtrasEnabled || isEventsEnabled
+    const isNewsActive = true
```
Schedulaar, line 69: Probably this should only be true, if there are news available (just like with events)

amusameh: My idea was to fetch the tunews on the button click and assumed that there will be always data so no need to disable it if there is no push notification.
What do you think?

src/routes/news/components/PushNewsDetails.js
amusameh: I created the news name space but for this to work as I understand it needs the related translations values in the `locales.json`.

src/modules/app/reducers/index.js
```diff
@@ -58,7 +60,7 @@ export const finishFetchReducer = <T: PayloadDataType> (oldPayload?: Payload<T>,
 const defaultState = new Payload(false)
 
 type ReducerType = Reducer<StateType, StartFetchActionType<PayloadDataType> | FinishFetchActionType<PayloadDataType>>
-const reducers: { [actionName: string]: ReducerType } = endpointNames.reduce(
+const reducers: {[actionName: string]: ReducerType } = endpointNames.reduce(
```
jafkiitos, line 63: Let's keep formatting as is

src/modules/theme/constants/colors.js
```diff
@@ -8,7 +8,8 @@ export type ColorsType = {
   textSecondaryColor: string,
   textDecorationColor: string,
   textDisabledColor: string,
-  backgroundColor: string
+  backgroundColor: string,
+  tuNewsColor: string
```
jafkiitos, line 12: I feel that tuNewsColor should not exist here. 

This file is more for general 'theme' colors for the whole app, whereas Tu News is a very specific element that exists in one or two places only.

What does everybody think?

src/routes/news/components/PushNewsDetails.js
klinzo: There is a fallback to german if values in the current language are not available, so it should work nevertheless.

src/modules/layout/containers/LocationHeader.js
klinzo: Yeah, I'd agree. There are always TüNews (at least in german), so I guess it's okay to always show it.

PR #317, opened 2020-04-15
WEBAPP-450: Make text required for negative feedback

Taggotty:


src/modules/feedback/components/__tests__/__snapshots__/FeedbackBox.spec.js.snap
```diff
@@ -39,6 +39,7 @@ exports[`FeedbackBox should match snapshot 1`] = `
     onCommentChanged={[Function]}
   />
   <TextButton
+    disabled={true}
```
Taggotty, line 42: should I add test cases? e.g. comment not empty or positiveRatingSelected?

Taggotty:
image
we should probably add some kind of indication why you cannot submit an empty feedback. Maybe append "\*" to the question "Was können wir besser machen?". Opinions?

src/modules/feedback/components/__tests__/__snapshots__/FeedbackBox.spec.js.snap
klinzo: If you feel like it makes sense and you'd like to do it, sure, gladly.

klinzo:
Yes, we should definitely find a way to indicate it, a star is probably a good solution.
Another possible option would probably be to automatically focus the text field and show a red indicator if the focus is changed. Or leave the button clickable and show a red indicator if it is clicked and no text is entered.
You should probably take a look at how different pages are handling similar stuff and choose the option that suites you the most.

src/modules/common/components/TextButton.js
```diff
@@ -15,13 +15,14 @@ export const StyledButton = styled.button`
 
 type ButtonPropsType = {
   onClick: () => void,
-  text: string
+  text: string,
+  disabled: boolean
```
klinzo, line 19: There should be other usages of the `TextButton`. So you probably either have to make the prop optional or supply the prop there too. Otherwise flow should show errors. Do you already have flow configured?

Taggotty:
So now I added just a red star to the question, as changing the border of the comment field feels like we'd have to juggle the comment state around too much. If you still like that solution better we should probably consider using form fields or sth.

src/modules/feedback/components/__tests__/__snapshots__/FeedbackBox.spec.js.snap
Taggotty: so for now I will leave it as it is. Iwas just a little curious on your testing policy 😁

klinzo:
I guess this is fine, I like it :+1: Did you check how it looks with RTL languages?
I don't know if it makes sense to add the star after `Wofür ist das Feedback?` since the category is also required, although an option is already selected by default.

We should probably also think about whether we need a hint in the text are.

src/modules/feedback/components/FeedbackComment.js
```diff
@@ -8,18 +8,24 @@ const CommentField = styled.textarea`
...
 |}
 
 class FeedbackComment extends React.PureComponent<PropsType> {
   render () {
-    const { commentMessage, comment, onCommentChanged } = this.props
+    const { commentMessage, comment, onCommentChanged, required } = this.props
     return (
       <>
-        <Description>{commentMessage}</Description>
+        <Description>{commentMessage}{required ? <RequiredText>*</RequiredText> : ''}</Description>
```
klinzo, line 28: You could change this to 
```suggestion
        <Description>{commentMessage}{required && <RequiredText>*</RequiredText>'}</Description>
```

src/modules/feedback/components/__tests__/__snapshots__/FeedbackBox.spec.js.snap
klinzo: `Uncaught ReferenceError: 'TestingPolicy' is not defined.` :grin:

maxammann:
Make sure to name your PRs: WEBAPP-450: ----

It makes it easier to copy the reference to Jira :)

maxammann:
I think it actually could make sense to record downvotes. For example if one page gets many downvotes then it is maybe an empty page and nobody bothered to write a comment.

klinzo:
@maxammann actually I'm quite sure we discussed this at a conference (or even multiple times) and agreed on making the text required for negative feedback. We had the same arguments, but in the end agreed on this. And now I am also convinced this is right. Negative feedback without a comment could mean anything and does not provide any benefit in the most cases.

Taggotty:
> I don't know if it makes sense to add the star after `Wofür ist das Feedback?` since the category is also required, although an option is already selected by default.

I don't think this is necessary, as you would automatically open the dropdown to see what other options you have. Otherwise we would have to deal with checking for no selection.
 
> We should probably also think about whether we need a hint in the text area.
I looked at a few other solutions to submit feedback like Jira, OneDrive/Word (where you can submit empty feedback o.o), gitkraken, sourceTree (where every few versions the application crashes after submitting feedback), youtube (where you can't even write text), ... 
And I liked Jiras solution the best.

maxammann:
:+1:

klinzo:
@maxammann you currently just don't want to approve, do you? :stuck_out_tongue:

src/modules/common/components/TextButton.js
```diff
@@ -15,13 +15,14 @@ export const StyledButton = styled.button`
 
 type ButtonPropsType = {
```
lulo4907, line 16: I'd suggest to handle the cursor prop too if the button is disabled: `cursor: string` with style `cursor: ${props => props.disabled ? 'default' : 'pointer'};`

maxammann:
Apart from @lulo4907 comment and the failing test :+1:

PR #318, opened 2020-04-15
WEBAPP-538: Prevent feedback for dashboard

klinzo:


Taggotty:
Why don't we want any feedback for the dashboard?

klinzo:
The description is probably not 100% correct. The dashboard represents the city, and since we already have a feedback option for all the content in the city, a separate feedback option for the dashboard is not necessary.

image

maxammann:
Tests are failing

PR #320, opened 2020-04-20
WEBAPP-549: Add design proposals

klinzo:


klinzo:
@Taggotty thank you, but I'm not the designer (sadly :smile: ). This are just design proposals by various designers of the last two years. I just collected them and added them to the repo for (hopefully) future orientation. Regarding the space in the header, you're definitely right though.

maxammann:
Shouldn't his be a task for the webapp?

PR #321, opened 2020-04-21
Webapp 436 fix infinite loading spinner

Taggotty:

### DIFF
--- a/src/modules/app/containers/RouteContentSwitcher.js
+++ b/src/modules/app/containers/RouteContentSwitcher.js
@@ -32,12 +32,12 @@ class RouteContentSwitcher extends React.PureComponent<PropsType> {
 
   isWaitingForExtra = (extraType: 'offers' | 'sprungbrettJobs', payloads: {[string]: Payload<any>}): boolean => {
     const alias = extraType === 'offers' ? WOHNEN_EXTRA : SPRUNGBRETT_EXTRA
-    return payloads[extraType] && find(payloads.extras.data, extra => extra.alias === alias)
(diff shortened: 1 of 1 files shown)


src/modules/app/containers/RouteContentSwitcher.js
```diff
@@ -27,12 +29,24 @@ class RouteContentSwitcher extends React.PureComponent<PropsType> {
     return null
   }
 
+  renderExtraFailure = (payloads: {[string]: Payload<any>}): React.Node => {
+    if (payloads.extras && !payloads.extras.isFetching && payloads.extras.data && !payloads.extras.error) {
+      if (payloads.offers && !find(payloads.extras.data, extra => extra.alias === WOHNEN_EXTRA)) {
+        return <FailureSwitcher error={new Error('The Wohnen extra is not supported.')} />
```
klinzo, line 35: We should use a `ContentNotFoundError` here. The `FailureSwitcher` should then render a localized message.

src/modules/app/containers/RouteContentSwitcher.js
```diff
@@ -27,12 +29,24 @@ class RouteContentSwitcher extends React.PureComponent<PropsType> {
...
+      } else if (payloads.sprungbrettJobs && !find(payloads.extras.data, extra => extra.alias === SPRUNGBRETT_EXTRA)) {
+        return <FailureSwitcher error={new Error('The Sprunbrett extra is not supported.')} />
```
klinzo, line 37: Same as above.

maxammann:
Tested in Firefox :+1:

PR #323, opened 2020-04-27
WEBAPP-547: Use test cms for dev builds

klinzo:


src/modules/app/fetchData.js
```diff
@@ -28,19 +26,7 @@ async function fetchData<P, T> (
...
-    }
-    if (responseOverride) {
-      const data = endpoint.mapResponse(responseOverride, params)
-      const payload = new Payload(false, formattedUrl, data, null)
-      dispatch(finishFetchAction(endpoint.stateName, payload))
-      return payload
-    }
-
-    const payload = await endpoint.request(params, formattedUrl)
+    const payload = await endpoint.request(params)
```
maxammann, line 29: Whouls we create a task to remove the errorOverride functionality?

maxammann: I think it is  not used in the RN app. Not sure though.

klinzo: This is all a duplicate of code present in the api-client. Therefore I thought I'd just remove it.

klinzo: It is used for some tests, so I don't think we should remove the functionality entirely.

package.json
```diff
@@ -7,9 +7,9 @@
     "npm": ">=3.8"
   },
   "scripts": {
-    "start": "webpack-dev-server --config tools/webpack.config.babel.js --progress --env.prod=false --env.config_name=integreat",
+    "start": "webpack-dev-server --config tools/webpack.config.babel.js --progress --env.prod=false --env.config_name=integreat-debug",
```
Taggotty, line 10: So the regular start command disappears. Is it not used anywhere?

klinzo: No, this is just a debug command starting the local webpack server. Also, if you want to use the live cms with the start command, you can still set the url manually back to `cms.integreat-app.de`

tools/integreat-debug-config/index.js
```diff
@@ -0,0 +1,17 @@
...
+import brightTheme from '../../src/modules/theme/constants/theme'
+
+const IntegreatAppDebugConfig: AppConfigType = {
+  appTitle: 'Integreat',
+  itunesAppId: '1072353915',
+  theme: brightTheme,
+  cmsUrl: 'https://cms-test.integreat-app.de',
+  locationIcon: '/location-big.svg',
+  logoWide: '/integreat-app-logo.png',
+  internalLinksHijackPattern: 'https?:\\/\\/(cms\\.integreat-app\\.de|web\\.integreat-app\\.de|integreat\\.app)(?!\\/[^/]*\\/(wp-content|wp-admin|wp-json)\\/.*).*'
```
Schedulaar, line 14: I would suggest to adjust this to
```
internalLinksHijackPattern: 'https?:\\/\\/(cms-test\\.integreat-app\\.de|web\\.integreat-app\\.de|integreat\\.app)(?!\\/[^/]*\\/(wp-content|wp-admin|wp-json)\\/.*).*'
```

PR #324, opened 2020-04-28
WEBAPP-521: Make feedback dropdown accessible

klinzo:


src/modules/common/components/Dropdown.js
```diff
@@ -0,0 +1,102 @@
...
+
+const StyledMenuItem = styled(MenuItem)`
+  display: flex;
+  padding: 8px;
+  color: ${props => props.theme.colors.textSecondaryColor};
+  cursor: pointer;
+
+  :focus,
+  :hover {
+    background-color: ${props => props.theme.colors.textDecorationColor};
```
klinzo, line 60: Not exactly sure what color to use here.

klinzo:
image
image

src/modules/common/components/Dropdown.js
```diff
@@ -0,0 +1,102 @@
...
+  position: absolute;
+  top: 100%;
+  z-index: 1;
+  display: flex;
+  width: 100%;
+  box-sizing: border-box;
+  flex-direction: column;
+  margin: 8px 0;
+  border-radius: 4px;
+  background-color: ${props => props.theme.colors.backgroundColor};
```
maxammann, line 47: Minor: could be refactored to:

`(props: { theme: ThemeType }) => theme.colors.backgroundColor` 

That way flow types work I think, not sure though. Maybe it also works that way.

src/modules/common/components/Dropdown.js
klinzo: Black text color is fine, but the difference between #fafafa and white is nearly not visible here :/
image

src/modules/common/components/Dropdown.js
klinzo: We don't have this anywhere yet. But I think maybe we have to take a look at flow types for styled components at some time. I'd leave this as it is for now though (only because of consistency of course :stuck_out_tongue: )

maxammann:
@klinzo also yarn.lock seems out of date :)

klinzo:
I'd also say this is fine. I think too much work for too little benefit. Put feel free to open an issue.

src/modules/common/components/Dropdown.js
Taggotty: Maybe try #EDEDED. This should be the equivalent of the default color without an saturation, I think.

klinzo: @Taggotty we have some colors defined in our theme and should stick to them if possible. Therefore I would try not to introduce a new color here. What do you think about the color in the screenshot here https://user-images.githubusercontent.com/29095487/80475778-b0336e80-8949-11ea-8417-2fc35eaffbe0.png

Taggotty: Is the font black? It looks kind of lighter than the one above. I think textDecorationColor and black text should be fine.

Taggotty: Anyway I think the color is fine. It just doesn't really seem to pop.

klinzo: Now it should be black, yes. The screenshot below was before the change of the text color.

PR #331, opened 2020-05-06
WEBAPP-461: Fix i18n provider tests

Taggotty:


src/modules/i18n/containers/__tests__/I18nProvider.spec.js
```diff
@@ -149,32 +153,21 @@ describe.skip('I18nProvider', () => {
...
         language: expectedLanguage,
+        i18nLoaded: true,
         fonts: { lateef: true, openSans: true, raleway: true }
       })
       // $FlowFixMe
       I18nProvider.getSelectedFonts = originalGetSelectedFonts
     })
   })
-
-  it('should add direction style depending on language', () => {
```
Taggotty, line 169: The test below runs this test succesfully and shows perfectly why setUiDirection is quite difficult to test. Maybe we should consider moving setUiDirection and setLanguage from componentDidUpdate to ComponentWillUpdate, but I am sure there was some intention behind componentDidUpdate.
```javascript
it('should add direction style depending on language', async () => {
  const mockSetUiDirection = jest.fn(setUIDirection)
  const component = mount(<I18nProvider language='en' setUiDirection={mockSetUiDirection}>
    <div />
  </I18nProvider>)
  expect(component.find('div').at(0).prop('style').direction).toEqual('ltr')
  component.setProps({ language: 'ar' })
  component.update()
  await new Promise(resolve => setTimeout(resolve, 200))
  component.update()
  expect(mockSetUiDirection).toHaveBeenCalledWith('rtl')
  expect(component.find('div').at(0).prop('style').direction).toEqual('rtl')
})
```

maxammann: And without the `setTimeout` it does not work?

maxammann: componentWillUpdate is deprecated unfortunately: https://reactjs.org/docs/react-component.html#unsafe_componentwillupdate

src/modules/i18n/containers/I18nProvider.js
```diff
@@ -81,7 +81,7 @@ export class I18nProvider extends React.Component<PropsType, StateType> {
     )
   }
 
-  setLanguage (language: ?string) {
+  async setLanguage (language: ?string) {
```
klinzo, line 84: Right now `setLanguage` is not really async since you don't `await` the promise, is it?

src/modules/i18n/containers/__tests__/I18nProvider.spec.js
```diff
@@ -121,23 +127,24 @@ describe.skip('I18nProvider', () => {
       const originalGetSelectedFonts = I18nProvider.getSelectedFonts
       // $FlowFixMe
       I18nProvider.getSelectedFonts = jest.fn(I18nProvider.getSelectedFonts)
-      i18n.changeLanguage = jest.fn()
+      i18n.changeLanguage = jest.fn(i18next.changeLanguage)
```
klinzo, line 130: Do we have to use `i18next.changeLanguage` here? Couldn't we use a manual defined mock function or `jest.fn(i18n.changeLanguage)`?

src/modules/i18n/containers/__tests__/I18nProvider.spec.js
klinzo: It would also be possible to use `wait-for-expect` (https://github.com/TheBrainFamily/wait-for-expect) to omit using a error prone timeout.

src/modules/i18n/containers/I18nProvider.js
klinzo: It would probably be better code to just  use `async/await` in this case here, since the then callback doesn't even use the value returned by `this.i18n.changeLanguage`

src/modules/i18n/containers/__tests__/I18nProvider.spec.js
Taggotty: > componentWillUpdate is deprecated unfortunately: https://reactjs.org/docs/react-component.html#unsafe_componentwillupdate

did not know that. Is there another way to avoid calling update twice here? 

> It would also be possible to use `wait-for-expect` (https://github.com/TheBrainFamily/wait-for-expect) to omit using a error prone timeout.

I will try that.

klinzo: In fact I'm not sure why we need `update` anyway... according to the enzyme docs `setProps` should already force a re-render: https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/setProps.html
The second `update` should also be not needed AFAIK since in theory it should automatically rerender after `setLanguage`/`setUiDirection` is called :thinking:

Taggotty: Somehow the issue with the timimg was related to changeLanguage in itself. The state update did not trigger the ui-direction to change. I am actually not quite sure what the issue was, probably something with the I18NextProvider to be unmounted or sth, but I am not sure. However it works now, as I moved the code from then(..) into the callback of changeLanguage.

src/modules/i18n/containers/__tests__/I18nProvider.spec.js
```diff
@@ -121,16 +127,17 @@ describe.skip('I18nProvider', () => {
...
       // $FlowFixMe
       I18nProvider.getSelectedFonts = jest.fn(I18nProvider.getSelectedFonts)
-      i18n.changeLanguage = jest.fn()
+      i18n.changeLanguage = jest.fn(i18n.changeLanguage)
 
       component.instance().setLanguage()
 
       // $FlowFixMe
       expect(document.documentElement.lang).toEqual(expectedLanguage)
-      expect(i18n.changeLanguage).toHaveBeenCalledWith(expectedLanguage)
```
klinzo, line 130: Rather use `expect.any(Function)`

src/modules/i18n/containers/__tests__/I18nProvider.spec.js
klinzo: The other update is still needed though?

src/modules/i18n/containers/__tests__/I18nProvider.spec.js
```diff
@@ -169,8 +177,8 @@ describe.skip('I18nProvider', () => {
   it('should add direction style depending on language', () => {
     const mockSetUiDirection = jest.fn()
     const component = mount(<I18nProvider language='en' setUiDirection={mockSetUiDirection}>
-      <div />
```
klinzo, line 172: Indentation is strange

src/modules/i18n/containers/__tests__/I18nProvider.spec.js
Taggotty: nope.

klinzo: But it's still there :thinking:

Taggotty: wait. what exactly do you mean? I thought you meant the `component.update()` in line 183. And there you do need to call update once otherwise `componentDidUpdate` will not be triggered

klinzo: Yes, but you wrote `nope`, its not needed anymore :grin:

Taggotty: Ah. I thought you ment the second update in the code snippet above. To clarify: We do need to call update at least once.

PR #334, opened 2020-05-11
Handle error with user hint, if feedback endpoint is not available.

lulo4907:

### DIFF
--- a/src/modules/feedback/components/FeedbackBox.js
+++ b/src/modules/feedback/components/FeedbackBox.js
@@ -36,20 +36,24 @@ type PropsType = {|
   comment: string,
   onCommentChanged: SyntheticInputEvent<HTMLTextAreaElement> => void,
   onFeedbackOptionChanged: FeedbackVariant => void,
-  onSubmit: (value?: boolean) => void,
+  onSubmit: (value?: string) => void,
   t: TFunction,
   closeFeedbackModal: () => void,
   theme: ThemeType,
-  requestError: boolean
+  feedbackSent: string
 |}
 
 /**
  * Renders all necessary inputs for a Feedback and posts the data to the feedback endpoint
  */
 export class FeedbackBox extends React.PureComponent<PropsType> {
+  constructor (props: PropsType) {
+    super(props)
+    this.handleClick = this.handleClick.bind(this)
+  }
 
   handleClick = () => {
-    this.props.requestError !== false ? '' : this.props.onSubmit()
+    this.props.feedbackSent !== 'ERROR' ? this.props.onSubmit('SUCCESS') : this.props.onSubmit('ERROR')
   }
 
   render () {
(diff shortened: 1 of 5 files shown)


src/modules/feedback/components/FeedbackBoxContainer.js
```diff
@@ -72,7 +73,8 @@ export class FeedbackBoxContainer extends React.Component<PropsType, StateType>
       if (postFeedbackDataOverride) {
         postFeedbackDataOverride(feedbackData)
       } else {
-        await createFeedbackEndpoint(cmsApiBaseUrl).request(feedbackData)
+        const res = await createFeedbackEndpoint(cmsApiBaseUrl).request(feedbackData)
+        if (res) { return res }
```
klinzo, line 77: No need to check for the value of the result of the request, it always returns `{}` for the feedback endpoint if it does not fail. Just return `true` if you don't have to catch an error and `false` if you have to. You should also handle `postFeedbackDataOverride` above.

src/modules/feedback/components/FeedbackBox.js
```diff
@@ -73,9 +75,10 @@ export class FeedbackBox extends React.PureComponent<PropsType> {
           commentMessage={isPositiveRatingSelected ? t('positiveComment') : t('negativeComment')}
           onCommentChanged={onCommentChanged}
           required={!isPositiveRatingSelected} />
+        <Description>{requestError ? t('failedSendingFeedback') : null}</Description>
         <TextButton
           disabled={!isPositiveRatingSelected && !comment}
-          onClick={onSubmit}
+          onClick={requestError ? null : onSubmit}
```
klinzo, line 81: This does not seem right (especially in terms of ui/ux, but also where this is handled). You shouldn't just disable a button without giving the user a hint that it is disabled and providing an option to enable it again. Imo the logic to show that sending the feedback failed should not reside here, but instead either in the refactored `FeedbackThanksMessage` (possibly with a try again button) or a entire new component, both rendered in the `FeedbackModal`. If you use the suggestion of the enum, you could switch between its value to render either a `FeedbackBox`, a `FeedbackThanksMessage` or a `FeedbackError` there.

src/modules/feedback/components/FeedbackModal.js
```diff
@@ -64,7 +64,9 @@ type StateType = {|
 export class FeedbackModal extends React.Component<PropsType, StateType> {
   state = { feedbackSent: false }
 
-  handleSubmit = () => this.setState({ feedbackSent: true })
+  handleSubmit = (arg: any) => {
```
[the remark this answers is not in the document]
lulo4907: changed

src/modules/feedback/components/FeedbackBoxContainer.js
lulo4907: changed

src/modules/feedback/components/FeedbackBox.js
lulo4907: changed

PR #335, opened 2020-05-13
WEBAPP-399: Add error handling for feedback

lulo4907:

Test Case: 
- checkout the branch
- use feedback action with enabled/disabled network connection


lulo4907:
Several changes until closed #334 Handle errors ...
So this is needs to be the new PR

src/modules/feedback/components/FeedbackModal.js
```diff
@@ -57,34 +57,44 @@ type PropsType = {|
...
+    this.setState({ feedbackSent: 'IDLE' })
     this.props.closeFeedbackModal()
   }
 
   render () {
     const { theme, feedbackStatus, ...otherProps } = this.props
     const { feedbackSent } = this.state
 
+    let boxContainer
+    if (['IDLE', 'ERROR'].includes(feedbackSent)) {
```
klinzo, line 83: It feels kind of strange feedback errors are handled by the `FeedbackBoxContainer`. Imo it should
- either be handled by a separate component (similar to `FeedbackThanksMessage`)
- refactor `FeedbackThanksMessage` to a more generic component rendering different messages, i.e. a thanks message and an error message

src/modules/feedback/components/FeedbackModal.js
```diff
@@ -57,34 +57,44 @@ type PropsType = {|
...
+      boxContainer = <FeedbackBoxContainer isPositiveRatingSelected={feedbackStatus === POSITIVE_RATING}
+                                   {...otherProps}
+                                   onSubmit={this.handleSubmit}
+                                   feedbackSent={feedbackSent}
+                                   theme={theme} />
+    } else if (['SUCCESS'].includes(feedbackSent)) {
```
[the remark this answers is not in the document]
klinzo: In fact, no need for an `else if` at all since all possible values  are `IDLE, ERROR, SUCCESS` and you've already handled both of them above. So an `else` is enough.

src/modules/feedback/components/FeedbackModal.js
```diff
@@ -57,34 +57,44 @@ type PropsType = {|
...
-    this.setState({ feedbackSent: false })
+    this.setState({ feedbackSent: 'IDLE' })
     this.props.closeFeedbackModal()
   }
 
   render () {
     const { theme, feedbackStatus, ...otherProps } = this.props
     const { feedbackSent } = this.state
 
+    let boxContainer
```
klinzo, line 82: We want to avoid `let` because it may be reassigned later on. Either use a separate `render(Feedback)Content` method returning the right content (`FeedbackBoxContainer` or `FeedbackThanksMessage` or use the following:
```suggestion
    const boxContainer = ['IDLE', 'ERROR'].includes(feedbackSent) ? 'FeedbackBoxContainer` : `FeedbackThanksMessage`
```
Of course you would have to use the actual components instead of my simplified version.

src/modules/feedback/components/FeedbackBoxContainer.js
```diff
@@ -44,15 +44,19 @@ type PropsType = {|
...
   closeFeedbackModal: () => void,
-  onSubmit: () => void,
+  onSubmit: (value?: string) => void,
   t: TFunction,
-  theme: ThemeType
+  theme: ThemeType,
+  feedbackSent: string
 |}
 
+type SendingStatusType = 'IDLE' | 'SUCCESS' | 'ERROR'
```
klinzo, line 53: No need to have the same type again, just export it in the other file.

src/modules/feedback/components/FeedbackBoxContainer.js
```diff
@@ -44,15 +44,19 @@ type PropsType = {|
...
+
 type StateType = {|
   feedbackOptions: Array<FeedbackVariant>,
   selectedFeedbackOption: FeedbackVariant,
-  comment: string
+  comment: string,
+  feedbackSent: SendingStatusType
```
klinzo, line 59: No need to define this again, it's a duplication of state. Instead, you should pass `feedbackSent` of `FeedbackModal`'s state down to this component as prop.

src/modules/feedback/components/FeedbackBox.js
```diff
@@ -36,16 +36,26 @@ type PropsType = {|
   comment: string,
   onCommentChanged: SyntheticInputEvent<HTMLTextAreaElement> => void,
   onFeedbackOptionChanged: FeedbackVariant => void,
-  onSubmit: () => void,
+  onSubmit: (value?: string) => void,
   t: TFunction,
   closeFeedbackModal: () => void,
-  theme: ThemeType
+  theme: ThemeType,
+  feedbackSent: string
```
klinzo, line 43: You should use the right type here (`SendingStatusType`).

src/modules/feedback/components/FeedbackBox.js
```diff
@@ -36,16 +36,26 @@ type PropsType = {|
...
 |}
 
 /**
  * Renders all necessary inputs for a Feedback and posts the data to the feedback endpoint
  */
 export class FeedbackBox extends React.PureComponent<PropsType> {
+  constructor (props: PropsType) {
+    super(props)
+    this.handleClick = this.handleClick.bind(this)
```
klinzo, line 52: `bind` is only required when using normal functions, e.g. `handleClick () { .... }`. It is not required if you are using arrow functions.

klinzo:
Looking better already!

No need to close/reopen PRs while working on it again. Just keep working. Every reviewer sees already requested changes and can thus assume appropriate changes will be made by you.

src/modules/feedback/components/FeedbackBox.js
```diff
@@ -54,12 +64,17 @@ export class FeedbackBox extends React.PureComponent<PropsType> {
...
-      onSubmit,
       comment,
       closeFeedbackModal,
-      theme
+      theme,
+      feedbackSent
     } = this.props
 
+    let errorMessage
+    if (['ERROR'].includes(feedbackSent)) {
```
Taggotty, line 74: Why not `feedbackSent === 'ERROR'` ?

src/modules/feedback/components/FeedbackBoxContainer.js
```diff
@@ -232,17 +236,18 @@ export class FeedbackBoxContainer extends React.Component<PropsType, StateType>
...
+    await this.postFeedbackData(this.getFeedbackData(selectedFeedbackOption, comment))
   }
 
   render () {
-    const { closeFeedbackModal, isPositiveRatingSelected, theme } = this.props
+    const { closeFeedbackModal, isPositiveRatingSelected, theme, feedbackSent } = this.props
+
     return <FeedbackBox onFeedbackOptionChanged={this.handleFeedbackOptionChanged}
                         onCommentChanged={this.handleCommentChanged}
                         onSubmit={this.handleSubmit}
```
Taggotty, line 249: Typing feels off here. `onSubmit` is typed as `(value?: string) => void` but `this.handleSubmit` does not expect any parameter.

src/modules/feedback/components/__tests__/FeedbackBoxContainer.spec.js
```diff
@@ -356,6 +364,7 @@ describe('FeedbackBoxContainer', () => {
         extras={null}
         postFeedbackDataOverride={mockPostFeedbackData}
         theme={theme}
+        feedbackSent='SUCCESS'
```
Taggotty, line 367: could you add one test where submitting feedback fails?
And when looking at the tests it feels stange that feedback sent is a property of the FeedbackBoxContainer, as it is represents more the State of the container I think. What was the reasonin behind modelliin it as a property?

src/modules/feedback/components/FeedbackBox.js
```diff
@@ -36,16 +36,26 @@ type PropsType = {|
   comment: string,
   onCommentChanged: SyntheticInputEvent<HTMLTextAreaElement> => void,
   onFeedbackOptionChanged: FeedbackVariant => void,
-  onSubmit: () => void,
+  onSubmit: (value?: string) => void,
```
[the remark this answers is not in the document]
lulo4907: Param name and type changed

src/modules/feedback/components/FeedbackBox.js
lulo4907: Type changed

src/modules/feedback/components/FeedbackBox.js
lulo4907: Constructor and bind() removed

src/modules/feedback/components/FeedbackBox.js
lulo4907: Only to achieve equality with the code at Native-App.

src/modules/feedback/components/FeedbackBox.js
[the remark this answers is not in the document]
klinzo: I guess you forgot to push then :)

src/modules/feedback/components/FeedbackBoxContainer.js
lulo4907: Will use and handle the prop and its value of `feedbackSent` now. But to declare the param and type of  prop `onSubmit: (feedbackSent?: SendingStatusType) => void`, isn't it needful to set the type here?

src/modules/feedback/components/FeedbackBoxContainer.js
lulo4907: The prop `onSubmit` with its optional output value of `feedbackSent` has moved to `postFeedbackData`. It only generates an output depending on the request response. Imo there is no need to pass it here.

src/modules/feedback/components/FeedbackModal.js
```diff
@@ -57,34 +57,44 @@ type PropsType = {|
   theme: ThemeType
 |}
 
+type SendingStatusType = 'IDLE' | 'SUCCESS' | 'ERROR'
+
 type StateType = {|
-  feedbackSent: boolean
+  feedbackSent: SendingStatusType
```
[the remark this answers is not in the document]
lulo4907: `feedbackStatus` is of type `FeedbackRatingType`. I didn't change this.

src/modules/feedback/components/FeedbackModal.js
lulo4907: Refactored with `renderBox()`

src/modules/feedback/components/FeedbackModal.js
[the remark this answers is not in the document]
lulo4907: ok.

klinzo:
@lulo4907 Please start your commit messages with the issue key, e.g. `WEBAPP-399: Adjust naming`. Also, you shouldn't commit lots of changes in one single commit but rather split if many changes.

klinzo:
Apart from that, commenting `changed` or something is not always necessary (github shows an `outdated` note) and if you do it, you should push the code first. Otherwise, we look what you changed but can't see anything since it has not been pushed yet.

src/modules/feedback/components/FeedbackModal.js
lulo4907: should be discussed generally and with regards to the NATIVE-App

src/modules/feedback/components/FeedbackBox.js
```diff
@@ -28,6 +28,7 @@ export const StyledFeedbackBox = styled.div`
 export const Description = styled.div`
   padding: 10px 0 5px;
 `
+type SendingStatusType = 'IDLE' | 'SUCCESS' | 'ERROR'
```
klinzo, line 31: Still a duplicate of the corresponding type in `FeedbackModal`

src/modules/feedback/components/FeedbackBox.js
```diff
@@ -73,9 +85,10 @@ export class FeedbackBox extends React.PureComponent<PropsType> {
           commentMessage={isPositiveRatingSelected ? t('positiveComment') : t('negativeComment')}
           onCommentChanged={onCommentChanged}
           required={!isPositiveRatingSelected} />
+        {errorMessage}
```
klinzo, line 88: Just inline this to get rid of evil `let`s :imp: :
```suggestion
        {feedbackSent === 'ERROR' && <Description>{t('failedSendingFeedback')}</Description>}
```

src/modules/feedback/components/FeedbackBox.js
```diff
@@ -46,6 +48,11 @@ type PropsType = {|
  * Renders all necessary inputs for a Feedback and posts the data to the feedback endpoint
  */
 export class FeedbackBox extends React.PureComponent<PropsType> {
+  handleClick = () => {
```
klinzo, line 51: No need to handle this here. This should be handled in the top component (`FeedbackModal` in this case).

src/modules/feedback/components/FeedbackBoxContainer.js
```diff
@@ -33,6 +33,8 @@ import { DISCLAIMER_ROUTE } from '../../app/route-configs/DisclaimerRouteConfig'
 import { cmsApiBaseUrl } from '../../app/constants/urls'
 import type { ThemeType } from '../../theme/constants/theme'
 
+type SendingStatusType = 'IDLE' | 'SUCCESS' | 'ERROR'
```
klinzo, line 36: Another duplicate

src/modules/feedback/components/FeedbackBoxContainer.js
klinzo: @Taggotty is right. See my comment in `FeedbackBox`, it is not necessary to call `onSubmit` with `feedbackSent` in the `FeedbackBox`.

src/modules/feedback/components/FeedbackBoxContainer.js
```diff
@@ -66,16 +69,16 @@ export class FeedbackBoxContainer extends React.Component<PropsType, StateType>
   }
 
   postFeedbackData = async (feedbackData: FeedbackParamsType) => {
-    const { postFeedbackDataOverride } = this.props
-
+    const { postFeedbackDataOverride, onSubmit } = this.props
     try {
```
klinzo, line 73: Move the try/catch down below in `handleSubmit` to ensure everything is really catched and handled.

klinzo: The error should also be logged nevertheless.

src/modules/feedback/components/FeedbackModal.js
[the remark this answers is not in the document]
klinzo: But you should change it :smile: Before it was just a boolean which stated whether one had clicked on send. Now, it is more like a status since there are more values.

src/modules/feedback/components/FeedbackModal.js
klinzo: I thought the behavior in the native app was different, which is not the case. Since this implements the same behavior that is present in the native app, I'll resolve this. No further action required here.

src/modules/feedback/components/FeedbackModal.js
```diff
@@ -57,34 +57,46 @@ type PropsType = {|
...
+  }
 
   handleOverlayClick = () => {
-    this.setState({ feedbackSent: false })
+    this.setState({ feedbackSent: 'IDLE' })
     this.props.closeFeedbackModal()
   }
 
-  render () {
+  renderBox (): React.Node {
```
klinzo, line 78: Rename to smth like `renderContent`. It doesn't just render a box.

src/modules/feedback/components/FeedbackModal.js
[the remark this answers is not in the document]
lulo4907: I normally wouldn't change the nomenclature without an agreement, unless it would be an assignment. Please confirm, that this is an assignment. I would then suggest `feedbackRating` for the rating type and `sendingStatus` (as in the NATIVE-App) for the status of sending.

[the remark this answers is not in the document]
klinzo: You don't need an assignment to do something like this. If a variable, class or method name does not make sense anymore, it should be changed.

src/modules/feedback/components/FeedbackBoxContainer.js
```diff
@@ -236,17 +229,24 @@ export class FeedbackBoxContainer extends React.Component<PropsType, StateType>
...
   handleSubmit = async () => {
+    const { onSubmit } = this.props
     const { selectedFeedbackOption, comment } = this.state
-    await this.postFeedbackData(this.getFeedbackData(selectedFeedbackOption, comment))
+    try {
+      await this.postFeedbackData(this.getFeedbackData(selectedFeedbackOption, comment))
+      onSubmit('SUCCESS')
+    } catch (e) {
+      onSubmit('ERROR')
+      console.log(e)
```
klinzo, line 239:
```suggestion
      console.error(e)
      onSubmit('ERROR')
```

src/modules/feedback/components/FeedbackBoxContainer.js
```diff
@@ -46,8 +44,8 @@ type PropsType = {|
   extras: ?Array<ExtraModel>,
   postFeedbackDataOverride?: FeedbackParamsType => void,
   closeFeedbackModal: () => void,
-  feedbackSent: SendingStatusType,
-  onSubmit: (feedbackSent?: SendingStatusType) => void,
+  sendingStatus: SendingStatusType,
+  onSubmit: (sendingStatus?: SendingStatusType) => void,
```
klinzo, line 48:
```suggestion
  onSubmit: (sendingStatus: SendingStatusType) => void,
```
No need for an optional type.

src/modules/feedback/components/FeedbackModal.js
```diff
@@ -44,14 +44,13 @@ const FeedbackContainer = styled.div`
...
 `
-
 type PropsType = {|
   cities: ?Array<CityModel>,
   title?: string,
   path?: string,
   alias?: string,
   query?: string,
-  feedbackStatus: FeedbackRatingType,
+  feedbackRating: FeedbackRatingType,
```
klinzo, line 53: :+1:

klinzo:
Oh and a test case is probably still missing.

src/modules/feedback/components/__tests__/FeedbackBoxContainer.spec.js
lulo4907: Regarding the test I'd like to talk to you.
Regarding the sending status as a prop, I thought to place the error feedback text at the box component between the button and the text field. Therefor the prop seemes to be useful.

src/modules/feedback/components/FeedbackBoxContainer.js
```diff
@@ -235,8 +236,8 @@ export class FeedbackBoxContainer extends React.Component<PropsType, StateType>
       await this.postFeedbackData(this.getFeedbackData(selectedFeedbackOption, comment))
       onSubmit('SUCCESS')
     } catch (e) {
-      onSubmit('ERROR')
       console.log(e)
```
klinzo, line 239: Should be `console.error(e)`.

src/modules/feedback/components/__tests__/FeedbackBoxContainer.spec.js
lulo4907: @Taggotty The specs has been completed ;)

lulo4907: @klinzo The Flow type monitor `src/modules/feedback/components/__tests__/FeedbackBoxContainer.spec.js:391:5` cannot assign function to `console.error` because property `error` is not witeable. Any suggstions?

src/modules/feedback/components/__tests__/FeedbackBoxContainer.spec.js
```diff
@@ -356,13 +364,41 @@ describe('FeedbackBoxContainer', () => {
...
+        onSubmit={mockOnSubmit}
+        closeFeedbackModal={() => {}}
+        extras={null}
+        theme={theme}
+        sendingStatus='ERROR'
+        t={t} />
+    )
+    // set as workaround to override: console error to log
+    const prevError = console.error
+    console.error = error => console.log(`Some expected error was thrown: ${error}`)
```
Taggotty, line 391: @maxammann @klinzo can you tell me why we override the error stream in the `setupTest.js` like this
```
console.error = error => {
  throw Error(error)
}
```
It seems this only produces inconveniences.

klinzo: I don't know, not my code. I thought this was default behavior of jest to fail tests if something is logged to error... Thanks for pointing it out! @maxammann ?

src/modules/feedback/components/FeedbackBoxContainer.js
lulo4907: changed

PR #336, opened 2020-05-20
WEBAPP-562: Test content not found error poi

Taggotty:


Taggotty:
The changes in FailureSwitcher are just what I think fits better to our mindset to prefer testing functunality. If you don't like it feel free to object :)

src/modules/common/components/__tests__/Failure.spec.js
```diff
@@ -0,0 +1,34 @@
+// @flow
+
+import React from 'react'
+import { shallow } from 'enzyme'
+
+import { Failure } from '../Failure'
+
+describe('Failure', () => {
+  const mockTranslate = key => key || 'null'
```
klinzo, line 9: Why do we need `|| 'null'`?

src/modules/common/components/__tests__/FailureSwitcher.spec.js
```diff
@@ -29,16 +28,30 @@ describe('FailureSwitcher', () => {
       const error = new ContentNotFoundError({ type: 'event', id: '1234', language, city })
       expect(FailureSwitcher.renderContentNotFoundComponent(error)).toMatchSnapshot()
     })
+
+    it('should render a poi not found failure and match snapshot', () => {
+      const error = new ContentNotFoundError({ type: 'poi', id: 'poiId', language, city })
+      expect(FailureSwitcher.renderContentNotFoundComponent(error)).toMatchSnapshot()
```
klinzo, line 34: Do you think it makes sense to just assert on this method instead of really rendering the FailureSwitcher?

src/modules/common/components/__tests__/FailureSwitcher.spec.js
```diff
@@ -29,16 +28,30 @@ describe('FailureSwitcher', () => {
...
-    )).toMatchSnapshot()
+    const error = new ContentNotFoundError({ type: 'extra', id: 'sprungbrett', city, language })
+    const spy = jest.spyOn(FailureSwitcher, 'renderContentNotFoundComponent')
+
+    const component = mount(<FailureSwitcher error={error} />)
+
+    expect(spy).toHaveBeenCalledWith(error)
+    expect(component.find(Failure)).toHaveLength(1)
+
+    spy.mockRestore()
```
klinzo, line 47: Isn't this kind of a duplicate of `should render a extra not found failure and match snapshot`?

src/modules/common/components/__tests__/FailureSwitcher.spec.js
klinzo: Still relevant

Taggotty: Hmm. It seems github didn't post my answer. I think it makes sense, as the Failure.spec.js should make sure that rendering of the Failure works. FailureSwitcher should just make sure that the right arguments are passed, don't you think?

src/modules/common/components/Failure.js
```diff
@@ -30,14 +30,18 @@ type PropsType = {|
...
+
   render () {
     const { t, errorMessage, goToPath, goToMessage } = this.props
-    const message = t(goToMessage || 'goTo.start')
 
     return <Centered>
       <div>{t(errorMessage)}</div>
       <div><FontAwesomeIcon icon={faFrown} size='5x' /></div>
-      <Link to={goToPath || new I18nRedirectRouteConfig().getRoutePath({})}>{message}</Link>
+      <Link to={goToPath}>{t(goToMessage)}</Link>
```
maxammann, line 44: Why was this removed?

src/modules/common/components/__tests__/Failure.spec.js
```diff
@@ -0,0 +1,34 @@
...
+  it('should render a simple failure and match snapshot', () => {
+    const errorMessage = 'error message'
+    const component = shallow(<Failure errorMessage={errorMessage} t={mockTranslate} />)
+
+    expect(component.instance().props).toEqual({
+      errorMessage: errorMessage,
+      t: mockTranslate,
+      ...Failure.defaultProps
+    })
+    expect(component).toMatchSnapshot()
```
maxammann, line 20: I think instead of doing a snapshot test here it would be good to just assert whether the expected text exists using: https://enzymejs.github.io/enzyme/docs/api/ReactWrapper/text.html

Or alternatively whether a link exists to the target you expect.

src/modules/common/components/__tests__/Failure.spec.js
```diff
@@ -0,0 +1,34 @@
...
+  it('should render a failure with goToPath and goToMessage and match snapshot', () => {
+    const error = {
+      errorMessage: 'error message',
+      goToPath: 'goTo.extras',
+      goToMessage: 'goTo.extras'
+    }
+    const component = shallow(<Failure {...error} t={mockTranslate} />)
+
+    expect(component.instance().props).toEqual({ ...error, t: mockTranslate })
+    expect(component).toMatchSnapshot()
```
maxammann, line 32: I think instead of doing a snapshot test here it would be good to just assert whether the expected text exists using: https://enzymejs.github.io/enzyme/docs/api/ReactWrapper/text.html

Or alternatively whether a link exists to the target you expect.

src/modules/common/components/__tests__/__snapshots__/FailureSwitcher.spec.js.snap
```diff
@@ -1,39 +1,49 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
```
maxammann, line 1: These snapshots look clean :+1:

src/modules/common/components/__tests__/FailureSwitcher.spec.js
```diff
@@ -1,44 +1,49 @@
...
+    ${'localNewsItem'}  | ${'/augsburg/en/news/local/1'}
+    ${'tunewsItem'}     | ${'/augsburg/en/news/tu-news/1'}
+    ${'extra'}          | ${'sprungbrett'}
+    ${'poi'}            | ${'1234'}
+    `('render $type component not found failure',
+  ({ type, id }) => {
+    it(`should render a ${type} not found failure and match snapshot`, () => {
+      const error = new ContentNotFoundError({ type, id, language, city })
+      const component = FailureSwitcher.renderContentNotFoundComponent(error)
+      expect(component).toMatchSnapshot()
```
maxammann, line 28: Maybe a `.dive()` would be cool here. I think then the Translate HOC would be gone in the tests.

maxammann:
Examples for non-snapshot tests:

```
it('should show correct text', () => {
  const wrapper = shallow(<MyComponent />);
  expect(wrapper.text().includes('my text')).toBe(true);
});
```

```
const wrapper = shallow(<div><button className='btn btn-primary'>OK</button></div>);
const button = wrapper.find('.btn'); 
expect(button.text()).to.be.eql('OK');
```

src/modules/common/components/Failure.js
Taggotty: Replaced with default properties to test whether the properties rather than the dom elements.

src/modules/common/components/__tests__/FailureSwitcher.spec.js
Taggotty: does not work, as dive is an enzyme functionality and the component is not rendered with shallow or mount. The current solution definitly looks cleaner, than a render tree (even with shallow)

PR #338, opened 2020-05-25
WEBAPP-553: Build configs

klinzo:


.circleci/config.yml
```diff
@@ -8,43 +8,44 @@ parameters:
...
-          shouldBumpVersion: << pipeline.parameters.production >>
+          build_config: << pipeline.parameters.build_config >>
           requires:
             - build
       - deliver:
-          production: << pipeline.parameters.production >>
+          deliver_to: << pipeline.parameters.build_config >>
           requires:
             - build
             - bump_version
```
Taggotty, line 46: does that mean if you manually build integreat-debug the version number is increased?

klinzo: No it shouldn't, take a look at `bump_version`. It is only bumped if config is `integreat`. Didn't find a better solution.

build/webpack.config.babel.js
```diff
@@ -8,29 +8,36 @@ const babelConfig = require('../.babelrc.js')
...
+    throw new Error('You need to specify a config name!')
+  } else if (!validConfigNames.includes(appConfigName)) {
+    throw new Error(`Invalid config name! Allowed configs: ${validConfigNames}`)
   }
 
-  const isDebug = env.prod === 'false'
-  const appConfigName = env.config_name || 'integreat'
-  const appConfig = require(`./${appConfigName}-config`)
+  // If prod is not specified manually, default to true for 'integreat' and 'malte' configs, false for 'integreat-debug'
+  const production = prod || ['integreat', 'malte'].includes(appConfigName)
```
Taggotty, line 21: when looking at the package.json prod is not defined in Debug mode. Shouldn't it be the other way around then?

package.json
```diff
@@ -7,10 +7,11 @@
     "npm": ">=3.8"
   },
   "scripts": {
-    "start": "webpack-dev-server --config tools/webpack.config.babel.js --progress --env.prod=false --env.config_name=integreat-debug",
-    "build": "webpack --config tools/webpack.config.babel.js --env.prod=true --env.config_name=integreat",
-    "build:debug": "webpack --config tools/webpack.config.babel.js --env.prod=false --env.config_name=integreat-debug",
-    "build:lib": "webpack --config tools/lib.webpack.config.babel.js --env.prod=true && flow-copy-source -i '**/@(__mocks__|__tests__)/**' ./src ./lib-dist && mv ./lib-dist/lib.js.flow ./lib-dist/index.js.flow",
+    "start": "webpack-dev-server --config build/webpack.config.babel.js --progress --env.config_name=integreat-debug",
+    "start:malte": "webpack-dev-server --config build/webpack.config.babel.js --progress --env.config_name=malte",
```
Taggotty, line 11: As stated above env.prod is not defined meaning that prod is undefined

build/configs/integreat/index.js
```diff
@@ -1,14 +1,16 @@
 // @flow
 
-import type { AppConfigType } from '../../src/AppConfigType'
-
-import brightTheme from '../../src/modules/theme/constants/theme'
+import integreatTheme, { darkTheme as darkIntegreatTheme } from '../../themes/integreat'
+import type { AppConfigType } from '../AppConfigType'
+import featureFlags from '../featureFlags'
 
 const IntegreatAppConfig: AppConfigType = {
```
maxammann, line 7: The term `AppConfig` should not be used anymore imo. As stated in the issue the term is `BuildConfig`. Each build config contains Feature Flags.

build/configs/integreat/index.js
```diff
@@ -1,14 +1,16 @@
...
   appTitle: 'Integreat',
   itunesAppId: '1072353915',
-  theme: brightTheme,
+  theme: integreatTheme,
+  darkTheme: darkIntegreatTheme,
   cmsUrl: 'https://cms.integreat-app.de',
+  featureFlags,
```
maxammann, line 13: When writing the issue I also thought that `theme` and `appTitle` etc are feature flags. So you could just add here:
```suggestion
  pois: false,
  newsStream: false
```
When talking about build configs you can say: `I set the feature flag pois is the build config malte to true`.

build/configs/integreat-debug/assets/apple-touch-icon.png
```diff
@@ -0,0 +1 @@
+../../integreat/assets/apple-touch-icon.png
```
maxammann, line 1: Cool that symlinks work here :+1:

package.json
```diff
@@ -7,10 +7,11 @@
...
+    "build": "webpack --config build/webpack.config.babel.js --env.config_name=integreat",
```
maxammann, line 12: I think it is easier to understand if we keep `--env.prod=false` here. That way determining whether you just created a production build or not is just a single "hop" away

.circleci/config.yml
```diff
@@ -8,43 +8,44 @@ parameters:
     type: boolean
     description: Whether the pipeline was triggered through the CircleCi API (https://circleci.com/docs/api/v2/?shell#trigger-a-new-pipeline).
     default: false
-  production:
-    type: boolean
-    description: Whether to deploy to production (integreat.app) or not (webnext.integreat-app.de).
-    default: false
+  build_config:
```
maxammann, line 11: Based on my comment in the webpack config and package.json I think it makes sense to add here a boolean parameter `environment_production` or something named similar to set whether a development or production build should be made.

maxammann: That way it is possible to manually deploy:

- a malteser development build 
- integreat development build 
- integreat production build using the test cms.

build/configs/integreat/index.js
klinzo: As discussed: will keep this, `feature flags` is a `build option`.

.circleci/config.yml
klinzo: Honestly, I don't think we want to do this, especially I'm really sure we don't want to deploy a debug build/test cms build to production.

klinzo: The only use case I'm seeing is deploying a production build to webnext for debugging purposes like the ie11 error.

klinzo: So imo we should have the following possibilites:
- Build config: Integreat, Deploy to: integreat.app, Production build: true
- Build config: Integreat test cms, Deploy to: webnext, Production build: true/false
- Build config: Malte, Deploy to malte.xyz, Production build: true
(At some point: Build config: Malte test cms, Deploy to malte webnext, Production build: true/false)
And I honestly don't now if we need the possibility to deploy a production test cms build at all.

.circleci/config.yml
```diff
@@ -113,18 +133,9 @@ jobs:
...
-                command: yarn run build
-      - unless:
-          condition: << parameters.production >>
-          steps:
-            - run:
-                name: Debug build
-                command: yarn run build:debug
+      - run:
+          name: Build << parameters.build_config >>
+          command: yarn run webpack --config build/webpack.config.babel.js --env.config_name=<< parameters.build_config >>
```
klinzo, line 138: TODO: Specify production

.circleci/config.yml
klinzo: As discussed in slack, won't do.

.circleci/config.yml
```diff
@@ -149,30 +172,39 @@ jobs:
...
+                command: echo "Skip delivery"
 
   bump_version:
     parameters:
-      shouldBumpVersion:
-        description:
-        type: boolean
-        default: true
+      build_config:
+        description: Which build config to use. Only bump version if build_config is "integreat".
```
Taggotty, line 205: then why are "integreat-test-cms" and "malte" still part of the enum?

klinzo: We can't conditionally run jobs in circleci, therefore we have to run `bump_version` all the time we use a workflow which sometimes bumps a version and sometimes not, which is the case for the api_triggered workflow. Therefore, the passed build_config may also be malte and test cms.

klinzo: But we can conditonally run workflows... :thinking:  I guess we should do that then. Thanks!

klinzo: @Taggotty what do you think of my change?

.circleci/config.yml
```diff
@@ -54,23 +71,41 @@ workflows:
...
+      - schedule:
+          cron: "0 5 1,15 * *" # For sanity checks for this syntax: https://crontab.guru/
+          filters:
+            branches:
+              only:
+                - develop
+    jobs:
+      - build:
+          build_config: "malte"
+      - deliver:
```
maxammann, line 96: This will not work as the type is `approval`. A step called like `await_production_approval` is required.

maxammann:
```
      - await_production_approval:
          type: approval
          requires:
            - bump_version
```

klinzo: No, this is not needed. Will work anyway. No extra step required.

.circleci/config.yml
```diff
@@ -8,43 +8,60 @@ parameters:
...
             branches:
               only:
                 - develop
+          deliver_to: "integreat-test-cms"
+          requires:
+            - build
 
-  api-triggered-delivery: # Workflow for manual delivery
-    when: << pipeline.parameters.api_triggered >>
+  api-triggered-integreat-delivery: # Workflow for manual delivery of integreat which also bumps the version
```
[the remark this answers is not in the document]
klinzo: What do you mean by `commit does not deliver`?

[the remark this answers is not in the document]
maxammann: The commit workflow only delivers on develop so it makes sense that the other workflows have `delivery` in the name.

.circleci/config.yml
maxammann: Did you test this? For me it did not work when I tried it. I could click approve but the job was not executed.

.circleci/config.yml
```diff
@@ -8,43 +8,60 @@ parameters:
...
             - build
       - deliver:
-          production: << pipeline.parameters.production >>
+          deliver_to: << pipeline.parameters.build_config >>
           requires:
             - build
             - bump_version
 
-  bi-weekly-production-delivery:
+  api-triggered-delivery: # Workflow for manual delivery of integreat-test-cms and malte
```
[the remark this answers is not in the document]
klinzo: Yes, we probably should only deliver to both malte and integreat or malte-test-cms and integreat-test-cms at the same time. I'll create an issue for this. If we allow separate releases, we shouldn't bump the version for malte alone imo.

.circleci/config.yml
klinzo: Yes, already approved one delivery.

klinzo: Edit: Alright, I see. Yeah seems like this is also the case for me, thanks for pointing it out.

.circleci/config.yml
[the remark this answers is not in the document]
klinzo: Yeah I guess it would make sense to change the naming here.

.circleci/config.yml
klinzo: @maxammann  Should I keep the required approval or should we just deliver?

.circleci/config.yml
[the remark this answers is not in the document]
klinzo: I'll add this to https://issues.integreat-app.de/browse/WEBAPP-558

.circleci/config.yml
maxammann: For malte we can remove the approval, for integreat.app we should keep it.

klinzo: :+1:

klinzo: @maxammann done

PR #339, opened 2020-05-26
WEBAPP-416: Add test for pagedetail

Taggotty:


Taggotty:
I guess this is this kind of class where you have to use snapshots.

klinzo:
@Taggotty please use a ':' after issue number and upper case for the first word :eyes: It really hurts my longing for order and consistency :stuck_out_tongue_winking_eye:

src/modules/common/components/__tests__/PageDetail.spec.js
```diff
@@ -0,0 +1,15 @@
...
+import { shallow } from 'enzyme'
+import PageDetail from '../PageDetail'
+
+describe('PageDetail', () => {
+  const identifier = 'Date'
+  const information = 'May 22, 2020 1:00 AM'
+
+  it('should render', () => {
+    expect(shallow(<PageDetail identifier={identifier} information={information} />
+    )).toMatchSnapshot()
```
maxammann, line 13: I think we should just assert whether `'Date'` and `'May 22, 2020 1:00 AM'` exists in the DOM here.
Else this test will break because we switch the span to a div e.g.

PR #346, opened 2020-05-30
WEBAPP-452: Fix url matching for events and pois

Schedulaar:


src/routes/events/containers/EventsPage.js
```diff
@@ -34,9 +33,9 @@ export class EventsPage extends React.Component<PropsType> {
     <EventListItem event={event} language={language} key={event.path} />
 
   render () {
-    const { events, path, eventId, city, language, t } = this.props
+    const { events, eventId, city, language, t } = this.props
     if (eventId) {
-      const event = events.find(_event => _event.path === decodeURIComponent(path))
+      const event = events.find(_event => _event.path === `/${city}/${language}/events/${eventId}`)
```
klinzo, line 38: Why do we need to check for the whole path? Isn't there a property like `eventId`?

Schedulaar: Unfortunately, no there isn't.

Schedulaar:
@klinzo The other routes worked fined (I already had adjusted the POIs route) except for some links on the tunews / localnews route. I fixed those.

PR #347, opened 2020-05-31
WEBAPP-550: Disable events if flag is not set

Schedulaar:

### DIFF
--- a/src/modules/app/route-configs/EventsRouteConfig.js
+++ b/src/modules/app/route-configs/EventsRouteConfig.js
@@ -4,6 +4,7 @@ import type { AllPayloadsType } from './RouteConfig'
 import { RouteConfig } from './RouteConfig'
 import type { Route } from 'redux-first-router'
 import {
+  CityModel,
   createCitiesEndpoint,
   createEventsEndpoint,
   createLanguagesEndpoint,
@@ -13,8 +14,8 @@ import {
 import fetchData from '../fetchData'
 import { cmsApiBaseUrl } from '../constants/urls'
 
-type EventsRouteParamsType = {|city: string, language: string|}
-type RequiredPayloadsType = {|events: Payload<Array<EventModel>>|}
+type EventsRouteParamsType = {| city: string, language: string |}
+type RequiredPayloadsType = {| events: Payload<Array<EventModel>>, cities: Payload<Array<CityModel>> |}
 
 export const EVENTS_ROUTE = 'EVENTS'
 
@@ -35,6 +36,7 @@ const eventsRoute: Route = {
     ])
   }
 }
+
 class EventsRouteConfig implements RouteConfig<EventsRouteParamsType, RequiredPayloadsType> {
   name = EVENTS_ROUTE
   route = eventsRoute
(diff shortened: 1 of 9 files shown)


src/modules/app/route-configs/EventsRouteConfig.js
```diff
@@ -52,16 +54,20 @@ class EventsRouteConfig implements RouteConfig<EventsRouteParamsType, RequiredPa
...
-  getRequiredPayloads = (payloads: AllPayloadsType): RequiredPayloadsType => ({ events: payloads.eventsPayload })
+  getRequiredPayloads = (payloads: AllPayloadsType): RequiredPayloadsType => ({
+    cities: payloads.citiesPayload,
+    events: payloads.eventsPayload
+  })
 
-  getPageTitle = ({ t, payloads, cityName, location }) => {
-    if (!cityName) {
+  getPageTitle = ({ t, payloads, location }) => {
+    const city = payloads.cities.data && payloads.cities.data.find(_city => _city.code === location.payload.city)
```
klinzo, line 63: I really hate that we have to do this checks all the time...

klinzo: Maybe we could refactor this by adding the current city model to the location or something like that...

src/modules/app/route-configs/EventsRouteConfig.js
```diff
@@ -52,16 +54,20 @@ class EventsRouteConfig implements RouteConfig<EventsRouteParamsType, RequiredPa
...
 
-  getRequiredPayloads = (payloads: AllPayloadsType): RequiredPayloadsType => ({ events: payloads.eventsPayload })
+  getRequiredPayloads = (payloads: AllPayloadsType): RequiredPayloadsType => ({
+    cities: payloads.citiesPayload,
+    events: payloads.eventsPayload
+  })
 
-  getPageTitle = ({ t, payloads, cityName, location }) => {
-    if (!cityName) {
+  getPageTitle = ({ t, payloads, location }) => {
```
klinzo, line 62: The cityName isn't passed to this function? What about the other routes? Why does it work for e.g. `CategoriesRoute`?

src/modules/app/route-configs/ExtrasRouteConfig.js
```diff
@@ -47,12 +48,23 @@ class ExtrasRouteConfig implements RouteConfig<ExtrasRouteParamsType, RequiredPa
...
+  })
 
   getLanguageChangePath = ({ location, language }) =>
     this.getRoutePath({ city: location.payload.city, language })
 
-  getPageTitle = ({ t, cityName }) => cityName ? `${t('pageTitles.extras')} - ${cityName}` : null
+  getPageTitle = ({ t, payloads, location }) => {
+    const cityModel = payloads.cities.data
+      ? payloads.cities.data.find(cityModel => cityModel.code === location.payload.city)
+      : null
```
klinzo, line 62:
```suggestion
    const cityModel = payloads.cities.data
      && payloads.cities.data.find(cityModel => cityModel.code === location.payload.city)
```

src/modules/app/route-configs/ExtrasRouteConfig.js
```diff
@@ -47,12 +48,23 @@ class ExtrasRouteConfig implements RouteConfig<ExtrasRouteParamsType, RequiredPa
...
+    if (!cityModel || !cityModel.extrasEnabled) {
+      return null
```
klinzo, line 64: This means the page name defaults to `Integreat`, right?

src/modules/app/route-configs/EventsRouteConfig.js
klinzo: Of course not in the scope of this PR

src/modules/app/route-configs/EventsRouteConfig.js
Schedulaar: Oh yes, it is, but as I need the complete cityModel anyways, I decided not to take it from there

src/modules/app/route-configs/EventsRouteConfig.js
Schedulaar: Yeah, we could think about replacing `cityName: ?string` with `cityModel: ?CityModel` in `getPageTitle`

src/modules/app/route-configs/ExtrasRouteConfig.js
Schedulaar: Yes

Schedulaar:
@klinzo What changes did you request? Afaics, your comments are only slight codestyle preferences.

PR #352, opened 2020-06-02
WEBAPP-536: Create release notes

klinzo:


klinzo:
Tags not yet shown in the github tag/release section. Will be done in https://issues.integreat-app.de/browse/WEBAPP-552

build/create-jira-release
```diff
@@ -0,0 +1,120 @@
...
+  const { values: versions } = await getFromJira(
+    `/rest/api/2/project/${JIRA_PROJECT}/version?orderBy=-releaseDate&maxResults=2`
+  )
+
+  const lastVersion = versions[1]
+  const lastDate = lastVersion.releaseDate
+  const newVersion = versions[0]
+
+  // Get all issues which should be part of the new release
+  // TODO Make project configurable
```
klinzo, line 83: TODO

.circleci/config.yml
```diff
@@ -173,6 +173,27 @@ jobs:
...
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Calculate next version
+          command: echo "export NEW_VERSION_NAME=$(build/next-version calc | jq .versionName)" >> ${BASH_ENV}
+      - run:
+          name: Create jira release
+          command: build/create-jira-release createRelease ${NEW_VERSION_NAME} --private-key ${JIRA_PRIVATE_KEY} --access-token ${JIRA_OAUTH_TOKEN} --consumer-key ${JIRA_CONSUMER_KEY}
```
maxammann, line 189: To be consistent this should be `create_release` as `move_to` and `bump_to`

.circleci/config.yml
```diff
@@ -190,17 +211,12 @@ jobs:
...
-                command: |
-                  echo "export VERSION_NAME=$(cat version.json | jq .versionName)" >> $BASH_ENV
-                  source $BASH_ENV
             - run:
                 name: Integreat production delivery # StrictHostKeyChecking=no is not a security problem. The worst that could happen is a delivery to the wrong domain.
                 command: sftp -v -o StrictHostKeyChecking=no web@web.integreat-app.de:/integreat.app \<<< $'put -r /tmp/output/dist/*'
             - slack/notify:
-                message: Integreat WebApp version ${VERSION_NAME} was delivered successfully to integreat.app!
-                webhook: $SLACK_URL
+                message: Integreat WebApp version $NEW_VERSION_NAME was delivered successfully to integreat.app!
```
maxammann, line 218: Are you sure that the $NEW_VERSION_NAME gets resolved properly? The `prepare-integreat-delivery` job adds it to the BASH_ENV. But this bash env is not available here because it is run in another docker image

build/create-jira-release
maxammann: Should this be done in this task? This is no longer a Draft.

build/create-jira-release
```diff
@@ -0,0 +1,120 @@
...
+  const query = `project = integreat-webapp AND fixVersion IS EMPTY AND resolution = Done AND resolutiondate > ${lastDate}`
```
maxammann, line 84: This needs to be configurable

build/move-release-notes
```diff
@@ -0,0 +1,78 @@
...
+
+  // Tree of the 'release-notes' folder of the project
+  const releaseNoteTreeSha = rootTree.data.tree.find(it => it.path === RELEASE_NOTES_DIR).sha
+  const releaseNoteTree = await appOctokit.git.getTree({ owner, repo, tree_sha: releaseNoteTreeSha })
+
+  // Rename the dir 'unreleased' to <newVersionName> and keep everything else
+  const renameUnreleasedDir = releaseNoteTree.data.tree.map(it => {
+    const { size, url, ...rest } = it
+    if (it.path === UNRELEASED_DIR) {
+      return { ...rest, path: newVersionName }
```
maxammann, line 33: Does this mean the unreleased directory gets removed? Should we put an '.gitkeep' in that folder such that we do not have to recreate this directory every time after a new release?

.circleci/config.yml
klinzo: Yeah, I wanted to ask this anyway. We are mixing `-` and `_` all the time. Is there a convention for scripts? Or should we just decide on either one? I am fine with both @maxammann

build/create-jira-release
klinzo: Yep I know, but as this is only a minor issue I wanted to open the PR already in order not to wait that long for reviews.

.circleci/config.yml
klinzo: It is available, I tested it.

build/create-jira-release
klinzo: Yeah, thats why I added a TODO in the code and a comment here in the PR. Or what exactly should be configurable?

build/move-release-notes
klinzo: Yes. We could do that, yes. Good suggestion.

build/create-jira-release
```diff
@@ -0,0 +1,120 @@
...
+    releaseDate: newReleaseDate,
+    project: JIRA_PROJECT
+  }, accessToken, privateKey, consumerKey)
+
+  // Get the new and the last release
+  const { values: versions } = await getFromJira(
+    `/rest/api/2/project/${JIRA_PROJECT}/version?orderBy=-releaseDate&maxResults=2`
+  )
+
+  const lastVersion = versions[1]
```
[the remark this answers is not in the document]
klinzo: Done

build/move-release-notes
klinzo: Hmm quite some more work needed as we have to create a new tree and the .gitkeep every time in the script here. Also we'd probably have to delete it from the new release folder.

The reason for this is that we don't remove the unreleased directory but instead just rename it.

build/create-jira-release
```diff
@@ -0,0 +1,126 @@
...
+  const newVersion = versions[0]
+
+  if (newVersion.name !== newVersionName) {
+    throw Error(`There is already a release ${newVersion.name} with a later release date in the project!`)
+  }
+
+  // Get all issues which should be part of the new release
+  const query =
+    `project = ${projectName} AND fixVersion IS EMPTY AND resolution = Done AND resolutiondate > ${lastDate}`
+  const { issues } = await getFromJira(`/rest/api/2/search?jql=${query}&fields=id`)
```
Taggotty, line 93: this returns maximum of 50 issues. We should either paginate here or set maxResults to a higher number. Even though I am not sure of how many issues we are talking about here normally. ^^

build/move-release-notes
```diff
@@ -0,0 +1,78 @@
...
+const moveReleaseNotes = async ({ newVersionName, deliverinoPrivateKey, owner, repo, branch, commitSha }) => {
+  const appOctokit = await authenticate({ deliverinoPrivateKey, owner, repo })
+  const lastCommit = await appOctokit.git.getCommit({ owner, repo, commit_sha: commitSha })
+
+  // Tree of the root folder of the project
+  const rootTreeSha = lastCommit.data.tree.sha
+  const rootTree = await appOctokit.git.getTree({ owner, repo, tree_sha: rootTreeSha })
+
+  // Tree of the 'release-notes' folder of the project
+  const releaseNoteTreeSha = rootTree.data.tree.find(it => it.path === RELEASE_NOTES_DIR).sha
```
Taggotty, line 26: it could occur, that the release notes are not found, becouse getTree truncates if there are too many elements(?). Could you add a check, that if the release notes directory cannot be found we check whether `rootTree.data.truncated` is true and throw an Error message.
Even though I don't think the chances are high of thet occuring I still think it is better to be prepared for stupid bugs like that :)

build/create-jira-release
klinzo: Hmm in theory we do a release every twice a month, I don't think we'll ever make it to resolve more than 50 issues in that time :D

.circleci/config.yml
```diff
@@ -175,17 +196,12 @@ jobs:
...
+                webhook: ${SLACK_URL}
```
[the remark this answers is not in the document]
klinzo: AFAIK `${}` does prevent escaping of special characters, so I guess this is the way to go. Will change.

build/move-release-notes
klinzo: `If truncated is true in the response then the number of items in the tree array exceeded our maximum limit. If you need to fetch more items, use the non-recursive method of fetching trees, and fetch one sub-tree at a time.` Since we don't use recursive fetch, I think we should be fine here. But I'll check and throw an error.

build/create-jira-release
klinzo: Increased to 1000 issues. If this is not enough, I guess we have other problems :sunglasses:

build/move-release-notes
klinzo: Done

build/move-release-notes
Taggotty: Nice 👍

lulo4907:
Please let me know, how to review this PR. Would it be possible to run a test build for us? :)

klinzo:
If tested the code and made a sample run, but for obvious reasons deleted the release and everything else again. I guess we'll just have to fix it later if problems occur.

@lulo4907 kind of not possible without creating releases and so on. Reviewing the code is enough here.

.circleci/config.yml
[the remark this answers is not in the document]
maxammann: I think there should be no difference for our usecases. ${} is just more powerful

build/create-jira-release
```diff
@@ -0,0 +1,129 @@
...
+  // Create a new jira version
+  await postToJira('/rest/api/2/version', {
+    name: newVersionName,
+    released: true,
+    releaseDate: newReleaseDate,
+    project: projectKey
+  }, accessToken, privateKey, consumerKey)
+
+  // Get the new and the last release
+  const { values: versions } = await getFromJira(
```
maxammann, line 78:
```suggestion
  const { values: [lastVersion, newVersion] } = await getFromJira(
```

build/move-release-notes
```diff
@@ -0,0 +1,99 @@
...
+  // Tree of the unreleased release notes
+  const unreleasedTreeObject = releaseNoteTree.data.tree.find(it => it.path === UNRELEASED_DIR)
+  if (!unreleasedTreeObject) {
+    console.error(releaseNoteTree)
+    throw new Error('Release note tree does not contain unreleased tree, probably since the response is truncated.')
+  }
+  const unreleasedTree = await appOctokit.git.getTree({ owner, repo, tree_sha: unreleasedTreeObject.sha })
+
+  // Only keep file '.gitkeep' in 'unreleased' tree, move everything else to '<new-version-name>'
+  const unreleasedReleaseNotes = unreleasedTree.data.tree.filter(it => it.path !== GITKEEP_FILE)
```
maxammann, line 43: :+1:

build/move-release-notes
```diff
@@ -0,0 +1,99 @@
...
+  const newUnreleasedTree = await appOctokit.git.createTree({ owner, repo, tree: [keepFile] })
+
+  const newReleaseNotesContent = releaseNoteTree.data.tree.map(it => {
+    const { size, url, ...rest } = it
+    if (it.path === UNRELEASED_DIR) {
+      return { ...rest, sha: newUnreleasedTree.data.sha }
+    }
+    return rest
+  })
+  newReleaseNotesContent.push({ path: newVersionName, sha: newVersionTree.data.sha, type: 'tree', mode: '040000' })
```
maxammann, line 56: I think the default mode should be '100644'. That is the default on ext4 devices.

klinzo: No, because it is a tree and not a file. See https://developer.github.com/v3/git/trees/#create-a-tree mode.

PR #361, opened 2020-06-09
WEBAPP-581: Implement new nav bar

Schedulaar:
TODO:
* ~Need another locale for `localeInformation` and `pois` in `layout` section of locales~ Waiting for https://github.com/Integreat/integreat-locales/pull/15


src/__mocks__/react-onclickoutside.js
```diff
@@ -8,13 +8,14 @@ import * as React from 'react'
 const onClickOutside = <T> (WrappedComponent: React.ComponentType<T>) => class extends React.Component<T> {
   static displayName = `OnClickOutside(${WrappedComponent.displayName || WrappedComponent.name || typeof WrappedComponent})`
 
-  render () {
-    const component = <WrappedComponent {...this.props} />
-    // $FlowFixMe
-    if (typeof component.type.prototype.handleClickOutside !== 'function') {
+  checkRef = (instance: mixed) => {
```
Taggotty, line 11: Should be `React.ComponentType<T>` right?

Schedulaar: Hmm no... As this is an actual instance which might or might not have the necessary attribute `handleClickOutside`, actually I think in this case `mixed` might be the best fit here; at least `React.ComponentType<T>` throws an error that it doesn't know the attribute handleClickOutside.
I tried some other types which didn't work as well...

src/modules/layout/components/GeneralHeader.js
```diff
@@ -3,20 +3,27 @@
...
   render () {
+    const { viewportSmall, t } = this.props
     const getPath = new I18nRedirectRouteConfig().getRoutePath
-    return <Header viewportSmall={this.props.viewportSmall}
+    return <Header viewportSmall={viewportSmall}
                    logoHref={getPath({})}
-                   actionItems={[new HeaderActionItem({ href: getPath({}), iconSrc: landingIcon })]} />
+                   actionItems={[
+                     <HeaderActionItemLink key='landing' href={getPath({})} iconSrc={landingIcon}
+                                           text={t('changeLocation')} />
```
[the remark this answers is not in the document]
Schedulaar: Well this was only spotted because of my refactoring and flow :D

src/__mocks__/react-onclickoutside.js
Schedulaar: Hmm no... As this is an actual instance which might or might not have the necessary attribute `handleClickOutside`, actually I think in this case `mixed` might be the best fit here; at least `React.ComponentType<T>` throws an error that it doesn't know the attribute handleClickOutside.
I tried some other types which didn't work as well...

Schedulaar:
@klinzo Longer locales behave correctly now; for shorter ones: Unfortunately, they cannot shrink more without wrapping the German "Lokale Informationen" into 2 lines...

Schedulaar:
@klinzo  I've created a task for adjusting the native icons in the dashboard.

PR #362, opened 2020-06-10
WEBAPP-584: Fix general disclaimer

NKlug:
Only thing missing was a pair of parantheses.
The same issue occurred for the NOT_FOUND route, so I also fixed it there.

### DIFF
--- a/src/modules/app/route-configs/MainDisclaimerRouteConfig.js
+++ b/src/modules/app/route-configs/MainDisclaimerRouteConfig.js
@@ -2,11 +2,14 @@
 
 import { RouteConfig } from './RouteConfig'
 
+// This type should be exact, but flow has a bug preventing this: https://github.com/facebook/flow/issues/2977
+type RequiredPayloadsType = {}
+
 export const MAIN_DISCLAIMER_ROUTE = 'MAIN_DISCLAIMER'
 
 const mainDisclaimerRoute = '/disclaimer'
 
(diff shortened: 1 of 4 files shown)


src/modules/app/route-configs/MainDisclaimerRouteConfig.js
```diff
@@ -2,11 +2,14 @@
 
 import { RouteConfig } from './RouteConfig'
 
+// This type should be exact, but flow has a bug preventing this: https://github.com/facebook/flow/issues/2977
```
maxammann, line 5: Try to use an exact type here and use $FlowFixMe to supress the error.

src/modules/app/route-configs/NotFoundRouteConfig.js
```diff
@@ -3,7 +3,10 @@
 import { NOT_FOUND } from 'redux-first-router'
 import type { RouteConfig } from './RouteConfig'
 
-class NotFoundRouteConfig implements RouteConfig<void, void> {
+// This type should be exact, but flow has a bug preventing this: https://github.com/facebook/flow/issues/2977
+type RequiredPayloadsType = {}
```
maxammann, line 7: Try to use an exact type here and use $FlowFixMe to supress the error.

src/modules/app/route-configs/MainDisclaimerRouteConfig.js
NKlug: Done.

PR #363, opened 2020-06-12
WEBAPP-587: Add link to glossary

maxammann:


maxammann:
@lulo4907 what do you mean by that?

PR #364, opened 2020-06-15
WEBAPP-533: Supported browser

lulo4907:

### DIFF
--- a/docs/05-browser-stack.md
+++ b/docs/05-browser-stack.md
@@ -1,7 +1,7 @@
 # Browser Stack
 ## Supported Browser
 The Integreat Web-App has been tested on several OS environments with different browser clients.
-It can be used with most of the modern browsers that has been released up to the last six years.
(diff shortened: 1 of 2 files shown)


docs/05-browser-stack.md
```diff
@@ -0,0 +1,15 @@
+# Browser Stack
+## Supported Browser
+The Integreat Web-App has been tested on several OS environments with different browser clients.
+It can be used with most of the modern browsers that has been released up to the last six years.
```
klinzo, line 4:
```suggestion
It can be used with most of the modern browsers that have been released in the last six years.
```

lulo4907:
@Taggotty Yeah, the test instructions were followed. I think that the overall result is useful here.

klinzo:
> In the issue it says that you should test pdf generation, different screen sizes, etc. did you do this?
> Maybe you could add what has been tested in the browserstack readme

@lulo4907 I think it is a good idea to mention what has been tested here.

PR #365, opened 2020-06-15
WEBAPP-588: Fix IE11 glitches

Schedulaar:

### DIFF
--- a/src/modules/layout/components/Header.js
+++ b/src/modules/layout/components/Header.js
@@ -54,6 +54,10 @@ const Row = styled.div`
     justify-content: space-between;
     flex-wrap: wrap;
     min-height: ${props => props.theme.dimensions.headerHeightSmall}px;
+
+    :first-child { /* this is only necessary for IE11 */
+      min-height: ${props => props.theme.dimensions.headerHeightSmall + (props.hasTitle ? HEADER_TITLE_HEIGHT : 0)}px;
+    }
   }
 `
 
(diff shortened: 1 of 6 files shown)


src/modules/layout/components/Header.js
```diff
@@ -77,14 +78,14 @@ const ActionBar = styled.div`
 
   @media ${props => props.theme.dimensions.smallViewport} {
     order: 2;
-    flex: 1 1 0;
+    flex: 1 1 0px; /* stylelint-disable-line length-zero-no-unit */ /* The px unit is necessary for IE11 */
```
Taggotty, line 81:
```suggestion
    flex: 1 1 0%;
```
Stylelint doesn't say anything against 0%. We could also consider some postcss which fixes also some other [issues regarding flexboxes](https://github.com/philipwalton/flexbugs/blob/master/README.md). For example there is [this project](https://github.com/luisrudge/postcss-flexbugs-fixes), however I don't really like the naming of the "bugs" there. What do you think @maxammann @klinzo ?

Schedulaar:
@lulo4907 Thanks for spotting. It was tedious finding the right attribute, the IE wanted in this specific case, but I managed to find it.

src/modules/layout/components/Header.js
maxammann: I like this idea, and maybe this PR would be a good place to test this. Unfortunately we do not use postcss right now. I don't remember why but we removed it at some point.

Probably because it was not needed back then.

The configuration is simple though:

https://webpack.js.org/loaders/postcss-loader/

Webpack:
```
module.exports = {
  module: {
    rules: [
      {
        test: /\.css$/,
        use: [
          'style-loader',
          { loader: 'css-loader', options: { importLoaders: 1 } },
          'postcss-loader'
        ]
      }
    ]
  }
}
```


PostCSS:

```
module.exports = {
  parser: 'sugarss',
  plugins: {
    'postcss-import': {},
    'postcss-preset-env': {},
    'cssnano': {}
  }
}
```

Schedulaar: I guess the problem is, that styled-components doesn't use post css (not entirely sure, but I don't see how a post-css processor understands the styled-components declarations.
Hence I'd rather create a task for that.

Schedulaar: I researched a bit and unfortunately, i couldn't find any solution for that...

PR #366, opened 2020-06-16
WEBAPP-572: Bugfix calling console.error throws error in tests

Taggotty:

### DIFF
--- a/src/setupTest.js
+++ b/src/setupTest.js
@@ -15,10 +15,5 @@ global.__CONFIG__ = integreatTestCmsBuildConfig
 
 // $FlowFixMe
 console.error = error => {
-  throw Error(error)
-}
-
-// $FlowFixMe
-console.warn = warn => {
-  throw Error(warn)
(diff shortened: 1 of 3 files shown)


src/setupTest.js
```diff
@@ -15,10 +15,5 @@ global.__CONFIG__ = integreatTestCmsBuildConfig
...
 // $FlowFixMe
 console.error = error => {
-  throw Error(error)
-}
-
-// $FlowFixMe
-console.warn = warn => {
-  throw Error(warn)
+  console.log(error)
 }
```
klinzo, line 19: Do we still need this at all? Why not just use the default console.error behavior?

Taggotty: in my opinion we do not need it, but then I was qustioning why we had this code in first place. @maxammann do we need to redirect console.error to console.log?

maxammann: I think we can remove it, that way the app behaves exactly in the same way.

PR #368, opened 2020-06-17
APICLIENT-38: Remove occurences of lodash fn

Schedulaar:
Note: This depends on the changes of https://github.com/Integreat/integreat-api-client/pull/42.

Before:
Screenshot_20200617_151844

After:
Screenshot_20200617_152634


Schedulaar:
Thanks for the hint @maxammann, we also used recompose, which I manage to remove as well (it's also no longer maintained)

PR #370, opened 2020-06-21
WEBAPP-606: Remove unused moment components

Schedulaar:
The last PR in the series "Reduce bundle size with schedulaar" for now :blush: .

Before:
Screenshot_20200621_214706

After:
Screenshot_20200621_215525


package.json
```diff
@@ -42,6 +42,7 @@
     "i18next-browser-languagedetector": "^4.0.1",
     "lodash": "^4.17.11",
     "moment": "^2.24.0",
+    "moment-locales-webpack-plugin": "^1.2.0",
```
klinzo, line 45: Shouldn't this be a dev dependency?

tools/webpack.config.babel.js
```diff
@@ -4,14 +4,27 @@ const AssetsPlugin = require('assets-webpack-plugin')
...
 const readVersionName = () => {
-  const versionFile = fs.readFileSync(path.resolve(__dirname, '../version.json'))
-  return JSON.parse(versionFile).versionName
+  const versionFile = readJson(path.resolve(__dirname, '../version.json'))
+  return versionFile.versionName
+}
+
+const getSupportedLocales = () => {
+  const localesConfig = readJson(path.resolve(__dirname, '../locales/config.json'))
+  return [localesConfig.sourceLanguage, ...localesConfig.targetLanguages]
```
klinzo, line 27: Did you check whether the config is uptodate and contains all languages?

package.json
Schedulaar: Correct, I adjusted it.

tools/webpack.config.babel.js
Schedulaar: It contains all languages which should be translated I guess. Alternatively we could also determine this information by just looking for the root-keys of the `locales.json`. What do you think?

klinzo: I don't think there should be a difference between the keys in the localesConfig and the root keys of the locales.json as I don't see a reason why we would want to have languages in the locales.json that should not be translated?

maxammann: Using that config is fine. This config is used by the manage script. There can not be more or less languages in our locales.json

klinzo: So it's auto generated? I think the manage script is also executed if new translations are to be made, so if we would manually add a language, it would be missing here until the next translation process. However, this should be fine as that probably won't happen.

tools/webpack.config.babel.js
```diff
@@ -95,6 +108,11 @@ const createConfig = (env = {}) => {
     },
     // What information should be printed to the console
     stats: 'minimal',
+    performance: {
+      hints: isProductionBuild ? 'error' : false,
```
maxammann, line 112: @klinzo do you think this is dangerous? 

This line causes release builds to fail. If we do not build release builds on "commit" then we firstly know about big bundle sizes when merging to develop.

I think it is better to always 'error'.

maxammann: Else it is maybe not clear why some builds fail and some not.

klinzo: Yeah builds may fail, which is not dangerous per se, but rather good imo. If the bundle sice becomes a lot bigger than it is right now, it is better to have a failing build than to release it without checking.

klinzo: We do production builds on commit now, too. This should not be a problem therefore.

Schedulaar: We shouldn't error for dev builds as they are **way** larger (5MB or so) since all minimifcation etc. are turned off. Hence it only makes sense to check this for production builds.

tools/webpack.config.babel.js
maxammann: Yeah that will probably not happen. All our translations for new languages come from external services

PR #371, opened 2020-06-23
WEBAPP-602: Tooltips hinder navigation

lulo4907:


src/modules/layout/components/Layout.js
```diff
@@ -17,6 +17,11 @@ const RichLayout = styled.div`
   font-size-adjust: ${props => props.theme.fonts.fontSizeAdjust};
   background-color: ${props => props.theme.colors.backgroundColor};
   line-height: ${props => props.theme.fonts.decorativeLineHeight};
+
+  & a, button {
+   outline: none !important;
```
klinzo, line 22: `!important` should not be used.

lulo4907: done

maxammann:
Tested on Firefox for Mobile on Android

PR #378, opened 2020-07-02
WEBAPP-617: Pass commit sha

klinzo:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,7 +74,7 @@ jobs:
         command: yarn run test:coverage --ci --maxWorkers ${TOTAL_CPUS}
         name: Test coverage
     - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -74,7 +74,7 @@ jobs:
         command: yarn run test:coverage --ci --maxWorkers ${TOTAL_CPUS}
         name: Test coverage
     - run:
-        command: yarn build:integreat-test-cms
+        command: yarn build:integreat-test-cms --env.commit_sha=${CIRCLE_SHA1}
```
Taggotty, line 77: do we also need that for malte?

klinzo: Good catch :+1:

PR #380, opened 2020-07-07
WEBAPP-613: Show Matching Category Content in Search

Taggotty:


src/routes/search/containers/SearchPage.js
```diff
@@ -45,7 +45,7 @@ export class SearchPage extends React.Component<PropsType, LocalStateType> {
     // find all categories whose contents but not titles include the filter text and sort them lexicographically
     const categoriesWithContent = categories.toArray()
       .filter(category => !normalizeSearchString(category.title).includes(filterText))
-      .filter(category => normalizeSearchString(category.content).includes(filterText))
+      .filter(category => normalizeSearchString(category.content).replace(/(<([^>]+)>)/ig, '').includes(filterText))
```
klinzo, line 48: Please add a comment why this is necessary and what kind of regex this is. Is this used to strip html tags?

src/routes/categories/components/CategoryEntry.js
```diff
@@ -0,0 +1,151 @@
...
+      }
+    }
+    return null
+  }
+
+  renderTitle (): React.Node {
+    const { query, category, theme } = this.props
+    return <>
+      <Highlighter searchWords={query ? [query] : []} aria-label={category.title}
+                       sanitize={normalizeSearchString}
```
klinzo, line 124: Indentation

src/routes/categories/components/CategoryEntry.js
```diff
@@ -0,0 +1,151 @@
...
+
+  &:hover {
+    color: inherit;
+    text-decoration: inherit;
+    transition: background-color 0.5s ease;
+    background-color: ${props => props.theme.colors.backgroundAccentColor};
+  }
+`
+
+type PropsType = {
```
klinzo, line 73: Should be strict

src/routes/categories/components/CategoryEntry.js
```diff
@@ -0,0 +1,151 @@
...
+          .split(/[\s,]+/)
+        const matchedWord = (wordsBeforeMatch.pop() ?? '') + (wordsAfterMatch.shift() ?? '')
+        return <ContentMatchItem aria-label={matchedWord} searchWords={[query]}
+                                 sanitize={normalizeSearchString} textToHighlight={matchedWord} highlightStyle={{
+          backgroundColor: theme.colors.backgroundColor,
+          fontWeight: 'bold'
+        }} />
+      }
+    }
+    return null
```
klinzo, line 117: Moving `if (!query || query.length < 1) return null` would make this clearer imo.

src/routes/categories/components/CategoryEntry.js
```diff
@@ -0,0 +1,151 @@
...
+        </StyledLink>
+      </SubCategory>
+    )
+  }
+
+  getMatchedContent (category: CategoryModel): ContentMatchItem {
+    const { query, theme } = this.props
+    if (query && query.length >= 1) {
+      const normalizedFilter = normalizeSearchString(query)
+      const contentWithoutHtml = category.content.replace(/(<([^>]+)>)/ig, '')
```
klinzo, line 100: Please add comments or move regex to separate named var. Same for regexes below.

klinzo: Also, as this just removes tags, are html special characters shown correctly?

src/routes/categories/components/CategoryEntry.js
```diff
@@ -0,0 +1,151 @@
...
+      const matchIdx = contentWithoutHtml.toLowerCase().indexOf(normalizedFilter)
+      if (matchIdx !== -1) {
+        const wordsBeforeMatch = contentWithoutHtml
+          .slice(0, matchIdx)
+          .split(/[\s,]+/)
+        const wordsAfterMatch = contentWithoutHtml
+          .slice(matchIdx)
+          .split(/[\s,]+/)
+        const matchedWord = (wordsBeforeMatch.pop() ?? '') + (wordsAfterMatch.shift() ?? '')
```
klinzo, line 109: What does the `??` operator do?

src/routes/categories/components/CategoryEntry.js
klinzo: Maybe we should use a library for this, e.g. https://www.npmjs.com/package/html-to-text

What do you think?

maxammann:
I think the UI/UX is not that good yet:

image

For me it is not absolutly obvious that the text right to the title is from the content.

src/routes/categories/components/CategoryEntry.js
```diff
@@ -0,0 +1,151 @@
...
+  getMatchedContent (category: CategoryModel): ContentMatchItem {
+    const { query, theme } = this.props
+    if (query && query.length >= 1) {
+      const normalizedFilter = normalizeSearchString(query)
+      const contentWithoutHtml = category.content.replace(/(<([^>]+)>)/ig, '')
+      const matchIdx = contentWithoutHtml.toLowerCase().indexOf(normalizedFilter)
+      if (matchIdx !== -1) {
+        const wordsBeforeMatch = contentWithoutHtml
+          .slice(0, matchIdx)
+          .split(/[\s,]+/)
```
maxammann, line 105: Parsing HTML with regex is potentially dangerous. Are you sure this will work always?

src/routes/categories/components/__tests__/CategoryListItem.spec.js
```diff
@@ -4,7 +4,7 @@ import React from 'react'
 import { shallow } from 'enzyme'
 import moment from 'moment'
 
-import CategoryListItem from '../CategoryListItem'
+import CategoryEntry from '../CategoryEntry'
```
maxammann, line 7: I think the name of CategoryListItem would be fine here.

src/routes/categories/components/__tests__/__snapshots__/CategoryListItem.spec.js.snap
```diff
@@ -75,3 +75,95 @@ exports[`CategoryListItem should replace empty thumbnail 1`] = `
   </CategoryListItem__SubCategory>
 </CategoryListItem__Row>
 `;
+
+exports[`CategoryListItem should render and match snapshot 1`] = `
```
maxammann, line 79: I think this test is not really helpful. Maybe we should assert on something else than the snapshot.

Taggotty:
> I think the UI/UX is not that good yet:
> 
>... 
> 
> For me it is not absolutly obvious that the text right to the title is from the content.

Hmm. What Browser were you using? On my PC It looked fine (in a new line obviously)

src/routes/categories/components/CategoryEntry.js
Taggotty: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

If the first expression is null it returns the second character

src/routes/categories/components/__tests__/__snapshots__/CategoryListItem.spec.js.snap
maxammann: I'm using Firefox

maxammann:
@Taggotty I'm using firefox

src/routes/categories/components/__tests__/CategoryListItem.spec.js
Taggotty: As I took the City_Entry_ as model I came to the conclusion that it is quite confusing to have both names around. especially as the City_Entry_ also has a CityListItem. But I agree, that if it's about consistency, then I should've also renamed the test case. 😅 
But of course you can also tell me to revert to CategoryListItem if I am allowed to also rename the CityEntry 😄

src/routes/categories/components/CategoryEntry.js
Taggotty: I agree. There most certainly might be more use cases for that package later.

src/routes/categories/components/CategoryEntry.js
Taggotty: I also did not like the idea of using a regex here. Found a suitable solution using trim :)

Taggotty: Update. Actually you cannot avoid using \s+ here, as we don't want to display e.g. newlines or tabs in the search. And as far as I found out there is like no other suitable solution for that. According to [this package](https://github.com/davisjam/safe-regex) however this regex should be fine. So I would not avoid using it. But feel free to correct me if I am wrong here.

src/routes/categories/components/CategoryEntry.js
```diff
@@ -0,0 +1,175 @@
...
+import { CategoryModel } from '@integreat-app/integreat-api-client'
+import iconPlaceholder from '../assets/IconPlaceholder.svg'
+import styled, { withTheme } from 'styled-components'
+import Highlighter from 'react-highlight-words'
+import normalizeSearchString from '../../../modules/common/utils/normalizeSearchString'
+import Link from 'redux-first-router-link'
+import type { ThemeType } from '../../../modules/theme/constants/theme'
+import { fromString as htmlToText } from 'html-to-text'
+
+const NUM_WORDS_SURROUNDING_MATCH = 3
```
klinzo, line 14: Not sure if we should increase this number. Not sure though. Maybe put it up to a vote in #app.

src/routes/categories/components/CategoryEntry.js
```diff
@@ -0,0 +1,175 @@
...
+        </StyledLink>
+      </SubCategory>
+    )
+  }
+
+  getWords (content: string): Array<string> {
+    return content.split(/\s+/).filter(Boolean)
+  }
+
+  getMatchedContent (): ContentMatchItem {
```
klinzo, line 103: Quite long method. Best split or move functionality to a utility function to make it reusable. May be necessary for the future.

src/routes/search/containers/SearchPage.js
```diff
@@ -43,10 +45,13 @@ export class SearchPage extends React.Component<PropsType, LocalStateType> {
...
     // find all categories whose contents but not titles include the filter text and sort them lexicographically
-    const categoriesWithContent = categories.toArray()
-      .filter(category => !normalizeSearchString(category.title).includes(filterText))
-      .filter(category => normalizeSearchString(category.content).includes(filterText))
-      .sort((category1, category2) => category1.title.localeCompare(category2.title))
+    // only filter content if the filter text is at least 3 characters long due to performance issues
+    const categoriesWithContent = filterText.length < CONTENT_MIN_FILTER_LENGTH ? []
+      : categories.toArray()
+        .filter(category => !normalizeSearchString(category.title).includes(filterText))
+        .filter(category => normalizeSearchString(htmlToText(category.content,
```
klinzo, line 52: So `htmlToText` is ran twice for a search, right? Would it be possible to change this so that is only ran once?

klinzo: Also, only filtering for the content starting after three characters may seem weird for users. I think we should search for a better solution.

src/routes/categories/components/CategoryList.js
```diff
@@ -4,7 +4,7 @@ import React from 'react'
 
 import RemoteContent from '../../../modules/common/components/RemoteContent'
 import Caption from '../../../modules/common/components/Caption'
-import CategoryListItem from './CategoryListItem'
+import CategoryListItem from './CategoryEntry'
```
maxammann, line 7:
```suggestion
import CategoryEntry from './CategoryEntry'
```

This is maybe a problem somewhere else too

src/routes/categories/components/__tests__/CategoryEntry.spec.js
```diff
@@ -0,0 +1,203 @@
...
+  lastUpdate: moment('2017-11-18T19:30:00.000Z'),
+  hash: 'a36a58'
+})
+
+// helper to find StyledComponents without importing them
+function findComponent (wrapper: ShallowWrapper<*>, name: string): ShallowWrapper<*> {
+  return wrapper.findWhere(n => n.name().endsWith(name))
+}
+
+describe('CategoryListItem', () => {
```
maxammann, line 66:
```suggestion
describe('CategoryEntry', () => {
```

src/routes/search/containers/SearchPage.js
```diff
@@ -13,8 +13,9 @@ import type { StateType } from '../../../modules/app/StateType'
 import SearchFeedback from '../components/SearchFeedback'
 import type { LocationState } from 'redux-first-router'
 import normalizeSearchString from '../../../modules/common/utils/normalizeSearchString'
+import { fromString as htmlToText } from 'html-to-text'
 
-type CategoryListItemType = {| model: CategoryModel, subCategories: Array<CategoryModel> |}
+type CategoryListItemType = {| model: CategoryModel, contentWithoutHtml?: string, subCategories: Array<CategoryModel> |}
```
maxammann, line 18: Not sure about this:
```suggestion
type CategoryentryType = {| model: CategoryModel, contentWithoutHtml?: string, subCategories: Array<CategoryModel> |}
```

src/routes/categories/components/CategoryEntry.js
Taggotty: I guess we agreed on sth like 8+. Therefore 10 might be a good number

src/routes/categories/components/__tests__/CategoryEntry.spec.js
```diff
@@ -119,23 +119,85 @@ describe('CategoryListItem', () => {
...
+          .toBe('This is some test')
+      })
+    })
+
     it('should return null for undefined query', () => {
       const wrapper = shallow(
         <CategoryEntry theme={brightTheme}
                        category={noThumbCategory}
-                       contentWithoutHtml={noThumbCategory.content}
+                       contentWithoutHptml={noThumbCategory.content}
```
maxammann, line 188:
```suggestion
                       contentWithoutHtml={noThumbCategory.content}
```

src/routes/search/containers/SearchPage.js
```diff
@@ -45,14 +46,26 @@ export class SearchPage extends React.Component<PropsType, LocalStateType> {
...
+        model: category,
+        contentWithoutHtml: htmlToText(category.content, {
+          ignoreHref: true,
+          ignoreImage: true,
+          wordwrap: false,
+          singleLineParagraph: true,
+          unorderedListItemPrefix: ' '
+        }),
+        subCategories: []
+      })).filter(categoryListItem =>
```
klinzo, line 59: We should probably also use the `CategoryEntry` naming here as well.

src/routes/categories/components/CategoryEntry.js
```diff
@@ -0,0 +1,179 @@
...
+        </StyledLink>
+      </SubCategory>
+    )
+  }
+
+  getWords (content: string): Array<string> {
+    return content.split(/\s+/).filter(Boolean)
+  }
+
+  getContentBeforeMatchIdx (content: string, matchIdx: number, startOfWord: boolean, numWords: number): string {
```
klinzo, line 103: What does `Idx` stand for?

Edit: Apparently index, right? Maybe just write it completely.

package.json
```diff
@@ -116,6 +117,7 @@
     "flow-bin": "~0.117.0",
     "flow-typed": "^2.5.1",
     "html-loader": "^0.5.5",
+    "html-to-text": "^5.1.1",
```
klinzo, line 120: Why did you switch to `htmlparser2`? Probably you should remove `html-to-text` then.

src/routes/categories/components/CategoryEntry.js
klinzo: There are still occurrences of `Idx` that could be changed. Not a must though.

package.json
Taggotty: Oh. Forgot to remove it. `html-to-text` had a large overhead. @Markl recommended htmlparser2. It is also a lot faster than before.

src/routes/categories/components/CategoryEntry.js
```diff
@@ -0,0 +1,179 @@
...
+          />
+          {this.renderTitle()}
+        </StyledLink>
+        {this.renderSubCategories()}
+      </Row>
+    )
+  }
+}
+
+export default withTheme(CategoryEntry)
```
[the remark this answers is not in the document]
Taggotty: you are right. Separating works like a charm and now the code looks a lot cleaner. thx :)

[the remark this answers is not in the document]
Schedulaar: Looks good :)

PR #385, opened 2020-07-28
WEBAPP-607: New ToolbarContainer Styles

lulo4907:
…r item hints for small viewports


build-configs/configs/integreat-test-cms/assets/manifest.json
```diff
@@ -1 +1 @@
-../../integreat/assets/manifest.json
\ No newline at end of file
+"../../integreat/assets/manifest.json"
```
maxammann, line 1: What is this file about? I kind of think this should not exist in the first place. Is it generated by webpack?

maxammann:
Please prepend your commits with the issue ticket id

src/modules/layout/components/StyledToolbarItem.js
```diff
@@ -3,14 +3,18 @@
...
   display: inline-block;
   margin: 0 10px;
   padding: 8px;
   cursor: pointer;
-  color: ${props => props.theme.colors.textColor};
   border: none;
+  color: ${props => props.theme.colors.textColor};
   background-color: ${props => props.theme.colors.backgroundColor};
+  text-align: center;
+`
```
maxammann, line 15:
```suggestion
`

```

src/modules/layout/components/StyledToolbarItem.js
```diff
@@ -3,14 +3,18 @@
...
+export const SmallViewTip = styled.p`
```
maxammann, line 16: Move this to another file and keep the `export default`

src/modules/layout/components/Toolbar.js
```diff
@@ -24,22 +28,38 @@ const ToolbarContainer = styled.div`
...
+    console.log('switcher', viewportSmall)
     return <ToolbarContainer className={this.props.className}>
+      {viewportSmall && <Headline>{t('isThisSiteUseful')}</Headline>}
       {this.props.children}
     </ToolbarContainer>
   }
 }
 
-export default Toolbar
+export default withTranslation('feedback')(connect<*, *, *, *, *, *>(mapStateToProps)(Toolbar))
```
maxammann, line 65: Avoid connecting a component to the state. Instead only containers should be connected to the state.

src/modules/layout/components/ToolbarItem.js
```diff
@@ -1,30 +1,38 @@
...
       <StyledToolbarItem href={href} ariaLabel={text}>
         <FontAwesomeIcon icon={icon} data-tip={text} data-event='mouseover' data-event-off='click mouseout' />
+        {viewportSmall && <SmallViewTip>{text}</SmallViewTip>}
       </StyledToolbarItem>
     )
   }
 }
 
-export default ToolbarItem
+export default connect<*, *, *, *, *, *>(mapStateToProps)(ToolbarItem)
```
maxammann, line 38: Avoid connecting a component to the state. Instead only containers should be connected to the state.

src/modules/layout/components/__tests__/ToolbarItem.spec.js
```diff
@@ -9,14 +9,21 @@ import ReactTooltip from 'react-tooltip'
...
 describe('ToolbarItem', () => {
   it('should rebuild tooltip items on mount', () => {
+    const store = createReduxStore()
     const original = ReactTooltip.rebuild
     ReactTooltip.rebuild = jest.fn()
 
     mount(
       <ThemeProvider theme={theme}>
-        <ToolbarItem href='http://example.com' icon={faSmile} text='Click here!' />
+        <Provider store={store}>
```
maxammann, line 24: Toolbar item should probably not be connected to the state. Therefore this can be removed

build-configs/configs/integreat-test-cms/assets/manifest.json
lulo4907: Hm, even if this file has been gegenerated automatically, it resist in this path and produces a failure message, that it is not a valid json file. And my IDE suggests double qoutes around the path instead of single ones --- after that I got rid of that message. I've had no further intentions with it. Let's talk about.

src/modules/layout/components/StyledToolbarItem.js
lulo4907: moved into a single file

src/modules/layout/components/Toolbar.js
lulo4907: This has been reverted by using props propagation from parent components now.

src/modules/layout/components/ToolbarItem.js
lulo4907: This has been reverted by using props propagation from parent components now.

src/modules/layout/components/__tests__/ToolbarItem.spec.js
lulo4907: removed

build-configs/configs/integreat-test-cms/assets/manifest.json
```diff
@@ -1 +1 @@
-../../integreat/assets/manifest.json
\ No newline at end of file
+'../../integreat/assets/manifest.json'
```
klinzo, line 1: Are you sure this is correct? This is a symlink and does reference the integreat manifest.json.

src/modules/layout/components/Layout.js
```diff
@@ -99,7 +99,8 @@ type PropsType = {|
   toolbar?: Node,
   modal?: Node,
   children?: Node,
-  darkMode?: boolean
+  darkMode?: boolean,
+  viewportSmall?: boolean
```
klinzo, line 103: Looks like this is not used anywhere at all.

src/modules/layout/components/Toolbar.js
```diff
@@ -24,22 +26,33 @@ const ToolbarContainer = styled.div`
...
+    const { viewportSmall, t } = this.props
+    return <ToolbarContainer className={this.props.className} viewportSmall={viewportSmall}>
+      {viewportSmall && <Headline>{t('isThisSiteUseful')}</Headline>}
       {this.props.children}
     </ToolbarContainer>
   }
 }
 
-export default Toolbar
+export default withTranslation('feedback')(Toolbar)
```
klinzo, line 58: Would it be possible to move that up as well to LocationLayout or any other parent component?

src/modules/layout/components/__tests__/ToolbarItem.spec.js
```diff
@@ -16,7 +16,7 @@ describe('ToolbarItem', () => {
 
     mount(
       <ThemeProvider theme={theme}>
-        <ToolbarItem href='http://example.com' icon={faSmile} text='Click here!' />
+          <ToolbarItem href='http://example.com' icon={faSmile} text='Click here!' viewportSmall />
```
klinzo, line 19: Wrong indentation I guess.

src/modules/layout/components/Toolbar.js
lulo4907: Could be, but not the slim way imo. The reason to place it here was, that it's the generic toolbar container component that imports all necessary items at least.

src/modules/layout/components/__tests__/ToolbarItem.spec.js
lulo4907: solved

src/modules/layout/components/Toolbar.js
klinzo: Yep, okay for me :+1:

klinzo:
Delete the manifest.json file and readd it from another branch to fix the builds. The symlink is still broken.

build-configs/configs/integreat-test-cms/assets/manifest.json
```diff
@@ -1 +1 @@
-../../integreat/assets/manifest.json
\ No newline at end of file
+../../integreat/assets/manifest.json
```
maxammann, line 1: Why is this a change? II see no change in the diff

klinzo:
There is still one flow error. Apart from that we are good to go :tada:

klinzo:
AFAICS the alignment is a little bit off, could you maybe fix this?
image

PR #388, opened 2020-08-05
WEBAPP-491: Moved constants to Api Client

lulo4907:


.idea/compiler.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
klinzo, line 1: Please remove this file.

lulo4907: done

PR #390, opened 2020-08-15
WEBAPP-586: Locale override

klinzo:

### DIFF
--- a/src/.eslintrc
+++ b/src/.eslintrc
@@ -8,7 +8,7 @@
   "globals": {
     "__DEV__": false,
     "__VERSION__": false,
-    "__CONFIG__": false
+    "__BUILD_CONFIG__": false
   },
   "rules": {
   }
--- a/src/.eslintrc
+++ b/src/.eslintrc
@@ -8,7 +8,7 @@
   "globals": {
     "__DEV__": false,
     "__VERSION__": false,
-    "__BUILD_CONFIG__": false
+    "__BUILD_CONFIG__": "writable"
   },
   "rules": {
   }
--- a/src/modules/i18n/__mocks__/locales.json
+++ b/src/modules/i18n/__mocks__/locales.json
@@ -7,30 +7,6 @@
         "localNews": "Lokale Nachrichten"
       },
       "metaDescription": "Integreat ist Ihr digitaler Guide für Deutschland. Finden Sie lokale Informationen, Veranstaltungen und Beratung. Immer aktuell und in Ihrer Sprache."
-    },
-    "ar": {
-      "pageTitles": {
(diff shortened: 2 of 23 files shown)


src/.eslintrc
```diff
@@ -8,7 +8,7 @@
   "globals": {
     "__DEV__": false,
     "__VERSION__": false,
-    "__CONFIG__": false
+    "__BUILD_CONFIG__": false
```
maxammann, line 11: The type of `__BUILD_CONFIG__` is BuildConfigType. This could cause problems. Also it is not expected that `__BUILD_CONFIG__` is null or false.

src/modules/i18n/__mocks__/locales.json
```diff
@@ -0,0 +1,98 @@
+{
```
maxammann, line 1: I think it would be good to make these mocks shorter. That way the snapshots get shorter.

src/modules/i18n/loadLocales.js
```diff
@@ -0,0 +1,35 @@
...
+      }
+    })
+    return transformedLocales
+  },
+  {}
+)
+
+const loadLocales = (): TransformedLocalesType => {
+  const localesOverride = buildConfig().localesOverride
+  const locales = localesOverride ? merge(defaultLocales, localesOverride) : defaultLocales
```
maxammann, line 31: Merge ignores values in localesOverrides if they are missing in defualtLocales: https://lodash.com/docs/4.17.15#merge

We could throw a warning here if localesOverride contains values which do not exist in defaultLocales.

src/.eslintrc
maxammann: Maybe convert this file to a .js file and load some default.

maxammann, line 11: Maybe

maxammann (approved):

src/modules/i18n/loadLocales.js
klinzo: Do you know whether there is a lodash function to do this?

klinzo: AFAICS there is none. Do you want me to check recursively? Or should I just add a comment?

src/.eslintrc
klinzo: Do you have an example for a js eslintrc file?

klinzo: https://eslint.org/docs/user-guide/configuring#specifying-globals

Changed it to `readonly`

src/modules/i18n/loadLocales.js
klinzo: I think a comment should be enough.

PR #391, opened 2020-08-15
WEBAPP-627: Parse release notes

klinzo:

### DIFF
--- a/.circleci/src/commands/restore_environment_variables.yml
+++ b/.circleci/src/commands/restore_environment_variables.yml
@@ -1,0 +1,8 @@
+description: Sets the environment variables specified in the file 'environment_variables'. Make sure the file is persisted and has been attached.
+steps:
+  - run:
+      name: List environment variables
+      command: cat environment_variables
+  - run:
+      name: Restore environment variables
+      command: cat environment_variables >> ${BASH_ENV}
--- a/tools/bump-version
+++ b/tools/bump-version
@@ -20,8 +20,8 @@ const commitAndTag = async (versionName, versionCodeString, {
   const appOctokit = await authenticate({ deliverinoPrivateKey, owner, repo })
   const versionFileContent = await appOctokit.repos.getContent({ owner, repo, path: VERSION_FILE, ref: branch })
 
-  const versionCode = parseInt(versionCodeString)
-  if (isNaN(versionCode)) {
+  const versionCode = versionCodeString ? parseInt(versionCodeString) : null
+  if (versionCode && isNaN(versionCode)) {
     throw new Error(`Failed to parse version code string: ${versionCodeString}`)
   }
 
--- a/tools/bump-version
+++ b/tools/bump-version
@@ -11,12 +11,22 @@ program
   .requiredOption('--owner <owner>', 'owner of the current repository, usually "Integreat"')
   .requiredOption('--repo <repo>', 'the current repository, usually "integreat-webapp" or "integreat-react-native-app"')
   .requiredOption('--branch <branch>', 'the current branch')
+  .requiredOption('--release-notes <release-notes>', 'the release notes as stringified string')
+  .option('--development-release', 'whether the release is a development release which is not delivered to production')
 
-const commitAndTag = async (versionName, versionCode, { deliverinoPrivateKey, owner, repo, branch }) => {
+const commitAndTag = async (versionName, versionCodeString, {
+  deliverinoPrivateKey, owner, repo, branch, developmentRelease, releaseNotes
+}) => {
   const appOctokit = await authenticate({ deliverinoPrivateKey, owner, repo })
   const versionFileContent = await appOctokit.repos.getContent({ owner, repo, path: VERSION_FILE, ref: branch })
 
-  const contentBase64 = Buffer.from(JSON.stringify({ versionName, versionCode })).toString('base64')
+  const versionCode = parseInt(versionCodeString)
+  if (isNaN(versionCode)) {
+    throw new Error(`Failed to parse version code string: ${versionCodeString}`)
+  }
+
+  const contentBase64 = Buffer.from(JSON.stringify({ versionName, versionCode }))
+    .toString('base64')
 
   const message = versionCode
     ? `Bump version name to ${versionName} and version code to ${versionCode}\n[skip ci]`
@@ -35,7 +45,7 @@ const commitAndTag = async (versionName, versionCode, { deliverinoPrivateKey, ow
 
   const commitSha = commit.data.commit.sha
 
-  const releaseName = `${versionName}${versionCode ? ` - version code: ${versionCode}` : ''}`
+  const releaseName = `${developmentRelease ? '[Development release] ' : ''}${versionName}${versionCode ? `-${versionCode}` : ''}`
 
   const tag = await appOctokit.git.createTag({
     owner,
@@ -54,15 +64,18 @@ const commitAndTag = async (versionName, versionCode, { deliverinoPrivateKey, ow
     sha: tagSha
   })
 
-  const releaseNotesUrl = `https://github.com/${owner}/${repo}/tree/${branch}/release-notes/${versionName}`
-  const releaseMessage = `Release ${releaseName}. Release notes can be found here: ${releaseNotesUrl}.`
+  const releaseMessage = developmentRelease
+    ? `Development release ${releaseName}. This release is only delivered to development (Testflight and Beta channel).`
+    : `Release ${releaseName}.`
+
+  const body = `${releaseMessage}\n${JSON.parse(releaseNotes)}`
 
   await appOctokit.repos.createRelease({
     owner,
     repo,
     tag_name: versionName,
     name: versionName,
-    body: releaseMessage
+    body
   })
 }
 
--- a/tools/bump-version
+++ b/tools/bump-version
@@ -45,7 +45,7 @@ const commitAndTag = async (versionName, versionCodeString, {
 
   const commitSha = commit.data.commit.sha
 
-  const releaseName = `${developmentRelease ? '[Development release] ' : ''}${versionName}${versionCode ? `-${versionCode}` : ''}`
+  const releaseName = `${developmentRelease ? '[Development release] ' : ''}${versionName}${versionCode ? ` - ${versionCode}` : ''}`
 
   const tag = await appOctokit.git.createTag({
     owner,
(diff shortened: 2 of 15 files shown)


.circleci/src/commands/restore_environment_variables.yml
```diff
@@ -0,0 +1,8 @@
+description: Sets the environment variables specified in the file 'environment_variables'. Make sure the file is persisted and has been attached.
+steps:
+  - run:
+      name: List environment variables
+      command: cat environment_variables
```
Taggotty, line 5: Just to make sure, this command is especially used for debugging and stuff, right?

tools/parse-release-notes
```diff
@@ -0,0 +1,118 @@
...
+  if (ios) {
+    platforms.push(PLATFORM_IOS)
+  }
+  if (web) {
+    platforms.push(PLATFORM_WEB)
+  }
+
+  if (platforms.length === 0) {
+    throw new Error('No platforms selected! Use --ios, --android and --web flags.')
+  } else if (platforms.length === 3) {
```
Taggotty, line 54: here it might be better to be explicit: ```web && (ios || android)```

tools/parse-release-notes
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env node
+
+const fs = require('fs')
+const { GITKEEP_FILE, UNRELEASED_DIR, RELEASE_NOTES_DIR } = require('./constants')
+const { program } = require('commander')
+const yaml = require('js-yaml')
+
+const defaultReleaseNotesEn = 'We’ve been working hard behind the scenes to make sure everything is working as it should. If you notice anything that does not work, let us know!\n'
+const defaultReleaseNotesDe = 'Wir haben hinter den Kulissen hart gearbeitet, um sicherzustellen, dass alles so funktioniert, wie es sollte. Wenn Sie etwas bemerken, das nicht funktioniert, lassen Sie es uns wissen!\n'
```
Taggotty, line 9:
```suggestion
const defaultReleaseNotesDe = 'Wir haben hinter den Kulissen hart gearbeitet, um sicherzustellen, dass alles so funktioniert, wie es soll. Wenn Sie bemerken, das etwas nicht funktioniert, lassen Sie es uns wissen!\n'
```

tools/parse-release-notes
```diff
@@ -0,0 +1,118 @@
...
+      notesMap.android.push(note)
+    } else if (isNoteRelevant(PLATFORM_IOS)) {
+      notesMap.ios.push(note)
+    }
+    return notesMap
+  }, { common: [], android: [], ios: [], web: [] })
+
+  const commonNotes = formatNotes({ notes: notesMap.common, language, production })
+  const androidNotes = formatNotes({ notes: notesMap.android, language, production, platformName: PLATFORM_ANDROID })
+  const iosNotes = formatNotes({ notes: notesMap.ios, language, production, platformName: PLATFORM_IOS })
```
Taggotty, line 94: does that mean that web release notes are all common?

klinzo: Currently they are, yes, because webapp release notes are generated separately and we only have one platform there. I'll add a web notes option though, maybe we need that for mono repo :shrug:

klinzo:
@Taggotty  Yes, I have tested it already. But probably I missed anything nevertheless :D
Could you create an issue for this?

tools/bump-version
```diff
@@ -11,12 +11,22 @@ program
...
 
-const commitAndTag = async (versionName, versionCode, { deliverinoPrivateKey, owner, repo, branch }) => {
+const commitAndTag = async (versionName, versionCodeString, {
+  deliverinoPrivateKey, owner, repo, branch, developmentRelease, releaseNotes
+}) => {
   const appOctokit = await authenticate({ deliverinoPrivateKey, owner, repo })
   const versionFileContent = await appOctokit.repos.getContent({ owner, repo, path: VERSION_FILE, ref: branch })
 
-  const contentBase64 = Buffer.from(JSON.stringify({ versionName, versionCode })).toString('base64')
+  const versionCode = versionCodeString ? parseInt(versionCodeString) : null
```
maxammann, line 23: I think this should be the previous value and not `null` or `undefined`.

tools/parse-release-notes
```diff
@@ -0,0 +1,126 @@
...
+  if (web) {
+    platforms.push(PLATFORM_WEB)
+  }
+
+  if (platforms.length === 0) {
+    throw new Error('No platforms selected! Use --ios, --android and --web flags.')
+  } else if (web && (ios || android)) {
+    throw new Error('Usage of --ios or --android together with --web flag is not supported.')
+  } else if (platforms.length > 1 && production) {
+    throw new Error('Usage of multiple platforms in production mode is not supported.')
```
maxammann, line 62: Why is this a problem?

tools/parse-release-notes
```diff
@@ -0,0 +1,126 @@
...
+
+  // If the production flag is set, hide information that is irrelevant for users
+  if (production) {
+    const productionNotes = relevantNotes.filter(note => note.show_in_stores)
+
+    if (productionNotes.length > 0) {
+      return formatNotes({ notes: productionNotes, language, production })
+    }
+
+    return language === 'de' ? DEFAULT_NOTE_DE : DEFAULT_NOTE_EN
```
maxammann, line 81: I'd favor to have this a little bit more generic to support more default notes with german as fallback.

e.g. `return DEFAULT_LANGUAGES[language] || DEFAULT_NOTE_DE`

tools/parse-release-notes
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env node
+
+const fs = require('fs')
+const { GITKEEP_FILE, UNRELEASED_DIR, RELEASE_NOTES_DIR } = require('./constants')
+const { program } = require('commander')
+const yaml = require('js-yaml')
+
+const DEFAULT_LANGUAGE = 'en'
```
maxammann, line 8: Should be 'de' like in the translations of the app.

maxammann (approved):

tools/parse-release-notes
```diff
@@ -0,0 +1,126 @@
...
+
+  // Load all notes not belonging to a release
+  const relevantNotes = fileNames
+    .filter(fileName => fileName !== GITKEEP_FILE)
+    .map(fileName => yaml.safeLoad(fs.readFileSync(`${source}/${fileName}`)))
+    .filter(note => isNoteRelevant({ note, platforms }))
+
+  // If the production flag is set, hide information that is irrelevant for users
+  if (production) {
+    const productionNotes = relevantNotes.filter(note => note.show_in_stores)
```
maxammann, line 75: What do you think about this:

IF there are only non-production notes THEN show a random non-production note instead of the default note.

That way we avoid generating the same release notes.

tools/parse-release-notes
```diff
@@ -0,0 +1,126 @@
...
+  .option('-d, --debug', 'enable extreme logging')
+  .option('--ios', 'include release notes for ios')
+  .option('--android', 'include release notes for android')
+  .option('--web', 'include release notes for web. may not be used with --android or --ios flags')
+  .option('--production', 'whether to hide extra information, e.g. issue keys, hidden notes and platforms and prepare the notes for a store. may not be used with multiple platforms.')
+  .option('--destination <destination>', 'if specified the parsed notes are saved to the directory')
+  .requiredOption('--source <source>', 'the directory of the release notes to parse', `${RELEASE_NOTES_DIR}/${UNRELEASED_DIR}`)
+  .requiredOption('--language <language>', 'the language of the release notes to parse', DEFAULT_LANGUAGE)
+
+const formatNotes = ({ notes, language, production, platformName }) => {
```
maxammann, line 27: I think we should prepend here the DEFAULT_NOTE_EN or DEFAULT_NOTE_DE. Else the notes may sound a little bit technical.

I think adding some nice words would be good for stores at least.

tools/parse-release-notes
```diff
@@ -0,0 +1,126 @@
...
+
+const DEFAULT_NOTE_EN = 'We’ve been working hard behind the scenes to make sure everything is working as it should. If you notice anything that does not work, let us know!\n'
+const DEFAULT_NOTE_DE = 'Wir haben hinter den Kulissen hart gearbeitet, um sicherzustellen, dass alles so funktioniert, wie es soll. Wenn Sie bemerken, dass etwas nicht funktioniert, lassen Sie es uns wissen!\n'
+
+const PLATFORM_ANDROID = 'android'
+const PLATFORM_IOS = 'ios'
+const PLATFORM_WEB = 'web'
+
+program
+  .option('-d, --debug', 'enable extreme logging')
```
maxammann, line 18:
```suggestion
```

This is unused.

tools/parse-release-notes
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env node
+
+const fs = require('fs')
+const { GITKEEP_FILE, UNRELEASED_DIR, RELEASE_NOTES_DIR } = require('./constants')
+const { program } = require('commander')
+const yaml = require('js-yaml')
+
+const DEFAULT_LANGUAGE = 'en'
+
+const DEFAULT_NOTE_EN = 'We’ve been working hard behind the scenes to make sure everything is working as it should. If you notice anything that does not work, let us know!\n'
```
maxammann, line 10: Should we add here an array of messages out of which we randomly select one?

tools/parse-release-notes
klinzo: Because it does not make sense and probably a usage error. Nobody wants to generate "production release notes" (meaning release notes that are shown in the stores) for multiple platforms. If you want to have release notes for multiple platforms, it is always for internal purposes (e.g. github release notes).

tools/parse-release-notes
klinzo: Are you sure? Store descriptions are nearly everywhere english if the language is not provided.

tools/bump-version
klinzo: What is the previous value if there is no version code string? :thinking:

tools/parse-release-notes
klinzo: I don't really like it to be honest.
Firstly, there may be no non production notes as well and this just adds more complexcity to the script.
Secondly, Non production notes are probably quite technically and not user related at all. I think users could be really confused if there is suddenly a release note saying e.g. Added design for Malte or Fixed some stuff in the CI.

Maybe we could rather add a production and a non production release notes for technical tasks? E.g. if working on the CI:
We made sure that you are always getting your app updates in time vs Fixed failing tests in the CI.

Also, we could add more default texts and pick one randomly. I would do this in a separate task though. If you like the idea could you create an issue?

tools/parse-release-notes
klinzo: Should only be done for production release notes, so I'll add this below.. But good idea.

tools/parse-release-notes
klinzo: Yep, but probably better to do this in a separate issue. I'll create one.

tools/parse-release-notes
maxammann: Oh it is the length of platforms. Then it is good!

PR #396, opened 2020-08-30
WEBAPP-621: Configurable Icons

Taggotty:


Taggotty:
Actually the icons were already configurable in the webapp or am I wrong there? Therefore I just wrapped the Icons to make it more intuitive, I guess. What do you think?

build-configs/configs/BuildConfigType.js
```diff
@@ -14,8 +14,10 @@ export type BuildConfigType = {|
   manifestUrl?: string,
   itunesAppId?: string,
   localesOverride?: LocalesType,
-  logoWide: string,
-  locationIcon: string,
+  icons: {
+    locationIcon: string,
+    logoWide: string,
```
maxammann, line 19: Minor: If you want you can rename this to `headerLogo`

maxammann (approved):

maxammann:
Build is failing because the buildConfig is not a funciton like in NATIVE. You can convert it such that buildConfig() is a function

PR #401, opened 2020-09-09
WEBAPP-644: EventJsonLd timezone is wrong

lulo4907:

### DIFF
--- a/tools/webpack.config.babel.js
+++ b/tools/webpack.config.babel.js
@@ -4,7 +4,6 @@ const AssetsPlugin = require('assets-webpack-plugin')
 const HtmlWebpackPlugin = require('html-webpack-plugin')
 const { CleanWebpackPlugin } = require('clean-webpack-plugin')
 const CopyPlugin = require('copy-webpack-plugin')
-const MomentTimezoneDataPlugin = require('moment-timezone-data-webpack-plugin')
 const MomentLocalesPlugin = require('moment-locales-webpack-plugin')
 const babelConfig = require('../.babelrc.js')
 const fs = require('fs')
@@ -145,8 +144,6 @@ const createConfig = (env = {}) => {
         debug: !isProductionBuild,
         minimize: isProductionBuild
       }),
-      // We only use moment-timezone for parsing GMT time-zoned data in the integreat-api-client
-      new MomentTimezoneDataPlugin({ matchZones: 'GMT' }),
       // moment has no support for 'ti' (Tigrinya) and 'so' (Somali), hence we have to use the ignoreInvalidLocales flag
       new MomentLocalesPlugin({ localesToKeep: getSupportedLocales(), ignoreInvalidLocales: true })
     ],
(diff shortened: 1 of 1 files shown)


Schedulaar (commented):
Well spotted, that this is caused by the plugin.
Indeed, I probably overlooked the usage in the events endpoint.
As we don't get a UTC normed date, we need to use timezone data for this.

Still, you can use MomentTimezoneDataPlugin to reduce the bundle size (including all timezone data would probably increase bundle size by ~200kB). What makes sense for us is to only include timezone data ranging from "this year two years onwards" or so. (see https://github.com/gilmoreorless/moment-timezone-data-webpack-plugin#all-zones-with-a-limited-date-range).
```js
const currentYear = new Date().getFullYear();
const plugin = new MomentTimezoneDataPlugin({
  startYear: currentYear,
  endYear: currentYear + 2
})
```
Additionally, one could think about limiting the data to only certain regions, but that's a bit risky...

Your build is currently failing, as your bundle exceeds the current budget of 1.5MiB. If your build is still failing with the optimizations applied above, you can think about increasing the budget.

maxammann:
Thanks for your input @Schedulaar !
Yeah lets apply the time ranges and then increase the max. bundle size.

lulo4907:
The plugin has been reentered, the entrypoint is now at 1.51 MiB.
What we need to discuss is the technical side of the time zones at the CMS. because not all selectable time zones are compatible with the plugin and runs into JS errors.

maxammann:
@lulo4907 that is weird, can you explain and mention an example? The name should be the same as here: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

maxammann:
To fix the build increase the asset size limit to 1.6 MiB

maxammann:
Screenshot_1600074411


I noticed this on NATIVE. Could this be related?

Schedulaar (approved):

klinzo (approved):

tools/webpack.config.babel.js
```diff
@@ -146,7 +148,7 @@ const createConfig = (env = {}) => {
         minimize: isProductionBuild
       }),
       // We only use moment-timezone for parsing GMT time-zoned data in the integreat-api-client
```
klinzo, line 150: Maybe update this comment accordingly

lulo4907:
@maxammann 
The Webapp event example using "UTC-12" : ~/testumgebung/de/events/testevent-3
image

Depending on the configuration script more or less time zones are found. But some others, like "UTC-12" do not exist.

tools/webpack.config.babel.js
```diff
@@ -147,7 +147,7 @@ const createConfig = (env = {}) => {
         debug: !isProductionBuild,
         minimize: isProductionBuild
       }),
-      // We only use moment-timezone for parsing GMT time-zoned data in the integreat-api-client
+      // We use moment-timezone for parsing a limited range of years here with GTM data in the integreat-api-client
```
Schedulaar, line 150:
```suggestion
      // As all parsed dates are in the near future, we limit the timezone-data to only include the next two years.
```

Schedulaar (approved):

maxammann (approved):

